PR #34, opened 2017-03-10
Creates infobox part

tbiScireum:


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.8.2</version>
     </parent>
     <artifactId>sirius-biz</artifactId>
-    <version>1.9.3</version>
+    <version>1.9.4</version>
```
andyHa: 1.10 - not a bugfix

PR #50, opened 2017-05-05
Adds infoboxElement Template

tbiScireum:
Moves the element code of the infobox template into an own template.

### DIFF
--- a/src/main/resources/default/view/parts/infoboxElement.html
+++ b/src/main/resources/default/view/parts/infoboxElement.html
@@ -1,0 +1,12 @@
+@args String key, String value
+
+<div class="row info-line">
+    <div class="col-md-4 name">
+        @key
+    </div>
+    <div class="col-md-8 value">
+        <div>
+            @value
+        </div>
+    </div>
+</div>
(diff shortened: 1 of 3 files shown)


src/main/resources/default/view/parts/infoboxElement.html
```diff
@@ -0,0 +1,12 @@
...
+<div class="row info-line">
+    <div class="col-md-4 name">
+        @i18n(key)
+    </div>
+    <div class="col-md-8 value">
+        <div>
+            @value
+        </div>
+    </div>
+</div>
```
duchuyctlk: Please add an empty line at end of file.

tbiScireum: why?

src/main/resources/default/view/parts/infoboxElement.html
```diff
@@ -0,0 +1,12 @@
+@args String key, String value
```
andyHa: labelKey - heisst es sonst überall auch

PR #51, opened 2017-05-17
Moves the presaveHandler to be executed after the entity load

tbiScireum:
Adds an assertTenant Method and an onSaveHandler to assert EntityRefs during the save process.

### DIFF
--- a/src/main/java/sirius/biz/web/BizController.java
+++ b/src/main/java/sirius/biz/web/BizController.java
@@ -17,6 +17,7 @@ import sirius.db.mixing.Entity;
 import sirius.db.mixing.EntityRef;
 import sirius.db.mixing.OMA;
 import sirius.db.mixing.Property;
+import sirius.db.mixing.annotations.NullAllowed;
 import sirius.db.mixing.properties.BooleanProperty;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.ConfigValue;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -78,6 +79,24 @@ protected void assertTenant(TenantAware tenantAware) {
...
     /**
+     * Ensures that the tenant of the current user matches the tenant of the given referenced object.
+     * <p>
+     * Will also check if the referenced object is null and if null is allowed.
+     *
+     * @param entity       source object holding the reference
+     * @param tenantAware  reference object of the actual tenant aware object
+     * @param tenantColumn Column of the tenantAware {@link EntityRef}
+     */
+    protected void assertTenant(Entity entity, EntityRef<? extends TenantAware> tenantAware, Column tenantColumn) {
```
andyHa: is there any tenantware anywhere, where tenant may be null??

src/main/java/sirius/biz/web/BizController.java
```diff
@@ -319,6 +352,11 @@ public boolean saveEntity(Entity entity) {
                 } else {
                     load(ctx, entity);
                 }
+
```
andyHa: couldn't we just move the presave handler here?

tbiScireum: did

PR #52, opened 2017-05-19
Adds assertionMethod to check for currentTenant or its parent

tbiScireum:


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -82,6 +82,22 @@ protected void assertTenant(TenantAware tenantAware) {
...
+     *
+     * @param tenantAware {@link TenantAware} entity to be asserted
+     */
+    protected void assertTenantOrParentTenant(TenantAware tenantAware) {
+        if (tenantAware == null) {
+            return;
+        }
+
+        if (!tenantAware.getTenant().is(currentTenant()) && !tenantAware.getTenant()
+                                                                        .is(currentTenant().getParent().getValue())) {
```
andyHa: getParent().getId()

PR #54, opened 2017-06-06
Shows the current tenant name in addition to the username

tbiScireum:


src/main/resources/default/view/wondergem/menu-right-user.html
```diff
@@ -1,7 +1,7 @@
 @if (user.getCurrentUser().isLoggedIn()) {
     <li class="dropdown">
         <a href="#" class="dropdown-toggle" data-toggle="dropdown">
-            @user.getCurrentUser().getUserName()
+            @user.getCurrentUser().getUserName() (@user.getCurrentUser().getTenantName())
```
andyHa: surround with strings.isfilled maybe?

PR #59, opened 2017-06-21
Load explicit named columns in addition to all autoloaded columns

tbiScireum:
This will reduce the number of columns that needs to be explicitly named


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.10</version>
     </parent>
     <artifactId>sirius-biz</artifactId>
-    <version>3.7</version>
+    <version>3.7.1</version>
```
andyHa: not a bugfix -> 3.8

PR #60, opened 2017-06-23
Won't apply the filter query, if no filter value is set

tbiScireum:
Otherwise the default filter value would be false instead of fetching all entries
no matter if true or false

### DIFF
--- a/src/main/java/sirius/biz/web/PageHelper.java
+++ b/src/main/java/sirius/biz/web/PageHelper.java
@@ -233,9 +233,9 @@ public class PageHelper<E extends Entity> {
         facet.addItem("false", NLS.get("NLS.no"), -1);
 
         return addFacet(facet, (f, q) -> {
-            Value isUsed = Value.of(f.getValue());
-            if (isUsed.isFilled()) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/web/PageHelper.java
```diff
@@ -232,7 +232,12 @@ private PageHelper() {
         facet.addItem("true", NLS.get("NLS.yes"), -1);
         facet.addItem("false", NLS.get("NLS.no"), -1);
 
-        return addFacet(facet, (f, q) -> q.eqIgnoreNull(Column.named(f.getName()), Value.of(f.getValue()).asBoolean()));
+        return addFacet(facet, (f, q) -> {
+            Value isUsed = Value.of(f.getValue());
```
andyHa: "filterValue" oder so?

src/main/java/sirius/biz/web/PageHelper.java
```diff
@@ -232,7 +232,12 @@ private PageHelper() {
...
+            if (isUsed.isFilled()) {
+                q.eqIgnoreNull(Column.named(f.getName()), isUsed.asBoolean());
```
Tobsucht: ignoreNull can be removed

PR #66, opened 2017-06-29
Provides an initial implementation of the object storage system

andyHa:


src/main/java/sirius/biz/storage/DownloadBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+    private String version;
+    private String physicalKey;
+    private boolean eternallyValid;
+
+    @Part
+    private static VersionManager versionManager;
+
+    /**
+     * Creates a new builder with a direct reference to the storage and the object key.
+     *  @param storage   the storage implementation
```
fkrone: Code format

src/main/java/sirius/biz/storage/Storage.java
```diff
@@ -0,0 +1,568 @@
...
+     * Tries to find the object with the given id, for the given tenant and bucket.
+     *
+     * @param tenant the tenant to filter on
+     * @param bucket the bucket to search in
+     * @param key    the globally unique id to search by
+     * @return the object in the given bucket with the given key wrapped as optional or an empty optional if no such
+     * object exists.
+     */
+    public Optional<StoredObject> findByKey(@Nullable Tenant tenant, String bucket, String key) {
+        return Optional.ofNullable(oma.select(VirtualObject.class)
```
fkrone: use .first() instead

andyHa: doesn't work due to Javas broken type system...

PR #69, opened 2017-07-04
Makes pageSizes adjustable in the PageHelper

PhilHup:
Creates the possibility to adjust the pageSize for rendered Pages while using the PageHelper.
Renames static PAGE_SIZE field to DEFAULT_PAGE_SIZE.

### DIFF
--- a/src/main/java/sirius/biz/web/PageHelper.java
+++ b/src/main/java/sirius/biz/web/PageHelper.java
@@ -247,7 +247,7 @@ public class PageHelper<E extends Entity> {
      * @param pageSize the number of items shown per page
      * @return the helper itself for fluent method calls
      */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/web/PageHelper.java
```diff
@@ -241,14 +242,25 @@ private PageHelper() {
     }
 
     /**
+     * Specifies the number of items shown on the page that gets rendered using this pageHelper.
+     *
+     * @param pageSize the number of items shown per page
+     * @return the helper itself for fluent method calls
+     */
+    public PageHelper<E> setPageSize(int pageSize) {
```
tbiScireum: withPageSize

PR #75, opened 2017-07-11
Implements the tracing template for tagliattele

tbiScireum:


src/main/resources/taglib/w/tracing.html.pasta
```diff
@@ -0,0 +1,41 @@
+<i:arg name="trace" type="sirius.biz.model.TraceData"/>
+<i:pragma name="description" value="Renders a trace button and the trace information for a TracedEntity"/>
```
andyHa: inline plz

PR #76, opened 2017-07-13
Adds the menu settings dropdown menu entry

tbiScireum:
and the codelists entry

### DIFF
--- a/src/main/resources/component-biz.conf
+++ b/src/main/resources/component-biz.conf
@@ -58,6 +58,20 @@ jobs {
 }
 
 content.extensions {
+    templates-page-menu-left {
+        settings {
+            priority = 1000
+            template = "templates/wondergem/menu-left-settings.html.pasta"
+        }
+    }
+
+    template-menu-settings {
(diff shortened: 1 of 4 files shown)


src/main/resources/component-biz.conf
```diff
@@ -58,6 +58,20 @@ jobs {
...
 
 content.extensions {
+    templates-page-menu-left {
+        settings {
+            priority = 1000
+            template = "templates/wondergem/menu-left-settings.html.pasta"
+        }
+    }
+
+    template-menu-settings {
```
andyHa: templates-

src/main/resources/default/templates/wondergem/menu-left-settings.html.pasta
```diff
@@ -0,0 +1,3 @@
+<w:menuDropdown titleKey="Menu.settings" permission="flag-logged-in">
+    <i:extensions name="template-menu-settings"/>
```
andyHa: templates-

PR #112, opened 2017-10-30
Implements a minimum size for image versions

jmuscireum:
The version string of VersionManager now supports the specification of a minimum size. The new syntax is a comma separated string of key-value pairs in the form of `key:value`. The available keys are `size` for the requested maximum size, and `min` for the minimum size. The value is the size in the format `<width>x<height>`.

Example: The version `size:500x500,min:100x100` would scale an image with 1000x800 pixels to 500x400. An image with 1000x150 would be scaled to 500x75 and then extended with a white border to 500x100. An image with 400x50 pixels would not be scaled but extended to 400x100.

### DIFF
--- a/src/main/java/sirius/biz/storage/VersionManager.java
+++ b/src/main/java/sirius/biz/storage/VersionManager.java
@@ -155,37 +155,28 @@ public class VersionManager {
     private void performVersionComputation(VirtualObjectVersion objectVersion) {
         VirtualObject object = objectVersion.getVirtualObject().getValue();
         try {
-            int width = 0;
-            int height = 0;
-            int extendWidth = 0;
-            int extendHeight = 0;
+            Tuple<Integer, Integer> size = Tuple.create(0, 0);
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/biz/storage/VersionManager.java
```diff
@@ -150,11 +155,37 @@ private void computeVersion(VirtualObject object, String version) {
...
-            int height = Integer.parseInt(widthAndHeight.getSecond());
+            int width = 0;
+            int height = 0;
+            int extendWidth = 0;
+            int extendHeight = 0;
+
+            for (String part : objectVersion.getVersionKey().split(",")) {
+                Tuple<String, String> keyValuePair = Strings.split(part, ":");
+
+                switch (keyValuePair.getFirst().toLowerCase().trim()) {
```
andyHa: if und jeweils in ne methode - switch = pistole+fuss

PR #124, opened 2018-01-23
Implements a getRequiredValue function for code lists

jmuscireum:


src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -71,6 +75,26 @@ public boolean hasValue(@Nonnull String codeList, @Nullable String code) {
         return oma.select(CodeListEntry.class).eq(CodeListEntry.CODE_LIST, cl).eq(CodeListEntry.CODE, code).exists();
     }
 
+    /**
+     * Tries to return the value from the given code list associated with the given code.
+     *
+     * @param codeList the code list to search in
+     * @param code     the code to lookup
+     * @return the value associated with the code
```
andyHa: @throws

src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -71,6 +75,26 @@ public boolean hasValue(@Nonnull String codeList, @Nullable String code) {
         return oma.select(CodeListEntry.class).eq(CodeListEntry.CODE_LIST, cl).eq(CodeListEntry.CODE, code).exists();
     }
 
+    /**
+     * Tries to return the value from the given code list associated with the given code.
```
andyHa: Returns.... throws an exception if no matching entry exists

src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -71,6 +75,26 @@ public boolean hasValue(@Nonnull String codeList, @Nullable String code) {
...
+     *
+     * @param codeList the code list to search in
+     * @param code     the code to lookup
+     * @return the value associated with the code
+     */
+    public String getRequiredValue(@Nonnull String codeList, @Nullable String code) {
+        if (!hasValue(codeList, code)) {
```
andyHa: call getValues directly and either throw or return the 1st value

PR #126, opened 2018-01-30
Completion handling on file update

fkrone:
Updating a file in the storage engine now supports a completion handling for using the updates StoredObject after being updated in the storage engine.

### DIFF
--- a/src/main/java/sirius/biz/storage/Storage.java
+++ b/src/main/java/sirius/biz/storage/Storage.java
@@ -20,6 +20,7 @@ import sirius.db.mixing.OMA;
 import sirius.db.mixing.SmartQuery;
 import sirius.db.mixing.constraints.FieldOperator;
 import sirius.kernel.Sirius;
+import sirius.kernel.async.CompletionHandler;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/storage/Storage.java
```diff
@@ -439,6 +440,23 @@ public OutputStream updateFile(@Nonnull StoredObject file) {
...
+     * Note that most probably, the file will be updated once the stream is closed and not immediatelly on a write.
+     * Also note that it is essential to close the stream to release underlying resources.
+     * <p>
+     * After closing the stream, based on the result of writing the file, the complectionHandler is called.
+     *
+     * @param file              the file to update.
+     * @param completionHandler completion handler called once the file is written
+     * @return an output stream to write the contents to.
+     */
+    public OutputStream updateFileWithCompletionHandling(@Nonnull StoredObject file,
```
andyHa: isn't close on the outputstream called on the caller of this method - which also obviously has the StoredObject at hands? so when tell the caller that he called close?

fkrone:
Not needed any longer

PR #137, opened 2018-04-19
Add option for buckets to always search with like

fkrone:
For some buckets it is desired that the search is executed as if you would like to search only parts of the file name.


src/main/java/sirius/biz/storage/StorageController.java
```diff
@@ -150,15 +150,24 @@ private void findObjectsForQuery(String bucketName, String query, Consumer<Autoc
...
+     * @param alwaysUseLikeSearch determines wheter we always use a like on for matching
      */
-    private void applyQuery(String query, SmartQuery<VirtualObject> baseQuery) {
+    private void applyQuery(String query, SmartQuery<VirtualObject> baseQuery, boolean alwaysUseLikeSearch) {
         if (Strings.isEmpty(query)) {
             return;
         }
 
         String queryString = query;
+        if (alwaysUseLikeSearch) {
```
andyHa: öhm bitte direkt Like.contains nehmen

PR #138, opened 2018-04-19
Aha/sexy ur ls

andyHa:


src/main/java/sirius/biz/storage/Storage.java
```diff
@@ -581,6 +581,10 @@ private String buildURL(DownloadBuilder downloadBuilder) {
             result.append(computeHash(downloadBuilder.getPhysicalKey(), 0));
         }
         result.append("/");
+        if (Strings.isFilled(downloadBuilder.getAddonText())) {
+            result.append(Strings.reduceCharacters(downloadBuilder.getAddonText()).replaceAll("[^a-zA-Z0-9\\-_.]", "-"));
```
tbiScireum: We use this replaceAll in sirius-kernel Files.java as well
How about a Files.replaceSth Function ?

andyHa: for all seo strings ( url parts) we use "-" as replacement. Files uses "_"

andyHa: but we could optimize the regex a bit

PR #143, opened 2018-05-08
 Clears the object version cache when removing versions

jmuscireum:
When a virtual object was updated the versions where only removed from the database, but not the cache. This lead to outdated URLs and therefore broken images.

### DIFF
--- a/src/main/java/sirius/biz/storage/Storage.java
+++ b/src/main/java/sirius/biz/storage/Storage.java
@@ -90,6 +90,9 @@ public class Storage {
     @Part
     private GlobalContext context;
 
+    @Part
+    private VersionManager versionManager;
+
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/biz/storage/Storage.java
```diff
@@ -424,7 +427,8 @@ public void updateFile(@Nonnull StoredObject file,
             virtualObjectCache.remove(object.getObjectKey());
 
             // Delete all (now outdated) versions
-            oma.select(VirtualObjectVersion.class).eq(VirtualObjectVersion.VIRTUAL_OBJECT, object).delete();
```
andyHa: könnte man auch in den afterdelete handler von VirtualObjectVersion packen - dann tuts immer net nur meistens

PR #148, opened 2018-06-13
Fix changing the password in the backend

jmuscireum:
When changing the password under /user-accounts the old password was not checked and the new password was read from the wrong parameter.

### DIFF
--- a/src/main/java/sirius/biz/tenants/UserAccountController.java
+++ b/src/main/java/sirius/biz/tenants/UserAccountController.java
@@ -33,6 +33,7 @@ import sirius.web.mails.Mails;
 import sirius.web.security.LoginRequired;
 import sirius.web.security.Permission;
 import sirius.web.security.UserContext;
+import sirius.web.security.UserInfo;
 import sirius.web.services.JSONStructuredOutput;
 
 import java.util.Collections;
@@ -51,6 +52,8 @@ public class UserAccountController extends BizController {
     public static final String PERMISSION_MANAGE_USER_ACCOUNTS = "permission-manage-user-accounts";
 
     private static final String PARAM_PASSWORD = "password";
+    private static final String PARAM_OLD_PASSWORD = "oldPassword";
+    private static final String PARAM_NEW_PASSWORD = "newPassword";
     private static final String PARAM_CONFIRMATION = "confirmation";
     private static final String PARAM_NAME = "name";
     private static final String PARAM_USERNAME = "username";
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/tenants/UserAccountController.java
```diff
@@ -218,6 +224,17 @@ public void password(final WebContext ctx, String id) {
         ctx.respondWith().template("templates/tenants/user-account-password.html.pasta", userAccount);
     }
 
+    private void validateOldPassword(WebContext ctx, UserAccount userAccount) {
```
tbiScireum: same as in ProfileController? can we use the same method?

src/main/java/sirius/biz/tenants/UserAccountController.java
```diff
@@ -218,6 +224,17 @@ public void password(final WebContext ctx, String id) {
...
+        String oldPassword = ctx.get(PARAM_OLD_PASSWORD).asString();
+        UserInfo userInfo = UserContext.get()
+                                       .getUserManager()
+                                       .findUserByCredentials(ctx, userAccount.getLogin().getUsername(), oldPassword);
+
+        if (userInfo == null || userInfo.as(UserAccount.class).getId() != userAccount.getId()) {
+            throw Exceptions.createHandled().withNLSKey("ProfileController.invalidOldPassword").handle();
```
andyHa: bad i18n key?

src/main/java/sirius/biz/tenants/ProfileController.java
```diff
@@ -79,7 +79,7 @@ public void profileChangePassword(WebContext ctx) {
         ctx.respondWith().template("/templates/tenants/profile-change-password.html.pasta", userAccount);
     }
 
-    private void validateOldPassword(WebContext ctx, UserAccount userAccount) {
+    public static void validateOldPassword(WebContext ctx, UserAccount userAccount) {
```
andyHa: javadoc

src/main/java/sirius/biz/tenants/ProfileController.java
```diff
@@ -79,7 +79,13 @@ public void profileChangePassword(WebContext ctx) {
...
     }
 
-    private void validateOldPassword(WebContext ctx, UserAccount userAccount) {
+    /**
+     * Validates the old password for the given {@link UserAccount}.
+     *
+     * @param ctx         the current request to read the old password from
+     * @param userAccount the user account to validate the old password for
+     */
+    public static void validateOldPassword(WebContext ctx, UserAccount userAccount) {
```
andyHa: nein, das PW als parameter übergeben

src/main/java/sirius/biz/tenants/ProfileController.java
```diff
@@ -79,7 +79,13 @@ public void profileChangePassword(WebContext ctx) {
         ctx.respondWith().template("/templates/tenants/profile-change-password.html.pasta", userAccount);
     }
 
-    private void validateOldPassword(WebContext ctx, UserAccount userAccount) {
+    /**
+     * Validates the old password for the given {@link UserAccount}.
+     *
+     * @param ctx         the current request to read the old password from
+     * @param userAccount the user account to validate the old password for
```
andyHa: +@throws HandledException in case the given password is incorrect oder so

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -454,20 +454,33 @@ public UserInfo findUserByCredentials(@Nullable WebContext ctx, String user, Str
...
     }
 
+    /**
+     * Validates the password of the given {@link UserAccount}.
+     *
+     * @param userAccount the user account to validate the password for
+     * @param password    the password to validate
+     * @return <tt>true</tt> if the password is valid, <tt>false</tt> otherwise
+     */
+    public boolean validatePassword(UserAccount userAccount, String password) {
```
andyHa: wieso schieben wir eig nicht die ganze function in LoginData? hashPassword vllt sogar non static?

PR #153, opened 2018-06-25
Tenant aware check fixes

fkrone:


src/main/java/sirius/biz/jdbc/tenants/Tenants.java
```diff
@@ -147,10 +148,12 @@ public void assertTenantOrParentTenant(TenantAware tenantAware) {
             return;
         }
 
-        if (!Objects.equals(tenantAware.getTenantAsString(), getCurrentTenant().map(Tenant::getUniqueName).orElse(null))
```
andyHa: couldn't / shouldn't getTenantAsString return the uniqueName instead? As we use it everywhere in the TenantUserManager

src/main/java/sirius/biz/jdbc/tenants/SQLTenantAware.java
```diff
@@ -34,7 +35,7 @@
 
     @Override
     public String getTenantAsString() {
-        return getTenant().getUniqueObjectName();
+        return Value.of(getTenant().getId()).getString();
```
andyHa: String.valueOf(getTenant().getId()) ?

PR #163, opened 2018-07-02
Aha/searchable entity

andyHa:


src/main/java/sirius/biz/jdbc/tenants/TenantController.java
```diff
@@ -195,7 +196,7 @@ public void tenantConfig(WebContext ctx, String tenantId) {
     public void deleteTenant(WebContext ctx, String tenantId) {
         Optional<Tenant> t = oma.find(Tenant.class, tenantId);
         if (t.isPresent()) {
-            if (t.get().getId() == tenants.getRequiredTenant().getId()) {
+            if (!Objects.equals(t.get().getId(), tenants.getRequiredTenant().getId())) {
```
fkrone: Entity equal?

src/main/java/sirius/biz/elastic/SearchableEntity.java
```diff
@@ -0,0 +1,153 @@
...
+            return;
+        }
+
+        StringBuilder contentBuilder = new StringBuilder();
+        this.getDescriptor()
+            .getProperties()
+            .stream()
+            .filter(p -> p.getAnnotation(SearchContent.class).isPresent())
+            .map(p -> p.getValue(this))
+            .filter(Objects::nonNull)
```
fkrone: Should we not filter for Strings::isFilled here?

PR #164, opened 2018-07-02
Autocomplete in storage always with wildcard

fkrone:
Autocompletes for the storage engine should search with wildcard at the end for easier autocompletion. Only the first part of an stored object is now needed instead of the complete query. This is also used in the list for the objects.

Tags: storage, autocomplete

### DIFF
--- a/src/main/java/sirius/biz/jdbc/storage/StorageController.java
+++ b/src/main/java/sirius/biz/jdbc/storage/StorageController.java
@@ -86,7 +86,7 @@ public class StorageController extends BizController {
                                                  .ne(VirtualObject.PATH, null)
                                                  .orderDesc(VirtualObject.TRACE.inner(TraceData.CHANGED_AT));
 
-        applyQuery(ctx.get("query").asString(), baseQuery, bucket.isAlwaysUseLikeSearch());
+        applyQuery(ctx.get("query").asString(), baseQuery, bucket.isAlwaysUseLikeSearch(), false);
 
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/jdbc/storage/StorageController.java
```diff
@@ -167,14 +167,12 @@ private void applyQuery(String query, SmartQuery<VirtualObject> baseQuery, boole
...
             queryString = "/" + queryString;
         }
-
-        if (queryString.contains("*")) {
-            baseQuery.where(OMA.FILTERS.or(OMA.FILTERS.like(VirtualObject.PATH).matches(queryString).build(),
-                                           OMA.FILTERS.like(VirtualObject.OBJECT_KEY).matches(query).build()));
-        } else {
-            baseQuery.where(OMA.FILTERS.or(OMA.FILTERS.eq(VirtualObject.PATH, queryString),
-                                           OMA.FILTERS.eq(VirtualObject.OBJECT_KEY, query)));
+        if (!queryString.contains("*")) {
```
tbiScireum: this might also change the behaviour outside of the autocomplete ?

andyHa: hatten wir nicht prefix gesagt oder startsWith oder wie das heisst

andyHa: ja nur beim autocomplete und da startWith oder prefix oder wie das hiess

PR #166, opened 2018-07-05
Only send generated passwords to users with an unique email

jmuscireum:
Also refactor the code for checking if password generation is possible


src/main/java/sirius/biz/jdbc/model/LoginData.java
```diff
@@ -207,6 +207,15 @@ public boolean checkPassword(String password, String defaultSalt) {
         return givenPasswordHash.equals(passwordHash);
     }
 
+    /**
+     * Determines if the current user is able to generate the password for <tt>this</tt> user.
+     *
+     * @return <tt>true</tt> if the current user can generate a password, <tt>false</tt> otherwise
+     */
+    public boolean isPasswordGenerationPossible() {
+        return !Strings.areEqual(username, UserContext.getCurrentUser().getUserName());
```
andyHa: nope

andyHa: move to UserAccount and check properly

PR #167, opened 2018-07-06
 Hide generated passwords after x amount of time

jmuscireum:
bildschirmfoto 2018-07-06 um 08 45 05

### DIFF
--- a/src/main/java/sirius/biz/jdbc/model/LoginData.java
+++ b/src/main/java/sirius/biz/jdbc/model/LoginData.java
@@ -170,7 +170,7 @@ public class LoginData extends Composite {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/biz/jdbc/model/LoginData.java
```diff
@@ -274,6 +286,30 @@ public void setLastExternalLogin(LocalDateTime lastExternalLogin) {
         this.lastExternalLogin = lastExternalLogin;
     }
 
+    public LocalDateTime getLastPasswordChange() {
+        return lastPasswordChange;
+    }
+
+    /**
+     * Determines if the generated password should ne displayed.
```
tbiScireum: should be

PR #169, opened 2018-07-09
Makes the AuditLog i18n aware and extends logging.

andyHa:
We now also record the "causing" user to that it is clear
who caused what.


src/main/java/sirius/biz/protocol/AuditLog.java
```diff
@@ -89,6 +88,33 @@ public AuditLogBuilder forUser(@Nullable String id, @Nullable String name) {
...
 
+        /**
+         * Marks the entry as caused by the current user in {@link UserContext#getCurrentUser()}.
+         *
+         * @return the builder for fluent method calls
+         */
+        @CheckReturnValue
+        public AuditLogBuilder causedByCurrentUser() {
+            UserInfo user = UserContext.getCurrentUser();
+            return causedByUser(user.getUserId(), user.getUserName()).forTenant(user.getTenantId(),
```
fkrone: so this could overwrite the tenant set in forcurrentuser? isn't this wrong as the tenant should be the tenant of the user to log for?

src/main/resources/biz_de.properties
```diff
@@ -1,268 +1,283 @@
...
-UserAccountController.noUserFoundForEmail = Für die angegebene eMail-Adresse wurde kein Benutzer gefunden.
-UserAccountController.tooManyUsersFoundForEmail = Die angegebene eMail-Adresse ist nicht eindeutig.
-VirtualObject.download = Herunterladen
-VirtualObject.fileSize = Größe
-VirtualObject.physicalKey = Physikalischer Schlüssel
-VirtualObject.rename = Umbenennen
-VirtualObject.replaceContents = Inhalt ersetzen
-VirtualObject.unreference = Referenzierung löschen
-VirtualObject.versions = Versionen
+AddressData.badZip=Fehler in '${name}': Die Postleitzahl '${zip}' ist f\u00FCr das gew\u00E4hlte Land ung\u00FCltig.
```
fkrone: UTF-8!

src/main/java/sirius/biz/jdbc/tenants/UserAccountController.java
```diff
@@ -218,10 +217,13 @@ public void generatePassword(final WebContext ctx, String id) {
...
-        auditLog.neutral("%s generated a new password for %s (%s)",
-                         UserContext.getCurrentUser().getUserName(),
-                         userAccount.toString(),
-                         userAccount.getUniqueName()).forCurrentUser().log();
+        auditLog.neutral("AuditLog.passwordGenerated")
+                .causedByCurrentUser()
+                .forUser(userAccount.getUniqueName(), userAccount.getLogin().getUsername())
+                .forTenant(String.valueOf(userAccount.getTenant().getId()),
+                           userAccount.getTenant().getValue().getName())
+                .forCurrentUser()
```
fkrone: This overwrites previously setted values

src/main/resources/biz_de.properties
```diff
@@ -1,48 +1,49 @@
-AddressData.badZip = Fehler in '${name}': Die Postleitzahl '${zip}' ist für das gewählte Land ungültig.
+AddressData.badZip = Fehler in '${name}': Die Postleitzahl '${zip}' ist f\u00FCr das gew\u00E4hlte Land ung\u00FCltig.
```
fkrone: UTF-8!

PR #171, opened 2018-07-13
Filter for unset acquired datetime

Tobsucht:


src/main/java/sirius/biz/locks/Locks.java
```diff
@@ -116,7 +117,10 @@ public void gather(MetricsCollector collector) {
         collector.metric("locks-count", "Active Locks", locks.size(), null);
         collector.metric("locks-long-running",
                          "Long locks",
-                         locks.stream().filter(l -> l.getAcquired().isBefore(limitForAcquired)).count(),
+                         locks.stream()
```
meg1502: code is a little bit cleaner if stream is mapped before the filtering

```
locks.stream().map(LockInfo::getAcquired)
                              .filter(Objects::nonNull)
                              .filter(date -> date.isBefore(limitForAcquired))
                              .count()
```

src/main/java/sirius/biz/locks/Locks.java
```diff
@@ -116,7 +117,10 @@ public void gather(MetricsCollector collector) {
...
+                              .filter(lock -> Objects.nonNull(lock.getAcquired()))
```
andyHa: isn't that rather a bug when acquired is null?

Tobsucht:
This is a bug in sirius-db -> the null objects should not be created

PR #178, opened 2018-09-24
Aha/login stability

andyHa:


andyHa:
Nope, thats what the version check is for...

PR #188, opened 2018-10-16
Allows to recalculate the searchPrefixes field

D3nnisH:
If an Entity becomes a `PrefixSearchableEntity` or if some fields of an `PrefixSearchableEntity` get or lose the `@PrefixSearchContent` annotation, the `searchPrefixes` field is not automatically updated. The new `setForceUpdateSearchPrefixes` method allows to create a command that forces to recalculate the `searchPrefixes` field manually after the patch. Furthermore, the `searchPrefixes` field is created if it didn't exist already.

### DIFF
--- a/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
+++ b/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
@@ -12,6 +12,7 @@ import sirius.biz.protocol.NoJournal;
 import sirius.db.mixing.Mapping;
 import sirius.db.mixing.annotations.BeforeSave;
 import sirius.db.mixing.annotations.Index;
+import sirius.db.mixing.annotations.Transient;
 import sirius.db.mixing.types.NestedList;
 import sirius.db.mixing.types.StringList;
 import sirius.db.mixing.types.StringListMap;
@@ -39,10 +40,15 @@ public abstract class PrefixSearchableEntity extends MongoEntity {
     @NoJournal
     private final StringList searchPrefixes = new StringList();
 
+    @Transient
+    private boolean forceUpdateSearchPrefixes;
+
     @BeforeSave
     protected void updateSearchField() {
         if (!isNew()
-            && getDescriptor().isFetched(this, getDescriptor().getProperty(SEARCH_PREFIXES))
+            && !forceUpdateSearchPrefixes
+            && getDescriptor().isFetched(this,
+                                         getDescriptor().getProperty(SEARCH_PREFIXES))
             && !isAnyMappingChanged()) {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -110,4 +118,20 @@ private void appendSingleToken(String subToken) {
     public StringList getSearchPrefixes() {
         return searchPrefixes;
     }
+
+    /**
+     * Returns whether the search prefixes will be definitely recalculated on the next {@link BeforeSave} event
+     *
+     * @return whether the search prefixes will be definitely recalculated on the next {@link BeforeSave} event
+     */
+    public boolean isForceUpdateSearchPrefixes() {
```
andyHa: why?

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -110,4 +118,20 @@ private void appendSingleToken(String subToken) {
...
+        return forceUpdateSearchPrefixes;
+    }
+
+    /**
+     * Forces the search prefixes to be recalculated on the next {@link BeforeSave} event
+     */
+    public void setForceUpdateSearchPrefixes(boolean forceUpdateSearchPrefixes) {
```
andyHa: why not just "forceUpdateOfSearchPrefixes"

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -39,9 +40,16 @@
...
 
+    @Transient
+    private boolean forceUpdateSearchPrefixes;
+
     @BeforeSave
     protected void updateSearchField() {
-        if (!isNew() && !isAnyMappingChanged()) {
+        if (!isNew()
+            && !forceUpdateSearchPrefixes
+            && getDescriptor().isFetched(this,
```
andyHa: why?

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
D3nnisH: in TraceData gibts auch nen getter:
https://github.com/scireum/sirius-biz/blob/91cb8324fee3db683c3cdb5d2e360fea4d28b064/src/main/java/sirius/biz/protocol/TraceData.java#L101

aber stimmt, is unnötig

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
D3nnisH: > Furthermore, the searchPrefixes field is created if it didn't exist already.

andyHa: but that hasn't todo anything with fetching it in a query?!

D3nnisH: it only checks if the field is present or not

PR #189, opened 2018-10-18
Allows to set the base url per tenant instead of only once per product

tbiScireum:
Backend users of the tenant should not always the the product url but
their own url
- Fixes: SE-4956

### DIFF
--- a/src/main/java/sirius/biz/web/BizController.java
+++ b/src/main/java/sirius/biz/web/BizController.java
@@ -141,6 +141,11 @@ public class BizController extends BasicController {
                 LOG.WARN("product.baseUrl is not filled. Please update the system configuration!");
             }
         }
+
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -141,6 +141,11 @@ protected String getBaseUrl() {
                 LOG.WARN("product.baseUrl is not filled. Please update the system configuration!");
             }
         }
+
+        if (UserContext.getSettings().getConfig().hasPath("tenant.baseUrl")) {
+            return UserContext.getSettings().get("tenant.baseUrl").asString();
```
andyHa:
```suggestion
            return UserContext.getSettings().get("tenant.baseUrl").asString(baseUrl);
```

and probably skip if

PR #197, opened 2018-10-31
Aha/jobz

andyHa:


src/main/java/sirius/biz/process/ProcessContext.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.process;
+
+import sirius.kernel.async.TaskContextAdapter;
+
+public interface ProcessContext extends TaskContextAdapter {
```
tbiScireum: do we really need this one as an interface? 
are there several different implementations planed?

src/main/java/sirius/biz/process/ProcessController.java
```diff
@@ -0,0 +1,99 @@
...
+import sirius.web.security.UserInfo;
+
+@Register(classes = Controller.class)
+public class ProcessController extends BizController {
+    //TODO
+    private static final String PERMISSION_MANAGE_PROCESSES = "manage-processes";
+
+    @Routed("/ps")
+    @LoginRequired
+    public void processes(WebContext ctx) {
```
tbiScireum: is this something like /system/tasks ?

src/main/java/sirius/biz/process/ProcessEnvironment.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    @Override
+    public void warn(Object message) {
+        log(ProcessLog.warn(NLS.toUserString(message)));
+    }
+
+    @Override
+    public void error(Object message) {
```
tbiScireum:
```suggestion
    public void error(String message, Object... objects) {
       // use Strings.apply(message, objects) 
```

tbiScireum: ?

PR #199, opened 2018-11-06
Adds a small test for RedisUserMessageCache

meg1502:


src/main/java/sirius/biz/cluster/RedisUserMessageCache.java
```diff
@@ -33,7 +33,7 @@
     private static final long DEFAULT_TTL = Duration.ofMinutes(1).getSeconds();
 
     @Part
-    private Redis redis;
+    private static Redis redis;
```
andyHa: nope

PR #202, opened 2018-11-06
Implements versioned files for sirius biz

tbiScireum:
Enables us to version e.g. javascript written jobs
- Fixes: SE-4561


src/main/java/sirius/biz/storage/VersionedFile.java
```diff
@@ -0,0 +1,98 @@
...
+import sirius.db.mixing.annotations.NullAllowed;
+import sirius.db.mixing.annotations.Trim;
+import sirius.kernel.commons.Explain;
+
+import java.time.LocalDateTime;
+
+/**
+ * Entity holding meta information about a versioned file.
+ */
+@Explain("No need to override the equals method. Mapping and property can have the same name, this is sirius standard.")
```
andyHa: check settings - this should be suppressed anyway

src/main/java/sirius/biz/storage/VersionedFile.java
```diff
@@ -0,0 +1,95 @@
...
+import sirius.db.mixing.annotations.Lob;
+import sirius.db.mixing.annotations.NullAllowed;
+import sirius.db.mixing.annotations.Trim;
+
+import java.time.LocalDateTime;
+
+/**
+ * Entity holding meta information about a versioned file.
+ */
+public class VersionedFile extends SQLTenantAware {
```
fkrone: check via storage framework if active?

tbiScireum: done

andyHa: still add a @Framework(XXX) here so that the table isn't created if the framework isn't active.

also - shoudln't we move all this into a sub-package "versions" to not confuse it with VirtualObjectVersion(s)

PR #217, opened 2018-11-20
Provides an SQLEntity autocomplete Parameter for jobs

tbiScireum:
A generic way to create sql entity parameters


src/main/java/sirius/biz/jobs/params/SQLEntityParameter.java
```diff
@@ -0,0 +1,45 @@
+package sirius.biz.jobs.params;
```
tbiScireum: no longer needed?

src/main/resources/default/templates/jobs/params/entity-autocomplete.html.pasta
```diff
@@ -0,0 +1,19 @@
+<i:arg type="sirius.biz.jobs.params.Parameter" name="param"/>
+<i:arg type="Map" name="context"/>
+<i:local name="param" value="@param.as(sirius.biz.jobs.params.EntityParameter.class)"/>
```
tbiScireum: use other name. this generates a tagliatelle warning

PR #220, opened 2018-11-26
Bugfix for the fetchProcessForUser Method

tbiScireum:
Also adds the links to menu


src/main/resources/default/templates/wondergem/menu-right-user.html.pasta
```diff
@@ -6,6 +6,8 @@
 
     <w:menuSection framework="biz.storage">
         <w:menuItem titleKey="StorageController.buckets" url="/storage" permission="permission-manage-files"/>
+        <w:menuItem titleKey="JobsController.jobs" url="/jobs" permission="permission-execute-jobs"/>
```
andyHa: also check if framewörks are enabled using w:framework

PR #222, opened 2018-11-27
Implements a file upload parameter

tbiScireum:
Contains a selection of current files and an upload link


src/main/resources/default/templates/jobs/params/virtual-object-autocomplete-script.html.pasta
```diff
@@ -0,0 +1,50 @@
+<script>
```
andyHa: can't we include this in the parameter page directly?

tbiScireum: we include this only once. If we move this to the parameter page this would be included for every file parameter

PR #228, opened 2018-12-03
Increase lengths for by fields in TraceData

fkrone:
As LoginData allows an username of 150 characters the TraceData should be able to store this length, too.

Tags: tracing


src/main/java/sirius/biz/protocol/TraceData.java
```diff
@@ -56,7 +57,7 @@
     public static final Mapping CHANGED_BY = Mapping.named("changedBy");
     @NoJournal
     @NullAllowed
-    @Length(50)
+    @Length(150)
```
andyHa: why?

PR #233, opened 2018-12-18
Allow to set the external address to announce in FTP passive mode

jmuscireum:
This is needed for dockerization as the FTP server has to announce the correct address in passive mode for clients to connect to.

### DIFF
--- a/src/main/java/sirius/biz/vfs/ftp/FTPBridge.java
+++ b/src/main/java/sirius/biz/vfs/ftp/FTPBridge.java
@@ -45,6 +45,9 @@ public class FTPBridge {
     @ConfigValue("vfs.ftp.bindAddress")
     private String bindAddress;
 
+    @ConfigValue("vfs.ftp.passiveExternalAddress")
+    private String passiveExternalAddress;
+
     @ConfigValue("vfs.ftp.idleTimeout")
     private Duration idleTimeout;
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/vfs/ftp/FTPBridge.java
```diff
@@ -131,14 +134,18 @@ private void setupNetwork(ListenerFactory factory) {
...
+        if (Strings.isFilled(passivePorts)) {
             dataConnectionConfigurationFactory.setPassivePorts(passivePorts);
+        }
 
-            factory.setDataConnectionConfiguration(dataConnectionConfigurationFactory.createDataConnectionConfiguration());
+        if (Strings.isFilled(passiveExternalAddress)) {
+            dataConnectionConfigurationFactory.setPassiveExternalAddress(passiveExternalAddress);
         }
+
+        factory.setDataConnectionConfiguration(dataConnectionConfigurationFactory.createDataConnectionConfiguration());
```
tbiScireum: is this a bug or why is this called even if passivePorts isEmpty?

jmuscireum: Because passiveExternalAddress could've been set too. Calling this without setting passivePorts nor passiveExternalAddress has the same effect as not calling it. It will simply use the defaults.

PR #237, opened 2019-01-11
Implements a ip range check for all users of a specific tenant

tbiScireum:
Will remove any role which was not configured to keep if the users ip does not
match the configured ip range.

Config Values to set within the tenant edit page:
    the ip ranges to check and the roles to keep
 
- Fixes: SE-5247


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +297,77 @@ public UserInfo findUserByUserId(String accountId) {
         return asUser(account, null);
     }
 
+    @Nonnull
+    @Override
+    public UserInfo bindToRequest(@Nonnull WebContext ctx) {
+        UserInfo info = super.bindToRequest(ctx);
+        return verifyIpRange(ctx, info);
```
andyHa: why not GenericUserManager?

src/main/java/sirius/biz/tenants/Tenant.java
```diff
@@ -152,6 +160,42 @@
...
+     */
+    public static final Mapping IP_RANGE = Mapping.named("ipRange");
+    @Trim
+    @Autoloaded
+    @NullAllowed
+    @Length(255)
+    private String ipRange;
+
+    /**
+     * The parsed ip ranges.
```
andyHa: Contains the...
```suggestion
     * Contains the parsed ip ranges.
```

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +294,65 @@ public UserInfo findUserByUserId(String accountId) {
...
+    /**
+     * Removes all roles from the given user other than the roles to keep.
+     * <p>
+     * Will only allow the user to keep roles defined in rolesToKeep. Will not give the user any other role.
+     *
+     * @param info        the user info to modify
+     * @param rolesToKeep the roles not to remove
+     * @return the modified user info
+     */
+    private UserInfo removeRoles(UserInfo info, Set<String> rolesToKeep) {
```
andyHa:
```suggestion
    private UserInfo createUserWithLimitedRoles(UserInfo info, Set<String> rolesToKeep) {
```

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +294,65 @@ public UserInfo findUserByUserId(String accountId) {
...
+
+        UserInfo realUser = findUserByUserId(actualUser);
+
+        if (realUser == null) {
+            return UserInfo.NOBODY;
+        }
+
+        return realUser.tryAs(Tenant.class)
+                       .filter(tenant -> !tenant.matchesIPRange(ctx))
+                       .map(tenant -> removeRoles(info, tenant.getRolesToKeepAsSet()))
```
andyHa:
```suggestion
                       .map(tenant -> createUserWithLimitedRoles(info, tenant.getRolesToKeepAsSet()))
```

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +294,65 @@ public UserInfo findUserByUserId(String accountId) {
...
+     *
+     * @param info        the user info to modify
+     * @param rolesToKeep the roles not to remove
+     * @return the modified user info
+     */
+    private UserInfo removeRoles(UserInfo info, Set<String> rolesToKeep) {
+        Set<String> oldRoles = info.getPermissions();
+
+        Set<String> roles = new HashSet<>();
+        roles.add("flag-logged-in");
```
andyHa: use constants

src/main/java/sirius/biz/tenants/Tenant.java
```diff
@@ -185,6 +229,49 @@ protected void onModify() {
...
+    public boolean matchesIPRange(WebContext ctx) {
+        if (Strings.isEmpty(ipRange)) {
+            return true;
+        }
+
+        if (rangeSet == null) {
+            try {
+                rangeSet = IPRange.paraseRangeSet(ipRange);
+            } catch (IllegalArgumentException e) {
+                // if an invalid range was configured we can not remove any permission
```
andyHa: Exceptions.ignore(e);

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +300,65 @@ public UserInfo findUserByUserId(String accountId) {
...
+     * ip range.
+     * <p>
+     * This is a security feature to prevent unwanted access to certain features from anywhere but a valid ip range.
+     *
+     * @param ctx  the current request
+     * @param info the current user
+     * @return the current user but possibly with less permissions
+     */
+    private UserInfo verifyIpRange(WebContext ctx, UserInfo info) {
+        String actualUser = ctx.getSessionValue("default-user-id").asString();
```
andyHa: this.scope.getScopeId() + "-user-id"

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -293,6 +300,65 @@ public UserInfo findUserByUserId(String accountId) {
...
+
+        UserInfo realUser = findUserByUserId(actualUser);
```
andyHa: why not fetchAccount(..).getTenant().getValue().... with appropriate null checks

andyHa: this should all be served from the cache

tbiScireum: findUserByUserId calls fetchAccount(accountId, null).
tryAs should use the cached tenant if the tenant already exists within the cache

PR #243, opened 2019-01-30
Database independent tenants

andyHa:


sabieber:
General codestyle issues i noticed:
- Variable names like `cl` and `cle` are non descriptive, something like `codeList` and `entry` would be better suited
- This is subjective, but I would prefer it if the abstract methods would all be grouped above or below the class methods, to give a better overview of whats external code.

PR #245, opened 2019-01-30
Aha/mongo tenants

andyHa:


src/main/java/sirius/biz/mongo/MongoBizEntity.java
```diff
@@ -0,0 +1,47 @@
...
+    public TraceData getTrace() {
+        return trace;
+    }
+
+    /**
+     * Checks whether any {@link Mapping} (except {@link NoJournal no journal data}) of the current {@link MongoBizEntity} changed.
+     *
+     * @return <tt>true</tt> if at least one column was changed, <tt>false</tt> otherwise.
+     */
+    public boolean isAnyColumnChangedExceptNoJournal() {
```
Tobsucht: Doens't seem to be implementation specific. can we move this to a parent class for every db-implementation?

andyHa: difficult, as their base classes meet in sirius-db where NoJournal is undefined - the only other option would be an interface with default method - as we do not expect many more of those entity classes, two methods isn't that bad anyway..

PR #250, opened 2019-01-30
Aha/cleanup

andyHa:


src/main/java/sirius/biz/jobs/params/EntityParameter.java
```diff
@@ -88,7 +88,7 @@ protected EntityParameter(String name) {
 
     @SuppressWarnings("unchecked")
```
Tobsucht: suppresswarnings can be removed

PR #254, opened 2019-02-11
Prevents the IP address-range check for anonymous users and also retu…

andyHa:
…rns the correct user and not "NOBODY"

Otherwise the public granted roles (security.scopes.default.publicRoles) would be lost.

Fixes: SIRI-32


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -319,7 +319,12 @@ public UserInfo findUserByUserId(String accountId) {
     @Override
     public UserInfo bindToRequest(@Nonnull WebContext ctx) {
         UserInfo info = super.bindToRequest(ctx);
-        return verifyIpRange(ctx, info);
+        if (info.isLoggedIn()) {
+            // We only need to verify the IP range for users being logged in....
```
welzfabian: I am a little bit confused:.
in the comment is "being logged in" but you check on "info.isLoggedIn()"
are we checking if the user is being logged in in this request or if the user is already logged in?

PR #281, opened 2019-02-28
Moves all templates into a "biz" sub-folder to prevent name conflicts.

andyHa:


src/main/java/sirius/biz/tenants/ProfileController.java
```diff
@@ -103,7 +103,7 @@ public void profileChangePassword(WebContext ctx) {
             }
         }
 
-        ctx.respondWith().template("/templates/tenants/profile-change-password.html.pasta", userAccount);
+        ctx.respondWith().template("/templates/biz/biz/tenants/profile-change-password.html.pasta", userAccount);
```
andyHa: OMG

sabieber: dublicate biz

PR #296, opened 2019-03-12
Adds the possibility to merge a dictionary into an other one.

PhilHup:
Adds the possibility to merge a dictionary into an other one. Before now we could not read or add the aliases of a dictionary to an other one.

-Tags: Feature


src/main/java/sirius/biz/jobs/batch/file/LineBasedImportJob.java
```diff
@@ -53,6 +54,15 @@ public LineBasedImportJob(VirtualObjectParameter fileParameter, Class<E> type, P
...
         this.type = type;
         this.descriptor = mixing.getDescriptor(type);
+        enhanceDictionary();
+    }
+
+    /**
+     * Adds the possibility to enhance a dicitonary during the setup of the job
+     */
+    @SuppressWarnings("squid:S1186")
+    @Explain("Do nothing by default since we only need this for imports which contain more then one Entity")
```
andyHa: than one entity

PR #299, opened 2019-03-14
Aha/siri 33

andyHa:


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -506,45 +511,58 @@ public UserInfo findUserByCredentials(@Nullable WebContext ctx, String user, Str
...
+
             return result;
         }
 
-        auditLog.negative("AuditLog.loginRejected")
-                .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
-                .forTenant(String.valueOf(account.getTenant().getId()),
-                           account.getTenant().getValue().getTenantData().getName())
-                .log();
+        completeAuditLogForUser(auditLog.negative("AuditLog.loginRejected"), account);
```
fkrone: This would result in a rejection in log caused by the user. This might me confusing to some people. ("I did not try to login but why does it state here that I did so?")

src/main/java/sirius/biz/tenants/ProfileController.java
```diff
@@ -106,6 +104,48 @@ public void profileChangePassword(WebContext ctx) {
...
+    @Routed("/profile/changeFingerprint")
+    public void profileChangeFingerprint(WebContext ctx) {
+        U userAccount = getUser().getUserObject(getUserClass());
+        userAccount = userAccount.getMapper().refreshOrFail(userAccount);
+        assertNotNew(userAccount);
+
+        userAccount.getUserAccountData().getLogin().resetFingerprint();
+        userAccount.getMapper().update(userAccount);
+
+        ctx.respondWith().redirectToGet("/profile");
```
fkrone: Wouldn't it be better to redirect the user to wondergem root in this case? I seems to me that this triggers a user being logged out instantly.

PR #309, opened 2019-03-31
Adds the tenant name to the roles of the user

tbiScireum:
This enables us to create tenant specific jobs based on the permission


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -713,6 +713,7 @@ private boolean isWithinInterval(LocalDateTime dateTime, Integer requiredInterva
         roles.addAll(user.getUserAccountData().getPermissions().getPermissions());
         roles.addAll(tenant.getTenantData().getPermissions().getPermissions());
         roles.add(UserInfo.PERMISSION_LOGGED_IN);
+        roles.add(tenant.getTenantData().getName());
```
andyHa: accountNumber (?)

PR #310, opened 2019-04-01
Job for iterating over entities

fkrone:
This job can iterate over entities and process them e.g. for changing some entities or exporting them.

Tags: job, entity

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/entity/EntityBatchJob.java
+++ b/src/main/java/sirius/biz/jobs/batch/entity/EntityBatchJob.java
@@ -11,10 +11,7 @@ package sirius.biz.jobs.batch.entity;
 import sirius.biz.jobs.batch.BatchJob;
 import sirius.biz.process.ProcessContext;
 import sirius.db.mixing.BaseEntity;
-import sirius.db.mixing.BaseMapper;
-import sirius.db.mixing.Mixing;
 import sirius.db.mixing.query.Query;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/jobs/batch/entity/EntityBatchJob.java
```diff
@@ -0,0 +1,73 @@
...
+        return (Q) entityMapper.select(type);
+    }
+
+    /**
+     * Adds filters to the query to select the {@link BaseEntity entity}.
+     *
+     * @param query the query to add filters to
+     * @param <Q>   the query properties
+     */
+    protected abstract <Q extends Query<?, E, ?>> void addQueryFilters(Q query);
```
andyHa: why not "createQuery" - then we could skip the mapper thingy. also the javadoc should encourage adding an orderBy constraint

src/main/java/sirius/biz/jobs/batch/entity/EntityBatchJob.java
```diff
@@ -0,0 +1,82 @@
...
+        this.descriptor = mixing.getDescriptor(type);
+    }
+
+    @Override
+    public void execute() throws Exception {
+        createQuery().iterateAll(this::handleEntity);
+    }
+
+    private void handleEntity(E entity) {
+        if (!process.isActive()) {
```
andyHa: iterateAll does this already(?)

PR #316, opened 2019-04-08
Parameter for accepting amounts

fkrone:
Parameter for using amounts in jobs.

Tags: job, parameter, amount

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/AmountParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/AmountParameter.java
@@ -17,7 +17,7 @@ import java.util.Optional;
 /**
  * Provides a parameter which accepts {@link Amount amounts}.
  */
-public class AmountParameter extends Parameter<Amount, AmountParameter> {
+public class AmountParameter extends TextParameter<Amount, AmountParameter> {
 
     /**
      * Creates a new parameter with the given name and label.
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/biz/jobs/params/AmountParameter.java
```diff
@@ -0,0 +1,47 @@
...
+import sirius.kernel.commons.Amount;
+import sirius.kernel.commons.Value;
+import sirius.kernel.nls.NLS;
+
+import java.util.Optional;
+
+/**
+ * Provides a parameter which accepts {@link Amount amounts}.
+ */
+public class AmountParameter extends Parameter<Amount, AmountParameter> {
```
andyHa: should't we add a withSuffix(...) which adds a suffix (auto-i18nd) to the field?

src/main/java/sirius/biz/jobs/params/TextParameter.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Specifies an addon text which is shown at the end of the input field.
+     *
+     * @param addonText the addon test to show
+     * @return the parameter itself for fluent method calls
+     */
+    public P withAddonText(String addonText) {
+        this.addonText = addonText;
```
andyHa: autotranslate this?

andyHa: s.b.

src/main/java/sirius/biz/jobs/params/TextParameter.java
```diff
@@ -0,0 +1,57 @@
...
+        return self();
+    }
+
+    /**
+     * Returns the addon text shown at the end of an input text field.
+     *
+     * @return the text to show at the end of an input text field
+     */
+    public String getAddonText() {
+        return addonText;
```
andyHa: autotranslate?

PR #319, opened 2019-04-10
Improved health checks and support for bleeding out cluster members

andyHa:
Improves rolling cluster upgrades by implementing a proper health check and support bleeding out a node before restarting it.

Using /system/cluster/ready as health check (in the load balancer) a node is only considered healthy when fully started up
(and not yet bleeding out).

Bleeding out can be started via the cluster UI or an API call (which should be the common case). This will
drop the node out of load balancing (via said health check) and also prevent the node from picking up
further background work (mainly distributed task), so that it can be restarted in a save way.

Additionally /system/cluster/halted can be invoked to determine if the node is already fully bled
out (not executing any distributed tasks anymore). Note that this implies that all long running
work is represented as distributed task (which is a good idea anyway...).

### DIFF
--- a/src/main/java/sirius/biz/cluster/BackgroundInfo.java
+++ b/src/main/java/sirius/biz/cluster/BackgroundInfo.java
@@ -17,11 +17,13 @@ import java.util.Map;
  */
 public class BackgroundInfo {
     private String nodeName;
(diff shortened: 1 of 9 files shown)


andyHa:
grafik

PR #322, opened 2019-04-11
Provides some documentation and fixes some typos.

andyHa:


src/main/java/sirius/biz/analytics/reports/README.md
```diff
@@ -0,0 +1,7 @@
+# Report Helpers
+
+Provides some helper classes for create and display a table / report based on
```
sabieber:
```suggestion
Provides some helper classes for creating and displaying a table / report based on
```

PR #324, opened 2019-04-11
Provides the indicator framework.

andyHa:


src/main/java/sirius/biz/analytics/indicators/IndicatorData.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param indicator the indicator to remove
+     * @return <tt>true</tt> if the indicator was present
+     */
+    @SuppressWarnings("squid:S2250")
+    @Explain("There should only be some idicators present, so there is no performance hot spot expected.")
+    public boolean clearIndication(String indicator) {
+        return indications.remove(indicator);
+    }
+
```
meg1502: Maybe add a method 
```
public boolean updateIndication(String indicator, boolean indicatorState) {
if(indicatorState) setIndication();
else clearIndication();
}
```

As this is already used in executeIndicator and batchIndicatorCheck, and this is pretty much the standard action after the new state of a indicator has been calculated.

src/main/java/sirius/biz/analytics/indicators/IndicatorData.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Sets or clears the given indicator / indication.
+     *
+     * @param indicator the indicator to set
+     * @param newState  <tt>true</tt> to set the indicator, <tt>false</tt> to clear it
+     * @return <tt>true</tt> if the indicator was not yet present
+     */
+    @SuppressWarnings("squid:S2250")
+    @Explain("There should only be some idicators present, so there is no performance hot spot expected.")
```
sabieber:
```suggestion
    @Explain("There should only be some indicators present, so there is no performance hotspot expected.")
```

PR #328, opened 2019-04-15
Don't mark standby processes as terminated when calling markCompleted()

Tobsucht:
Standby processes should only be created once and can be used for grouped logging from this point on. It should
not be possible to terminate such a process from within an application (conscious or not).

### DIFF
--- a/src/main/java/sirius/biz/process/Processes.java
+++ b/src/main/java/sirius/biz/process/Processes.java
@@ -389,7 +389,7 @@ public class Processes {
     /**
      * Marks a process as erroneous.
      *
-     * @param processId the process to update
+     * @param processId        the process to update
      * @param debuggingEnabled determines if debugging should be enabled or disabled
      * @return <tt>true</tt> if the process was successfully modified, <tt>false</tt> otherwise
      */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -408,17 +408,20 @@ protected boolean changeDebugging(String processId, boolean debuggingEnabled) {
...
-            }
-        });
+        return modify(processId,
+                      process -> process.getState() != ProcessState.TERMINATED
+                                 && process.getState() != ProcessState.STANDBY,
+                      process -> {
+                          process.setState(ProcessState.TERMINATED);
+                          process.setCompleted(LocalDateTime.now());
+
+                          if (timings != null) {
```
andyHa: shouldn't we run this code anyway?

src/main/java/sirius/biz/process/Processes.java
```diff
@@ -318,10 +318,7 @@ protected Process fetchStandbyProcess(String type, String tenantId) {
      * @return <tt>true</tt> if the process was successfully modified, <tt>false</tt> otherwise
      */
     private boolean modify(String processId, Predicate<Process> checker, Consumer<Process> modifier) {
-        Process process = process1stLevelCache.get(processId);
-        if (process == null) {
-            process = elastic.find(Process.class, processId).orElse(null);
-        }
+        Process process = fetchProcess(processId).orElse(null);
```
andyHa: i'd think we don't want to use the 2nd level cache here as it might contain a stale instance - we only use the 1st lvl cach to bypass the 1s write delay..

PR #329, opened 2019-05-02
Allows to import zipped files

jmuscireum:


src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -42,11 +47,48 @@ protected FileImportJob(VirtualObjectParameter fileParameter, ProcessContext pro
...
+     * Performs the import for a zipped file.
+     *
+     * @param in the data to import
+     * @throws Exception in case of an error during the import
+     */
+    protected void executeForZippedFile(InputStream in) throws Exception {
+        try (ZipInputStream zipInputStream = new ZipInputStream(in)) {
+            ZipEntry entry = zipInputStream.getNextEntry();
+
+            while (entry != null && !canHandleFileExtension(Files.getFileExtension(entry.getName()))) {
```
andyHa: why not process all files?

andyHa: also add a log statement for each file

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -42,11 +47,48 @@ protected FileImportJob(VirtualObjectParameter fileParameter, ProcessContext pro
...
     }
 
+    /**
+     * Performs the import for a zipped file.
+     *
+     * @param in the data to import
+     * @throws Exception in case of an error during the import
+     */
+    protected void executeForZippedFile(InputStream in) throws Exception {
+        try (ZipInputStream zipInputStream = new ZipInputStream(in)) {
```
andyHa: notify the user that the file is being unzipped

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -42,11 +51,68 @@ protected FileImportJob(VirtualObjectParameter fileParameter, ProcessContext pro
...
         VirtualObject file = process.require(fileParameter);
+
         try (InputStream in = storage.getData(file)) {
+            if (!canHandleFileExtension(file.getFileExtension()) && "zip".equalsIgnoreCase(file.getFileExtension())) {
+                executeForZippedFile(in);
+
+                return;
+            }
+
+            if (!canHandleFileExtension(file.getFileExtension())) {
```
andyHa: combine with the if above and also without loading the data first (above storage.getData())

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -42,11 +51,68 @@ protected FileImportJob(VirtualObjectParameter fileParameter, ProcessContext pro
...
+            while (entry != null) {
+                if (!isHiddenFile(entry.getName()) && canHandleFileExtension(Files.getFileExtension(entry.getName()))) {
+                    fileFound = true;
+
+                    process.log(ProcessLog.info()
+                                          .withNLSKey("FileImportJob.importingZippedFile")
+                                          .withContext("filename", entry.getName()));
+
+                    // Copy the entry data to another stream because executeForStream will close the given stream
+                    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
andyHa: bad idea - thse can be GBs!!

src/main/java/sirius/biz/jobs/batch/file/LineBasedImportJob.java
```diff
@@ -73,9 +73,14 @@ protected void executeForStream(String filename, InputStream in) throws Exceptio
...
 
+    @Override
+    protected boolean canHandleFileExtension(String fileExtension) {
+        return "xls".equals(fileExtension) || "xlsx".equals(fileExtension) || "csv".equals(fileExtension);
+    }
+
     @Override
     public void handleRow(int index, Values row) {
-        if (aliases == null) {
+        if (index == 1) {
```
andyHa: explain the change in a separate commit!!

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -42,11 +51,68 @@ protected FileImportJob(VirtualObjectParameter fileParameter, ProcessContext pro
...
+     */
+    protected void executeForZippedFile(InputStream in) throws Exception {
+        process.log(ProcessLog.info().withNLSKey("FileImportJob.importingZipFile"));
+
+        try (ZipInputStream zipInputStream = new ZipInputStream(in)) {
+            boolean fileFound = false;
+            ZipEntry entry = zipInputStream.getNextEntry();
+
+            while (entry != null) {
+                if (!isHiddenFile(entry.getName()) && canHandleFileExtension(Files.getFileExtension(entry.getName()))) {
```
andyHa: is this hidden check necessary - wouldn't the file extensions filter be enough?

src/main/java/sirius/biz/jobs/batch/file/LineBasedImportJob.java
andyHa: and also reset the aliases before every run, there is no guarantee that eacht file uses the same format

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
jmuscireum: MacOS adds hidden files with the same name and extension to the zip

PR #330, opened 2019-05-06
Aha/multi zip

andyHa:


src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -84,9 +75,7 @@ protected void executeForZippedFile(InputStream in) throws Exception {
                                           .withNLSKey("FileImportJob.importingZippedFile")
                                           .withContext("filename", entry.getName()));
 
-                    executeForStream(entry.getName(), zipInputStream);
-
-                    return;
```
jmuscireum: The Exception below will be always thrown

jmuscireum:
This will cause problems with the `aliases == null` in LineBasedImportJob#handleRow(int, Values)

PR #333, opened 2019-05-13
Permits to ignore i5 data conversion errors.

andyHa:
This is required if the "Java side" is updated before the i5 side as most probably the
left over byte buffer is uninitialized and will crash data conversions. The application
has to ensure itself that this data is never used up until it is populated correctly.

Also this feature is turned off by default for obvious reasons.

### DIFF
--- a/src/main/java/sirius/biz/i5/Transformer.java
+++ b/src/main/java/sirius/biz/i5/Transformer.java
@@ -24,6 +24,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Helps to transform byte oriented record (sent and received from the i5) into Java objects.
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/i5/Transformer.java
```diff
@@ -152,15 +154,25 @@ private int transform(Entry<Field, Transform> e, Object object, byte[] data, int
...
-            throw Exceptions.handle()
-                            .to(I5Connector.LOG)
-                            .error(ex)
-                            .withSystemErrorMessage("Error while transforming '%s.%s' to %s: %s (%s)",
-                                                    object.getClass().getName(),
-                                                    field.getName(),
-                                                    info.targetType().getName())
-                            .handle();
+            if (info.ignoreErrors()) {
+                I5Connector.LOG.FINE("Ignoring a conversion error for %s of %s: %s (%s)",
```
fkrone: Add Exceptions.ignore so the Exception can be retrieved later for debugging purpose

PR #337, opened 2019-05-15
 Add possibility to auto load StringList Properties

welzfabian:


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -198,7 +200,11 @@ protected void load(WebContext ctx, BaseEntity<?> entity, List<Mapping> properti
             String propertyName = property.getName();
 
             try {
-                property.parseValue(entity, ctx.get(propertyName));
+                if (property instanceof StringListProperty) {
```
welzfabian: maybe a method "parseValueFromWebContext" would be better, what do you guys think?

welzfabian: decision on: `property.parseValues(entity, Values.of(ctx.getParameters(propertyName)));`
`parseValues` needs to be implemented and calls `parseValue` with the first value by default

PR #339, opened 2019-05-17
Adds the possibility to add references to a process (and to filter on…

andyHa:
… them).


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -482,10 +486,25 @@ protected boolean setStateMessage(String processId, String state) {
...
+    /**
+     * Adds the given reference to the given process.
+     *
+     * @param processId the process to update
+     * @param reference the reference to attach
+     * @return <tt>true</tt> if the process was successfully modified, <tt>false</tt> otherwise
+     */
+    public boolean addReference(String processId, String reference) {
+        return modify(processId, process -> true, process -> {
+            if (!process.getReferences().contains(reference)) {
```
welzfabian: shouldn't we do this check in the "checker" lambda? (currently `process -> true`)

andyHa: The checker is in charge of ensuring the correct state, not for enforcing business rules...

src/main/java/sirius/biz/process/ProcessController.java
```diff
@@ -90,6 +90,11 @@ public void processes(WebContext ctx) {
         ElasticPageHelper<Process> ph = ElasticPageHelper.withQuery(query);
         ph.withContext(ctx);
         ph.addTermAggregation(Process.STATE, ProcessState.class);
+        ph.addParameterFacet("reference",
```
sabieber: ph > pageHelper?

PR #345, opened 2019-05-20
Permits to select against which database a query is executed.

andyHa:


src/main/java/sirius/biz/jdbc/DatabaseController.java
```diff
@@ -63,7 +68,11 @@
     @Permission(TenantUserManager.PERMISSION_SYSTEM_TENANT)
     @Routed("/system/sql")
     public void sql(WebContext ctx) {
-        ctx.respondWith().template("templates/biz/model/sql.html.pasta");
+        // Only display selectable databases which are properly configured..
+        List<String> availableDatabases = selectableDatabases.stream()
+                                                             .filter(name -> databases.getDatabases().contains(name))
+                                                             .collect(Collectors.toList());
+        ctx.respondWith().template("templates/biz/model/sql.html.pasta", availableDatabases, defaultDatabase);
```
jmuscireum: Doesn't sql.html.pasta also have to be modified?

PR #346, opened 2019-05-21
Prevent exception when Mixin targets MongoUserAccount

welzfabian:
As it tries to assign the Mixin first to the UserAccountData which fails with an exception and therefore no other assignment is tried.

See SQLUserAccount#tryAs 48b8c462d98577ce2cc83d9f036a44bff3f82b4e


src/main/java/sirius/biz/tenants/mongo/MongoUserAccount.java
```diff
@@ -46,9 +46,11 @@
 
     @Override
     public <A> Optional<A> tryAs(Class<A> adapterType) {
-        Optional<A> result = getUserAccountData().tryAs(adapterType);
-        if (result.isPresent()) {
-            return result;
+        if (getUserAccountData().is(adapterType)) {
```
andyHa: tryAs should not require an is check...

PR #350, opened 2019-05-29
Create a wrapper for the security.packages-configurations

welzfabian:
Fixes: SIRI-75

### DIFF
--- a/src/main/java/sirius/biz/packages/Packages.java
+++ b/src/main/java/sirius/biz/packages/Packages.java
@@ -22,6 +22,7 @@ import java.util.function.Predicate;
 
 /**
  * This helper class is a wrapper for the <tt>security.packages</tt> configuratuions.
+ * <p>
  */
 @Register(classes = Packages.class)
 public class Packages {
@@ -102,16 +103,16 @@ public class Packages {
      * The predicate will be called with the required permission for the permission and should return true or false,
      * depending on if the required permission is present.
      *
-     * @param role                   the role in question
+     * @param permission             the permission in question
      * @param hasPermissionPredicate a predicate which will determine if the required permission is present
      * @return true if the object has the required permission (determined via the predicate), false if not
      */
-    public boolean hasRequiredPermissionForPermission(String role, Predicate<String> hasPermissionPredicate) {
+    public boolean hasRequiredPermissionForPermission(String permission, Predicate<String> hasPermissionPredicate) {
         Config config = Sirius.getSettings().getConfig("security.packages.required-permissions-for-permission");
-        if (!config.hasPath(role)) {
+        if (!config.hasPath(permission)) {
             return true;
         }
-        String requiredPermission = config.getString(role);
+        String requiredPermission = config.getString(permission);
         return hasPermissionPredicate.test(requiredPermission);
     }
 }
--- a/src/main/java/sirius/biz/packages/Packages.java
+++ b/src/main/java/sirius/biz/packages/Packages.java
@@ -14,6 +14,7 @@ import sirius.kernel.Sirius;
 import sirius.kernel.di.std.Register;
 import sirius.kernel.nls.NLS;
 import sirius.kernel.settings.Extension;
+import sirius.web.security.Permissions;
 import sirius.web.security.UserContext;
 import sirius.web.security.UserInfo;
 
@@ -114,6 +115,6 @@ public class Packages {
             return true;
         }
         String requiredPermission = config.getString(permission);
-        return hasPermissionPredicate.test(requiredPermission);
+        return Permissions.hasPermission(requiredPermission, hasPermissionPredicate);
     }
 }
--- a/src/main/java/sirius/biz/packages/Packages.java
+++ b/src/main/java/sirius/biz/packages/Packages.java
@@ -15,8 +15,10 @@ import sirius.kernel.di.std.Register;
 import sirius.kernel.nls.NLS;
 import sirius.kernel.settings.Extension;
 import sirius.web.security.UserContext;
+import sirius.web.security.UserInfo;
 
 import java.util.List;
+import java.util.function.Predicate;
 
 /**
  * This helper class is a wrapper for the <tt>security.packages</tt> configuratuions.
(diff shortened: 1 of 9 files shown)


src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,87 @@
...
+import sirius.kernel.settings.Extension;
+import sirius.web.security.UserContext;
+
+import java.util.List;
+
+/**
+ * This helper class is a wrapper for the <tt>security.packages</tt> configuratuions.
+ */
+@Register(classes = Packages.class)
+public class Packages {
```
tbiScireum: how do you calculate the permissions from the packages and upgrades?

src/main/resources/component-biz.conf
```diff
@@ -665,6 +665,25 @@ security {
...
+        tenant {
+            # the packages of the scope
+            packages = []
+
+            # the upgrades of the scope
+            upgrades = []
+
+            # indicates which permissions are needed to show a specific role for a user
+            # roles which have no requirements can be omitted
+            # have to be key-value pais, in which the key is the role and the value is the required permission
```
andyHa: pairs

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Returns the pricePackages for the given scope.
+     *
+     * @param scope the scope
+     * @return the pricePackages of the scope
+     */
+    public List<String> getPricePackages(String scope) {
```
andyHa: just "getPackages"?

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.List;
+
+/**
+ * This helper class is a wrapper for the <tt>security.packages</tt> configuratuions.
+ */
+@Register(classes = Packages.class)
+public class Packages {
+
+    /**
+     * Returns the pricePackages for the given scope.
```
andyHa: Returns the list of packages available for the given scope.

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
...
+import sirius.kernel.Sirius;
+import sirius.kernel.di.std.Register;
+import sirius.kernel.nls.NLS;
+import sirius.kernel.settings.Extension;
+import sirius.web.security.UserContext;
+
+import java.util.List;
+
+/**
+ * This helper class is a wrapper for the <tt>security.packages</tt> configuratuions.
```
andyHa: please add some lines of documentation of how the wohle sysgtem works and what eacht entity is.

Also add a README.md for the (java) package providing some insight.

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.tenants.packages;
```
andyHa: sirius.biz.packages?

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Checks whether the current user has the required permission for a given role.
+     *
+     * @param scope the scope
+     * @param role  the role in question
+     * @return true if the current user has the required permission, false if not
+     */
+    public boolean hasRequiredPermissionForRole(String scope, String role) {
```
andyHa: this could be also used for feature checks "hasRequiredPermission" 
-> scope isn't required here
-> role=permission
-> add a Function{String,Boolean} as parameter and a boilerplate provinding UserContext.getCurrentUser()::hasPermission to decouple them

src/main/java/sirius/biz/tenants/packages/Packages.java
```diff
@@ -0,0 +1,105 @@
...
+        Config config = getScopeExtension(scope).getConfig("required-permissions-for-role");
```
andyHa: this map can be placed outside of the scope

src/main/resources/component-biz.conf
```diff
@@ -665,6 +665,25 @@ security {
         }
     }
 
+    # defines the packages and upgrades for the different scopes
+    packages {
+
+        # the 'tenant' scope
```
andyHa: The 'tenant' scope used by the tenants framework (TenantUserManager etc.)

src/test/java/sirius/biz/tenants/packages/PackagesSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+        def scope1 = packages.getPricePackages("test-scope-1")
+        def scope2 = packages.getPricePackages("test-scope-2")
+        then:
+        scope1 == ["package1", "package2"]
+        scope2 == ["package3"]
+    }
+
+    def "test getPackageName"() {
+        when:
+        def name = packages.getPricePackageName("test-scope-1", "package1")
```
andyHa: do we really need a test to ensure that NLS.get works?

src/main/java/sirius/biz/tenants/packages/Packages.java
sabieber: that comes in one of the next PRs right? The packages and upgrades need to be defined in the security profiles and will get extrapolated to all applicable permissions

src/main/java/sirius/biz/packages/Packages.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Checks whether the permissionData object has the required permission for a given permission.
+     *
+     * @param permission     the permission in question
+     * @param permissionData the permissionData object to check
+     * @return true if the permissionData object has the required permission, false if not
+     */
+    public boolean hasRequiredPermissionForPermission(String permission, PermissionData permissionData) {
```
andyHa: rather use a Function<String, Boolean>

and also maybe transfer the permission logic from UserInfo to here and apply that before. The check features this will most probably not be called with either a user nor with a PermissionData object but rather a set of features

src/main/java/sirius/biz/packages/Packages.java
```diff
@@ -0,0 +1,119 @@
...
+     * <p>
+     * The predicate will be called with the required permission for the permission and should return true or false,
+     * depending on if the required permission is present.
+     *
+     * @param permission             the permission in question
+     * @param hasPermissionPredicate a predicate which will determine if the required permission is present
+     * @return true if the object has the required permission (determined via the predicate), false if not
+     */
+    public boolean hasRequiredPermissionForPermission(String permission, Predicate<String> hasPermissionPredicate) {
+        Config config = Sirius.getSettings().getConfig("security.packages.required-permissions-for-permission");
```
andyHa: maybe move the hasPermission logic of UserInfo to here (as separate method) and also apply it here)

PR #351, opened 2019-05-29
Miscellaneous fixes

andyHa:


src/main/java/sirius/biz/jobs/interactive/ReportJobFactory.java
```diff
@@ -25,6 +27,9 @@
  */
 public abstract class ReportJobFactory extends InteractiveJobFactory {
 
+    @Part
+    protected Cells cells;
```
sabieber: why?

andyHa: required by all subclasses to populare the report - documentation has been expanded.

PR #359, opened 2019-06-05
Fixes and additions to PackageData

welzfabian:


src/main/java/sirius/biz/packages/PackageData.java
```diff
@@ -147,6 +153,39 @@ public String getPackage() {
...
+
+    /**
+     * Checks if the permission is granted.
+     * <p>
+     * Will be checked against the {@link #getExpandedPermissions()}.
+     *
+     * @param permission the permission to check for
+     * @return true, if this obejct has the permission, false otherwise
+     */
+    public boolean hasPermission(String permission) {
```
andyHa: this is kind of dangerous as it would be quite inefficient to check repeatedly

PR #365, opened 2019-06-06
Adds shortcuts to Tenants to get userAccounts cached in the UserManager

meg1502:
& NPE fix

### DIFF
--- a/src/main/java/sirius/biz/tenants/Tenants.java
+++ b/src/main/java/sirius/biz/tenants/Tenants.java
@@ -312,6 +312,17 @@ public abstract class Tenants<I, T extends BaseEntity<I> & Tenant<I>, U extends 
     }
 
     /**
+     * Boilerplate to quickly fetch the name of the user with the given id.
+     * Uses {@link U#toString()}.
+     * 
+     * @param userId the user to fetch the name for
+     * @return the name of the user or an empty string if the user doesn't exist
+     */
+    public String fetchCachedUserName(String userId) {
+        return fetchCachedUserAccount(userId).map(UserAccount::toString).orElse("");
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/tenants/Tenants.java
```diff
@@ -284,4 +284,60 @@ public String fetchCachedTenantName(I tenantId) {
...
+
+    /**
+     * Provides access to the user account with the given id
+     * <p>
+     * This utilizes the cache maintained by the {@link TenantUserManager} and is therefore quite efficient
+     *
+     * @param userId the id of the user to fetch
+     * @return the user account with the given id or an empty optional if the user cannot be resolved
+     */
+    public Optional<U> fetchCachedUserAccount(I userId) {
```
andyHa: use a String here - also change for fetchCachedTenant - the lookup is based on strings anyway and this makes it easier when working with generic code.

src/main/java/sirius/biz/tenants/Tenants.java
```diff
@@ -284,4 +284,60 @@ public String fetchCachedTenantName(I tenantId) {
...
+        return Optional.ofNullable(getTenantUserManager().fetchAccount(String.valueOf(userId)));
+    }
+
+    /**
+     * Boilerplate to quickly fetch the email of the user with the given id.
+     *
+     * @param userId the user to fetch the email for
+     * @return the email of the user or an empty string if the user doesn't exist
+     */
+    public String fetchCachedUserMail(I userId) {
```
andyHa: maybe also add a fetchCachedUserName which outputs the person name or UserAccount.toString or the like?

PR #371, opened 2019-06-11
Attaches a flag permission to users of tenants with child tenants

sabieber:


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -74,6 +74,11 @@
      */
     public static final String PERMISSION_OUT_OF_IP_RANGE = "flag-out-of-ip-range";
 
+    /**
+     * This flag indicates that the current user belongs to a tenant with child tenants.
+     */
+    public static final String PERMISSION_HAS_CHILDREN = "flag-has-children";
```
andyHa: document in component-biz.conf in permissions map

PR #372, opened 2019-06-11
Add possibility to use a hashed apiToken to login

welzfabian:


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -493,7 +493,7 @@ public UserInfo findUserByCredentials(@Nullable WebContext ctx, String user, Str
         }
 
         LoginData loginData = account.getUserAccountData().getLogin();
-        if (acceptApiTokens && Strings.areEqual(password, loginData.getApiToken())) {
+        if (acceptApiTokens && loginData.checkApiToken(password)) {
```
andyHa: shouldn't we OR the latter?

src/main/java/sirius/biz/model/MD5ApiTokenHashFunction.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.common.base.Charsets;
+import com.google.common.hash.Hashing;
+import sirius.kernel.di.std.Register;
+
+/**
+ * Uses md5 hashing to hash the apiToken.
+ */
+@Register
+public class MD5ApiTokenHashFunction implements ApiTokenHashFunction {
```
andyHa: where's the difference to LegacyMD5ApiTokenHashFunction - can't we make that a protected method in the usermanager so that it can be overwritten on demand

src/main/java/sirius/biz/tenants/TenantUserManager.java
welzfabian: i don't think so, `acceptApiTokens` is a configuration, if apiTokens should be possible at all, so an AND seems correct to me

src/main/java/sirius/biz/model/LoginData.java
```diff
@@ -61,7 +61,6 @@
          * The given password is invalid.
          */
         INVALID
-
```
andyHa: keep this line or intellij will screw up reformatting

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -517,6 +520,33 @@ public UserInfo findUserByCredentials(@Nullable WebContext ctx, String user, Str
...
+            long timestampToCheck = currentTimestampInDays + i;
+            if (Strings.areEqual(getHashedApiToken(loginData.getApiToken(), timestampToCheck), givenApiToken)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    protected String getHashedApiToken(String apiToken, long time) {
```
andyHa: javadoc and time -> timestampInDays

PR #377, opened 2019-06-12
Some additions to tenants

welzfabian:


src/main/java/sirius/biz/tenants/Tenants.java
```diff
@@ -85,9 +85,12 @@
      */
     @Nonnull
     public U getRequiredUser() {
-        Optional<U> ua = getCurrentUser();
```
andyHa: getCurrentUser().orElseThrow(this::createMissingUserException);

src/main/java/sirius/biz/tenants/Tenants.java
```diff
@@ -85,9 +85,12 @@
...
-        if (ua.isPresent()) {
-            return ua.get();
+        return getUserOrThrowException(getCurrentUser());
+    }
+
+    private U getUserOrThrowException(Optional<U> userAccount) {
+        if (userAccount.isPresent()) {
+            return userAccount.get();
         }
         throw Exceptions.handle()
```
andyHa: private void createMissingUserException() {

PR #380, opened 2019-06-14
Save permissions of PermissionData as  StringList

welzfabian:


src/main/java/sirius/biz/model/PermissionData.java
```diff
@@ -41,14 +41,14 @@
...
     /**
-     * Contains all permissions as a single string, separated with commas.
+     * Contains all permissions.
      */
-    public static final Mapping PERMISSION_STRING = Mapping.named("permissionString");
+    public static final Mapping PERMISSIONS = Mapping.named("permissions");
     @Autoloaded
     @NullAllowed
     @AutoImport
     @Length(4096)
```
sabieber: Does this still need to be here?

welzfabian: yes, so the Column in an SQL DB knows how long it should be, because in MySQL it is still stored as an string

PR #384, opened 2019-06-14
Fixes tenant/user switching for accounts of the system tenant

sabieber:
As only accounts of the system tenant with the permission `user-administrator` get the `flag-syste-tenant` this was too restrictive for some use cases, e.g. when normal `administrator`s should also be able to support other users by switching to their tenant or account.

To enable this we check if the account is originaly from the system tenant when filtering the users/tenants and selecting them.

### DIFF
--- a/src/main/java/sirius/biz/tenants/mongo/MongoTenantController.java
+++ b/src/main/java/sirius/biz/tenants/mongo/MongoTenantController.java
@@ -58,7 +58,9 @@ public class MongoTenantController extends TenantController<String, MongoTenant,
 
     private MongoQuery<MongoTenant> queryPossibleTenants(MongoTenant currentTenant) {
         MongoQuery<MongoTenant> baseQuery = mango.select(MongoTenant.class);
-        if (!hasPermission(TenantUserManager.PERMISSION_SYSTEM_TENANT)) {
+        
+        TenantUserManager<?, ?, ?> userManager = tenants.getTenantUserManager();
+        if (!userManager.getSystemTenantId().equals(userManager.getOriginalTenantId())) {
             if (currentTenant.getTenantData().isCanAccessParent()) {
                 baseQuery.where(QueryBuilder.FILTERS.or(QueryBuilder.FILTERS.and(QueryBuilder.FILTERS.eq(Tenant.PARENT,
                                                                                                          currentTenant),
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/biz/tenants/mongo/MongoUserAccountController.java
```diff
@@ -44,7 +43,7 @@
     protected BasePageHelper<MongoUserAccount, ?, ?, ?> getSelectableUsersAsPage() {
         MongoQuery<MongoUserAccount> baseQuery = mango.select(MongoUserAccount.class);
 
-        if (!hasPermission(TenantUserManager.PERMISSION_SYSTEM_TENANT)) {
+        if (!isUserAccountOfSystemTenant()) {
```
andyHa: please be kind enough to also fix this in the SQLUserAccountController

src/main/java/sirius/biz/tenants/mongo/MongoTenantController.java
```diff
@@ -58,7 +57,8 @@
 
     private MongoQuery<MongoTenant> queryPossibleTenants(MongoTenant currentTenant) {
         MongoQuery<MongoTenant> baseQuery = mango.select(MongoTenant.class);
-        if (!hasPermission(TenantUserManager.PERMISSION_SYSTEM_TENANT)) {
+
+        if (!isUserAccountOfSystemTenant()) {
```
andyHa: please be kind enough to also fix this in the SQLTenantController

src/main/java/sirius/biz/web/BizController.java
```diff
@@ -124,6 +125,17 @@ private HandledException invalidTenantException() {
...
 
+
+    /**
+     * Checks whether the current user is an account of the system tenant. This can be useful for granting extra features.
+     *
+     * @return <tt>true</tt> if the user is an account of the system tenant, <tt>false</tt> otherwise.
+     */
+    protected boolean isUserAccountOfSystemTenant() {
+        TenantUserManager<?, ?, ?> userManager = tenants.getTenantUserManager();
+        return userManager.getSystemTenantId().equals(userManager.getOriginalTenantId());
```
andyHa:
```suggestion
        return Strings.areEqual(userManager.getSystemTenantId(), userManager.getOriginalTenantId()));
```


...you just never know...

PR #386, opened 2019-06-19
Aha/ox minifx

andyHa:


src/main/resources/component-biz.conf
```diff
@@ -654,9 +655,10 @@ security {
             permission-system-load = true
             permission-tasks = true
             permission-manage-all-processes = true
-            feature-provide-jobs = true
```
meg1502: Not Subject of this commit ?

src/main/resources/component-biz.conf
```diff
@@ -654,9 +655,10 @@ security {
...
             permission-view-rate-limits = true
             permission-delete-user-accounts = true
+
```
meg1502: is this extra line intended?

src/main/resources/component-biz.conf
andyHa: yep, this was a left-over of a pervious refactoring - the feature is long gone...

src/main/resources/component-biz.conf
andyHa: yep for readability

PR #400, opened 2019-06-25
Clears spy user and tenant session values on logout

meg1502:


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -853,6 +853,13 @@ protected String computeTenantname(@Nullable WebContext ctx, String tenantId) {
                                  .collect(Collectors.toList());
     }
 
+    @Override
+    public void logout(@Nonnull WebContext ctx) {
+        super.logout(ctx);
+        ctx.setSessionValue(scope.getScopeId() + SPY_ID_SUFFIX, null);
+        ctx.setSessionValue(scope.getScopeId() + TENANT_SPY_ID_SUFFIX, null);
```
welzfabian: why not clear the whole thing in web? `ctx.clearSession()`?

PR #409, opened 2019-06-26
Hides some AuditLogs from users.

andyHa:
Especially those where the system tenant selected / became other
users / tenants.

### DIFF
--- a/src/main/java/sirius/biz/tenants/TenantController.java
+++ b/src/main/java/sirius/biz/tenants/TenantController.java
@@ -375,7 +375,7 @@ public abstract class TenantController<I, T extends BaseEntity<I> & Tenant<I>, U
                     .hideFromUser()
                     .causedByUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
                     .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
-                    .forTenant(String.valueOf(account.getTenant().getId()),
+                    .forTenant(account.getTenant().getIdAsString(),
                                account.getTenant().getValue().getTenantData().getName())
                     .log();
 
@@ -399,7 +399,7 @@ public abstract class TenantController<I, T extends BaseEntity<I> & Tenant<I>, U
                 .hideFromUser()
                 .causedByCurrentUser()
                 .forCurrentUser()
-                .forTenant(String.valueOf(effectiveTenant.getId()), effectiveTenant.getTenantData().getName())
+                .forTenant(effectiveTenant.getIdAsString(), effectiveTenant.getTenantData().getName())
                 .log();
 
         ctx.setSessionValue(UserContext.getCurrentScope().getScopeId() + TenantUserManager.TENANT_SPY_ID_SUFFIX,
--- a/src/main/java/sirius/biz/tenants/TenantController.java
+++ b/src/main/java/sirius/biz/tenants/TenantController.java
@@ -53,7 +53,6 @@ public abstract class TenantController<I, T extends BaseEntity<I> & Tenant<I>, U
 
     /**
      * Contains the scope of packages and upgrades to use.
-     *
      */
     public static final String PACKAGE_SCOPE_TENANT = "tenant";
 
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -369,7 +370,14 @@ private String determineOriginalTenantId(WebContext ctx) {
     @Routed("/tenants/select/:1")
     public void selectTenant(final WebContext ctx, String id) {
         if ("main".equals(id) || Strings.areEqual(determineOriginalTenantId(ctx), id)) {
-            auditLog.neutral("AuditLog.switchedToMainTenant").causedByCurrentUser().forCurrentUser().log();
+            UserAccount<?, ?> account = UserContext.getCurrentUser().as(UserAccount.class);
+            auditLog.neutral("AuditLog.switchedToMainTenant")
+                    .hideFromUser()
+                    .causedByUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
+                    .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
+                    .forTenant(String.valueOf(account.getTenant().getId()),
```
meg1502: getIdAsString ?

src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -369,7 +370,14 @@ private String determineOriginalTenantId(WebContext ctx) {
...
+                               account.getTenant().getValue().getTenantData().getName())
```
meg1502: use cache -> tenants.fetchCachedTenantName ?

src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -386,10 +394,17 @@ public void selectTenant(final WebContext ctx, String id) {
             return;
         }
 
-        auditLog.neutral("AuditLog.selectedTenant").causedByCurrentUser().forCurrentUser().log();
+        T effectiveTenant = tenant.get();
+        auditLog.neutral("AuditLog.selectedTenant")
+                .hideFromUser()
+                .causedByCurrentUser()
+                .forCurrentUser()
+                .forTenant(String.valueOf(effectiveTenant.getId()), effectiveTenant.getTenantData().getName())
```
meg1502: same as above

src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -369,7 +370,14 @@ private String determineOriginalTenantId(WebContext ctx) {
     @Routed("/tenants/select/:1")
     public void selectTenant(final WebContext ctx, String id) {
         if ("main".equals(id) || Strings.areEqual(determineOriginalTenantId(ctx), id)) {
-            auditLog.neutral("AuditLog.switchedToMainTenant").causedByCurrentUser().forCurrentUser().log();
+            UserAccount<?, ?> account = UserContext.getCurrentUser().as(UserAccount.class);
+            auditLog.neutral("AuditLog.switchedToMainTenant")
+                    .hideFromUser()
+                    .causedByUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
+                    .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
```
welzfabian: why not `forCurrentUser()` and `causedByCurrentUser()`?

src/main/java/sirius/biz/tenants/TenantController.java
andyHa: the tenant in getValue is already populated from the cache...

src/main/java/sirius/biz/tenants/TenantController.java
andyHa: because the user is a malformed beast here which sill contains the wrong tenant...

welzfabian: `UserContext.getCurrentUser().as(UserAccount.class).getTenant().getTenantData().getName()` should be exactly the same thing as `user.getTenantName()`, because the beast with the wrong tenant also sets the tenantName of the UserInfo, or am I missing something?

`UserContext.getCurrentUser().as(UserAccount.class)` is also returning the UserAccount with the "not original" tenant, or am I wrong?

PR #411, opened 2019-06-26
Grants an additional flag ("flag-system-tenant-member") to all member…

andyHa:
…s of the system tenant.


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -788,11 +797,16 @@ private boolean isWithinInterval(LocalDateTime dateTime, Integer requiredInterva
         }
 
         Set<String> transformedRoles = transformRoles(roles);
-        if (isSystemTenant && transformedRoles.contains(PERMISSION_MANAGE_SYSTEM)) {
-            roles.add(PERMISSION_SYSTEM_TENANT);
+        roles.add(PERMISSION_SYSTEM_TENANT_MEMBER);
```
sabieber: this must be inside of the isSystemTenant if right?

PR #427, opened 2019-06-28
Set the value of the Facet for boolean-aggregations correctly

welzfabian:


src/main/java/sirius/biz/web/MongoPageHelper.java
```diff
@@ -140,9 +141,11 @@ protected MongoPageHelper(MongoQuery<E> query) {
      * @return the helper itself for fluent method calls
      */
     public MongoPageHelper<E> addBooleanAggregation(Mapping field) {
+        Objects.requireNonNull(ctx);
+
         Facet facet = new Facet(baseQuery.getDescriptor().findProperty(field.toString()).getLabel(),
                                 field.toString(),
-                                null,
+                                ctx.get(field.toString()).asString(),
```
andyHa: this isn't needed as BasePageHelper#addFacet will do this for you (L118)

PR #446, opened 2019-07-22
Allows for a job parameter based on a list of priorized parts

sabieber:
This allows to select from a list of classes that implement and register a common superclass when starting a job.
The list of provided select options is sorted by the priority of the parts.
As a label for the single select toString of the classes is called.
The job is provided with the registered implementation object of the selected part.

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/PriorizedPartListParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/PriorizedPartListParameter.java
@@ -1,0 +1,63 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.jobs.params;
+
+import sirius.kernel.di.std.Part;
+import sirius.kernel.di.std.Priorized;
+import sirius.kernel.nls.NLS;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/jobs/params/PriorizedPartListParameter.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Enumerates all parts implementing the common superclass part.
+     *
+     * @return the list of parts implementing the common superclass part sorted by priority
+     */
+    @Override
+    public Collection<E> getValues() {
+        if (parts == null) {
+            ArrayList<E> sortedParts = new ArrayList<>(globalContext.getParts(type));
```
andyHa:
```suggestion
            List<E> sortedParts = new ArrayList<>(globalContext.getParts(type));
```

src/main/java/sirius/biz/jobs/params/PriorizedPartListParameter.java
```diff
@@ -0,0 +1,63 @@
...
+            sortedParts.sort((o1, o2) -> {
```
andyHa: Comparators.comparing(Prioritized::getPriority) or the like

andyHa: result.sort(Comparator.comparingInt(Priorized::getPriority));

PR #447, opened 2019-07-23
Permits to specify a dedicated custom message type for facet filtering.

andyHa:
This seems to be a better approach than mis-using the message handler.


src/main/java/sirius/biz/process/logs/ProcessLog.java
```diff
@@ -110,6 +110,13 @@
     @NullAllowed
     private String messageHandler;
 
+    /**
+     * Contains a custome message type which will be shown as filter using {@link NLS#smartGet(String)}.Ø
```
sabieber:
```suggestion
     * Contains a custom message type which will be shown as filter using {@link NLS#smartGet(String)}.
```

PR #456, opened 2019-07-30
Marks users of the system tenants with a permission, even if they are in spy mode

meg1502:


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -82,6 +82,15 @@
      */
     public static final String PERMISSION_SYSTEM_TENANT_MEMBER = "flag-system-tenant-member";
 
+    /**
+     * This flag permission is granted to <b>all users</b> which originally belong to the system tenant.
+     * <p>
+     * The id of the system tenant can be set in the scope config. The system tenant usually is the administrative
+     * company which owns / runs the system.
+     * Unlike {@link #PERMISSION_SYSTEM_TENANT_MEMBER}, this flag is kept when when the user either has taken control over another tenant or uses account.
```
sabieber:
```suggestion
     * Unlike {@link #PERMISSION_SYSTEM_TENANT_MEMBER}, this flag is kept when the user either has taken control over another tenant or uses account.
```

welzfabian:
```suggestion
     * Unlike {@link #PERMISSION_SYSTEM_TENANT_MEMBER}, this flag is kept when when the user either has taken control over another tenant or user account.
```

sabieber: please clarify the last part after the `or`

welzfabian: should probably be "user account" instead of "uses account" ?

PR #457, opened 2019-07-31
 Correctly keeps 'flag-system-tenant' when switching to another tenant

meg1502:
Previously, this was only done for 'flag-system-administrator', but there might be other user profiles that are attached to 'flag-system-tenant'


src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -345,6 +345,10 @@ private String determineOriginalTenantId(WebContext ctx) {
 
     /**
      * Makes the current user belong to the given tenant.
+     * The user is will keep the permissions tied to its user,
```
andyHa:
```suggestion
    * <p>
     * The user is will keep the permissions tied to its user,
```

PR #474, opened 2019-08-15
Aha/l3 uplink fs

andyHa:


src/main/java/sirius/biz/storage/layer3/uplink/ConfigBasedUplink.java
```diff
@@ -0,0 +1,142 @@
...
+        this.config = config;
+    }
+
+    /**
+     * Creates the root directory of this uplink.
+     *
+     * @param parent the parent to use
+     * @return the root directory of this uplink
+     */
+    public VirtualFile getFile(VirtualFile parent) {
```
sabieber: Is this javadoc correct?

src/main/java/sirius/biz/storage/layer3/uplink/LocalDirectoryRoot.java
```diff
@@ -0,0 +1,294 @@
...
+            throw Exceptions.handle()
+                            .to(StorageUtils.LOG)
+                            .error(e)
+                            .withSystemErrorMessage("Layer 3/FS: Cannot move %s to %s: %s (%s)", file, name)
+                            .handle();
+        }
+    }
+
+    private boolean canMoveHandler(VirtualFile file) {
+        return !readonly && file.exists();
```
sabieber: Is this readonly flag in regards to the filesystem as a whole or the specific file? Single files could be readonly in the given filesystem too, or not?

andyHa: the idea is to mount the complete filesystem as readonly - all canXXX handlers can only try to predict if an operation is possible - the underlying FS always has the last word - sadly there is no api to ask...

PR #484, opened 2019-08-20
Actually checks the precision and scale of a field marked as numeric

sabieber:
Fixes: OX-5300


src/main/java/sirius/biz/importer/format/AmountScaleCheck.java
```diff
@@ -0,0 +1,76 @@
...
+        BigDecimal number = value.getBigDecimal();
+
+        if (number == null) {
+            throw new IllegalArgumentException(NLS.fmtr("AmountScaleCheck.errorMsg.notNumeric")
+                                                  .set(PARAM_VALUE, value.toString())
+                                                  .format());
+        }
+
+        BigDecimal rounded = new BigDecimal(number.unscaledValue(), number.scale()).setScale(scale, RoundingMode.FLOOR);
+        if (number.subtract(rounded).compareTo(BigDecimal.valueOf(0.00001)) > 0) {
```
andyHa: maybe use Doubles.EPSILON

src/main/resources/biz_de.properties
```diff
@@ -1,516 +1,515 @@
-AddressData.badZip = Fehler in '${name}': Die Postleitzahl '${zip}' ist für das gewählte Land ungültig.
```
andyHa: rather not reformat  this file as lokalise will revert all this again :-/

src/main/java/sirius/biz/importer/format/AmountScaleCheck.java
```diff
@@ -0,0 +1,76 @@
...
+
+import sirius.kernel.commons.Value;
+import sirius.kernel.nls.NLS;
+
+import javax.annotation.Nullable;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+/**
+ * Enforces a given numeric scale and precision.
```
andyHa: maybe add a single line of description what is actually checked?

src/test/java/sirius/biz/importer/format/ValueCheckSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+        when:
+        AmountScaleCheck scaleCheck = new AmountScaleCheck(5, 2)
+        Value number = Value.of("TEST")
+        and:
+        scaleCheck.perform(number)
+        then:
+        thrown IllegalArgumentException
+    }
+
+    def "numbers exceeding the provided precision are correctly marked as invalid"() {
```
andyHa: maybe also add a check which ensures that values outside of the range are detected like 123456

PR #498, opened 2019-09-04
Validates date FieldDefinitions against a specified date format

sabieber:
Fixes: OX-5375


src/main/java/sirius/biz/importer/format/FieldDefinition.java
```diff
@@ -147,11 +147,12 @@ public static String typeDate() {
...
      * Boilerplate to create a new date field.
      *
-     * @param name the name of the field
+     * @param name   the name of the field
+     * @param format the date format of the field
      * @return the newly created field
      */
-    public static FieldDefinition dateField(String name) {
-        return new FieldDefinition(name, typeDate());
+    public static FieldDefinition dateField(String name, String format) {
```
tbiScireum: how about a dateField method with a default format?

sabieber: this asks the question what a reasonable default date format should be

PR #504, opened 2019-09-06
Aha/vfs params

andyHa:


src/main/resources/default/templates/biz/storage/directoryfield.html.pasta
```diff
@@ -6,5 +6,10 @@
              name="@param.getName()"
              label="@param.getLabel()"
              help="@param.getDescription()"
+<<<<<<< Updated upstream
```
tbiScireum: this looks like it should not be here

src/main/resources/default/templates/biz/storage/filefield.html.pasta
```diff
@@ -6,5 +6,10 @@
              name="@param.getName()"
              label="@param.getLabel()"
              help="@param.getDescription()"
+<<<<<<< Updated upstream
```
tbiScireum: this looks like it should not be here

PR #505, opened 2019-09-06
Permits to supply a process file via an OutputStream.

andyHa:


src/main/java/sirius/biz/process/ProcessContext.java
```diff
@@ -218,4 +220,16 @@
      * @param data     the data to persist
      */
     void addFile(String filename, File data);
+
+    /**
+     * Adds a file to the process which will containd the data written into the {@link OutputStream}.
```
sabieber:
```suggestion
     * Adds a file to the process which will contain the data written into the {@link OutputStream}.
```

PR #524, opened 2019-09-18
Fixes the amount range check for FieldDefinitions

sabieber:
Fixes: OX-5413

### DIFF
--- a/src/test/java/sirius/biz/importer/format/AmountRangeCheckSpec.groovy
+++ b/src/test/java/sirius/biz/importer/format/AmountRangeCheckSpec.groovy
@@ -1,0 +1,108 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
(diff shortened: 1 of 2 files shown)


src/test/java/sirius/biz/importer/format/AmountRangeCheckSpec.groovy
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    def "values below the exclusive lower limit are correctly marked as invalid"() {
+        when:
+        new AmountRangeCheck(NumberFormat.NO_DECIMAL_PLACES).withLowerLimitExclusive(Amount.ONE).perform(Value.of(0))
+        then:
+        thrown IllegalArgumentException
+    }
+
+    def "values equal to the exclusive lower limit are correctly marked as valid"() {
```
welzfabian:
```suggestion
    def "values equal to the exclusive lower limit are correctly marked as invalid"() {
```

PR #544, opened 2019-10-08
Drops legacy code which was used to migrate Crunchlogs to Events.

andyHa:


src/main/java/sirius/biz/analytics/events/Event.java
```diff
@@ -67,25 +68,32 @@ protected void fill() {
...
         this.eventTimestamp = eventTimestamp;
     }
 
-    public void setEventDate(LocalDate eventDate) {
-        this.eventDate = eventDate;
-    }
-
-    public void setNode(String node) {
+    /**
+     * Ses a custom node on which this event was recorded.
```
idlira:
```suggestion
     * Sets a custom node on which this event was recorded.
```

PR #547, opened 2019-10-08
Move "isActive"-method to Processes-class

welzfabian:
To be more flexible where to request this info


andyHa:
consider using TaskContext.isActive if no ProcessContext is accessible / present, as this will work in all environments.

PR #550, opened 2019-10-09
Provides the docs and API of Processes.

andyHa:


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -496,6 +496,10 @@ protected boolean updateTitle(String processId, String newTitle) {
 
     /**
      * Adds the given link to the given process.
+     * <p>
+     * When running inside a process the preferred way to add a link is using
```
sabieber: Maybe also add one line to the class javadoc that using the context is preferred to directly using the processes part when accessing/altering the process?

PR #554, opened 2019-10-10
Implement "signalPut" in InterconnectCacheCoherence

welzfabian:
To synchronize "put"-calls in multi node systems for CoherentCaches

Fixes: OX-5352

Depends on https://github.com/scireum/sirius-kernel/pull/226

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
     <url>http://www.sirius-lib.net</url>
 
     <properties>
-        <sirius.kernel>DEVELOPMENT-SNAPSHOT</sirius.kernel>
+        <sirius.kernel>13.16</sirius.kernel>
         <sirius.web>24.0-rc24</sirius.web>
         <sirius.db>8.8</sirius.db>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
     <url>http://www.sirius-lib.net</url>
 
     <properties>
-        <sirius.kernel>13.12</sirius.kernel>
+        <sirius.kernel>DEVELOPMENT-SNAPSHOT</sirius.kernel>
         <sirius.web>24.0-rc24</sirius.web>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -17,7 +17,7 @@
     <url>http://www.sirius-lib.net</url>
 
     <properties>
-        <sirius.kernel>13.12</sirius.kernel>
+        <sirius.kernel>DEVELOPMENT-SNAPSHOT</sirius.kernel>
```
welzfabian: waiting on https://github.com/scireum/sirius-kernel/pull/226

src/main/java/sirius/biz/cluster/InterconnectCacheCoherence.java
```diff
@@ -59,6 +71,12 @@ public void handleEvent(JSONObject event) {
         } else if (Strings.areEqual(type, TYPE_REMOVE)) {
             String key = event.getString(MESSAGE_KEY);
             CacheManager.removeCoherentCacheKeyLocally(cache, key);
+        } else if (Strings.areEqual(type, TYPE_PUT)) {
+            String node = event.getString(MESSAGE_NODE);
+            if (!Strings.areEqual(node, CallContext.getNodeName())) {
```
andyHa: why is this check required here and not for remove etc?

src/main/java/sirius/biz/cluster/InterconnectCacheCoherence.java
```diff
@@ -59,6 +71,12 @@ public void handleEvent(JSONObject event) {
...
+                String key = event.getString(MESSAGE_KEY);
+                CacheManager.removeCoherentCacheKeyLocally(cache, key);
```
andyHa: maybe add a line comment to indicate that this is the intentional behaviour as deleting is cheaper and more efficient thant emulating a put via an extra lookup...

PR #556, opened 2019-10-13
Some cleanup - changes the way BlobStorageSpaces are created and improves some docs / updates somde versions

andyHa:


src/main/java/sirius/biz/storage/layer2/variants/BlobVariant.java
```diff
@@ -64,16 +64,16 @@
     Optional<FileHandle> download();
 
     /**
-     * Delivers the data of this blob into the given request.
+     * Determines if currently a conversion is in progress.
```
idlira: Determines if a conversion is currently in progress.

andyHa:
```suggestion
     * Determines if a conversion is currently in progress.
```

PR #560, opened 2019-10-15
Base implementation of storage converters

andyHa:


src/main/java/sirius/biz/storage/layer2/variants/ConversionEngine.java
```diff
@@ -0,0 +1,189 @@
...
+ * <p>
+ * Each blob can have any number of variants (e.g. resized JPG images of a given raw or EPS file). Each variant has
+ * a distinctive name. This name is used to lookup the appropriate conversion pipeline by inspecting the system
+ * config section {@link #CONFIG_KEY_VARIANTS}.
+ * <p>
+ * These settings will be used and passed on to the selected {@link Converter}. Note that the converter isn't
+ * specified directly, but another config section in {@link #CONFIG_KEY_CONVERTERS} is used to determine the effective
+ * implementation (addressing the appropriate {@link ConverterFactory} along with default settings).
+ * <p>
+ * Using this approach we can provide some standard converters and standard variants which cen be fully customized
```
idlira:
```suggestion
 * Using this approach we can provide some standard converters and standard variants which can be fully customized
```

src/main/resources/component-biz.conf
```diff
@@ -590,8 +590,25 @@ storage {
             enabled = true
 
             # Contains a list of hostnames or IP addresses to which a conversion can be delegated by
-            # forwarding a request.
+            # forwaring a request.
```
idlira:
```suggestion
            # forwarding a request.
```

PR #571, opened 2019-10-22
Cleanup of the SQL impl

andyHa:


src/main/java/sirius/biz/storage/layer2/Blob.java
```diff
@@ -207,16 +203,18 @@
...
      */
-    List<BlobVariant> getVariants();
+    List<? extends BlobVariant> fetchVariants();
 
     /**
-     * Lists all known revisions of this blob.
+     * Tries to find the variant with the given name.
      * <p>
-     * A revision is a backup copy of previously provided data of this blob.
+     * This will check if a variant with the given name exists and return it. If no such varian exists,
```
idlira: ... If no such variant **exists**, ...

_Sorry, suggestions cannot be applied_

src/main/java/sirius/biz/storage/layer2/BlobStorageSpace.java
```diff
@@ -166,4 +166,11 @@
      * @param variant  the variant to deliver. Use {@link URLBuilder#VARIANT_RAW} to deliver the blob itself
      */
     void deliver(@Nonnull String blobKey, @Nonnull String variant, @Nonnull Response response);
+
+    /**
+     * Performes some housekeeping and maintenance tasks.
```
idlira: Performs

PR #572, opened 2019-10-22
Aha/storage layer2 mongo

andyHa:


src/main/java/sirius/biz/storage/layer2/mongo/MongoVariant.java
```diff
@@ -21,29 +21,60 @@
...
     private String physicalObjectKey;
 
+    /**
+     * Contains the file size of the converted file.
+     */
     public static final Mapping SIZE = Mapping.named("size");
     private long size = 0;
 
+    /**
+     * Contains the timestamp when the last conversion was attempted.
```
sabieber: Why is this not nullable but the node name on which the conversion was tried on is?

PR #573, opened 2019-10-22
Aha/jobs storage

andyHa:


src/main/java/sirius/biz/process/Process.java
```diff
@@ -156,7 +165,7 @@
      * Contains files generated by the process.
      */
     public static final Mapping FILES = Mapping.named("files");
-    private final NestedList<ProcessFile> files = new NestedList<>(ProcessFile.class);
+    private final BlobContainer files = new BlobContainer(this, SPACE_NAME_PROCESSES);
```
sabieber: Wont this result in problems for products where files were already attached to jobs?

PR #581, opened 2019-10-24
Make the parsing of dates strict in DateTimeFormatCheck

welzfabian:
This detects wrong dates like the 31.09.2019 (September only has 30 days).

The trade off to this solution is, that years have to be declared with 'u' in pattern instead of 'y'.
The other option one could achieve strict parsing with is the 'SimpleDateFormat'-class, but it comes with its own set of problems, also it is legacy java.

Additional reading from StackOverflow:

DateTimeFormatter:
`uuuu` versus `yyyy`: https://stackoverflow.com/questions/41177442/uuuu-versus-yyyy-in-datetimeformatter-formatting-pattern-codes-in-java
Issues with `yyyy` isntead of `uuuu`: https://stackoverflow.com/questions/41103603/issue-with-datetimeparseexception-when-using-strict-resolver-style

SimpleDateFormat:
Ignores additional/missing characters in pattern: https://stackoverflow.com/questions/16014488/simpledateformat-parse-ignores-the-number-of-characters-in-pattern
Make SimpleDateFormat more strict (also see comments of top answer): https://stackoverflow.com/questions/13088140/java-how-to-parse-a-date-strictly

### DIFF
--- a/src/main/java/sirius/biz/importer/format/DateTimeFormatCheck.java
+++ b/src/main/java/sirius/biz/importer/format/DateTimeFormatCheck.java
@@ -20,6 +20,10 @@ import java.time.format.ResolverStyle;
 
 /**
  * Determines if the given values matches a date format.
+ * <p>
+ * The formating notation used is the notation described in {@link DateTimeFormatter}.
+ * <p>
+ * <i>Hint</i>: Use 'u' instead of 'y' for years.
  */
 public class DateTimeFormatCheck implements ValueCheck {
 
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/biz/importer/format/DateTimeFormatCheck.java
```diff
@@ -26,12 +32,22 @@
...
+     * The formating noation used is the notation decribed in {@link DateTimeFormatter}.
+     * <p>
+     * <i>Hint</i>: Use 'u' instead of 'y' for years.
      *
-     * @param format the {@link org.joda.time.format.DateTimeFormat} which will be used to check the value
+     * @param format the String describing the formats of the dates
      */
     public DateTimeFormatCheck(@Nonnull String format) {
+        if (format.contains("y")) {
+            // We want to use the 'strict' resolver, so dates like 30.02.2019 are marked invalid and not resolved to a
```
andyHa: put all that in the javadoc above?

src/main/java/sirius/biz/importer/format/DateTimeFormatCheck.java
```diff
@@ -26,12 +32,24 @@
 
     /**
      * Creates a new check using the given date format.
+     * <p>
+     * The formating noation used is the notation decribed in {@link DateTimeFormatter}.
```
sabieber:
```suggestion
     * The formatting notation used is the notation described in {@link DateTimeFormatter}.
```

src/main/java/sirius/biz/importer/format/DateTimeFormatCheck.java
```diff
@@ -13,11 +13,17 @@
...
 import javax.annotation.Nullable;
+import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
 
 /**
  * Determines if the given values matches a date format.
+ * <p>
+ * The formating notation used is the notation described in {@link DateTimeFormatter}.
```
sabieber:
```suggestion
 * The formatting notation used is the notation described in {@link DateTimeFormatter}.
```

PR #589, opened 2019-10-29
Uses the proper method to flush a BatchContext / ImporterContext.

andyHa:
Otherwise the newly added use-after-close check will trigger.


andyHa:
BUILD

PR #591, opened 2019-10-29
Direct access via InputStream

andyHa:


src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
```diff
@@ -103,6 +107,30 @@ public void upload(String objectId, InputStream inputStream, long contentLength)
...
+        try {
+            if (Strings.isEmpty(objectId)) {
+                return Optional.empty();
+            }
+            return Optional.ofNullable(engine.getAsStream(name, objectId));
+        } catch (IOException e) {
+            throw Exceptions.handle()
+                            .error(e)
+                            .to(StorageUtils.LOG)
+                            .withSystemErrorMessage(
```
idlira: Parameters missing: 2 arguments x 4 %s placeholders

andyHa: The two last values are automagically added by withSystemErrorMessage where the first is the error message and the second is the name of the exception class

PR #593, opened 2019-11-05
Fixes the logic to allow the creation of new folders via ftp

tbiScireum:
The current logic will create a placeholder object and returns it with findChild.
Because of that we cannot just do a null check. We also need to check the exists method of the child
- Fixes: SE-8975


src/main/java/sirius/biz/storage/layer3/downlink/ftp/BridgeFile.java
```diff
@@ -34,6 +34,7 @@
 
     BridgeFile(VirtualFile file) {
         this.file = file;
+        this.parent = file.parent();
```
andyHa: why is that needed?

PR #594, opened 2019-11-05
SIRI-91 Adds indices to codelist entries for faster lookups

jmuscireum:


src/main/java/sirius/biz/codelists/mongo/MongoCodeListEntry.java
```diff
@@ -20,6 +22,10 @@
  */
 @Framework(MongoCodeLists.FRAMEWORK_CODE_LISTS_MONGO)
 @TranslationSource(CodeListEntry.class)
+@Index(name = "lookup",
+        columns = {"codeList", "codeListEntryData_code"},
+        columnSettings = {Mango.INDEX_ASCENDING, Mango.INDEX_ASCENDING},
+        unique = true)
```
andyHa: does this imply anything for mongo? (unique)

andyHa:
maybe also add one for (SQL|Mongo)CodeList

PR #600, opened 2019-11-11
Enhances errors for retrieving and verifying entities in BizControllers

meg1502:


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -127,29 +127,36 @@ protected void assertTenant(@Nonnull String tenantId) {
...
+     * @param parentEntity the entity the reference must point to
+     * @param <E>          the generic type the the parent being referenced
+     * @param <I>          the type of the id column of E
      * @throws sirius.kernel.health.HandledException if the entities do no match
      */
-    protected <I, E extends BaseEntity<I>> void setOrVerify(BaseEntity<?> owner, BaseEntityRef<I, E> ref, E entity) {
-        if (!Objects.equals(ref.getId(), entity.getId())) {
-            if (owner.isNew()) {
-                ref.setValue(entity);
+    protected <I, E extends BaseEntity<I>> void setOrVerifyParent(BaseEntity<?> child,
```
andyHa: this isn't necessarrly a parent child relation so i'd rather keep the neutral name

src/main/resources/biz_de.properties
```diff
@@ -36,12 +36,12 @@ BizController.deleteProcess = Löschvorgang
 BizController.deleteProcessTitle = Löschen von '${entity}'
 BizController.deletingInBackground = Das Datenobjekt enthält einige verknüpfte Datenobjekte und wird deshalb im Hintergrund gelöscht.
 BizController.illegalArgument = Fehlerhafte Eingabe. Bitte korrigieren Sie die markierten Felder
-BizController.invalidReference = Eine Beziehung zwischen den gegebenen Datenobjekten ist inkonsistent.
+BizController.invalidReference = Im Objekt '${child}' wurde das Übergeordnete Objekt '${parent}' erwartet, aber es wurde '${actual}' gefunden.
```
sabieber:
```suggestion
BizController.invalidReference = Im Objekt '${child}' wurde das referenzierte Objekt '${parent}' erwartet, aber es wurde '${actual}' gefunden.
```

src/main/resources/biz_en.properties
```diff
@@ -26,12 +26,12 @@ BizController.deleteProcess = Delete process
 BizController.deleteProcessTitle = Deletion of '${entity}'
 BizController.deletingInBackground = The entity contains some linked entities and is therefore deleted in the background.
 BizController.illegalArgument = Incorrect input. Please correct the marked fields.
-BizController.invalidReference = A relationship between the given data objects is inconsistent.
+BizController.invalidReference = A relationship in '${child}' should point to parent object '{$parent}', but points to '${actual}' instead.
```
sabieber:
```suggestion
BizController.invalidReference = A relationship in '${child}' should point to refrenced object '{$parent}', but points to '${actual}' instead.
```

src/main/resources/biz_de.properties
sabieber: Can we also include the type in this error message?

meg1502: $parent , $child are the unique-name of the respective entity, which is something like 'MongoTenant-123' , i did this as i found the error message with 'id' of 'type' too long.

src/main/java/sirius/biz/web/BizController.java
meg1502: The first javadoc line describing the method was "Enusures or establishes a parent child relation." how would you replace the javadoc?

andyHa: Maybe, "Properly maintains a reference to an entity with write-once semantic"
and also add  a @see BaseEntityRef#writeOnceOn and encourage using it

src/main/resources/biz_de.properties
sabieber: sounds good

PR #606, opened 2019-11-14
VFS Jobs Root

andyHa:
Provides a virtual VFS directory for each job containing another virtual directory for each preset of said job.

If a file is uploaded there it is moved into the hidden "tmp" folder and an execution of the job is scheduled.


src/main/java/sirius/biz/storage/layer2/Blob.java
```diff
@@ -183,6 +183,21 @@
      */
     OutputStream createOutputStream(@Nullable String filename);
 
+    /**
+     * Creates a output stream which can be used to update the contents of this blob.
+     * <p>
+     * Stores the once the stream is closed just like {@link #createOutputStream(String)} but permits to add a callback
```
idlira:
```suggestion
     * Stores the contents once the stream is closed just like {@link #createOutputStream(String)} but permits to add a callback
```

src/main/java/sirius/biz/storage/layer2/BlobStorageSpace.java
```diff
@@ -107,12 +107,33 @@
      * Such objects will be automatically deleted if the referencing entity is
      * deleted. It is made permanent as soon as the referencing entity is saved, otherwise it will be deleted
      * automatically.
+     * <p>
+     * Note that there is almost no use-case to directory call this method (outside of {@link BlobHardRef}.
```
idlira:
```suggestion
     * Note that there is almost no use-case to directly call this method (outside of {@link BlobHardRef}.
```

src/main/java/sirius/biz/storage/layer3/TmpRoot.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Provides a stealth directory to store temporary files.
+ * <p>
+ * Neither this directory itself nor its files are listed / enumerated in the VFS. However the directory as well
+ * as each blob in the <b>tmp</b> {@link sirius.biz.storage.layer2.BlobStorageSpace} can be resolved and therefore
+ * be accessed.
+ * <p>
+ * This will resolve paths like <tt>/tmp/BLOB-KEY/FILENAME</tt>. Note that the filename as well as the tenant which
+ * tries to access the file must match. Most probably these blobs are created using
+ * {@link sirius.biz.storage.layer2.BlobStorageSpace#createTemporaryBlob(String)} and the persisted once completely
```
idlira:
```suggestion
 * {@link sirius.biz.storage.layer2.BlobStorageSpace#createTemporaryBlob(String)} and then persisted once completely
```

PR #607, opened 2019-11-14
Repairs an update loop which is based on optimistic locking.

andyHa:
The retry-loop which was previously implemented is pretty useless for two reasons.
1) The entity wasn't marked as Versioned - therefore no optimistic locking was enabled
2) The loop didn't refresh the entity before re-trying.

Fixes: OX-5352 (hopefully)

### DIFF
--- a/src/main/java/sirius/biz/process/Processes.java
+++ b/src/main/java/sirius/biz/process/Processes.java
@@ -401,6 +401,7 @@ public class Processes {
                 return true;
             } catch (OptimisticLockException e) {
                 Wait.randomMillis(250, 500);
+                process = elastic.find(Process.class, processId).orElse(null);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -401,6 +401,7 @@ private boolean modify(String processId, Predicate<Process> checker, Consumer<Pr
                 return true;
             } catch (OptimisticLockException e) {
                 Wait.randomMillis(250, 500);
+                process = elastic.find(Process.class, processId).orElse(null);
```
meg1502: move L387 nullcheck into retry loop, as process can now become null while looping ?

sabieber: shouldnt we move the if process null return false into the while loop then to prevent NPEs for deleted processes (shouldnt be really neccessary)

PR #612, opened 2019-11-19
Aha/process counters

andyHa:


src/main/java/sirius/biz/process/ProcessContext.java
```diff
@@ -73,6 +73,13 @@
      */
     void addDebugTiming(String counter, long millis);
 
+    /**
+     * Increents the given performance counter by one.
```
meg1502: Increments
```suggestion
     * Increments the given performance counter by one.
```

src/main/java/sirius/biz/process/ProcessContext.java
```diff
@@ -73,6 +73,13 @@
...
+     *
+     * @param counter the counter to increment
+     */
+    void addCounter(String counter);
```
meg1502: i think it may be worthwhile to rename this to `incrementCounter` as a method addCounter especially with no int parameter rather implies that this creates/attaches something new to the process instead of incrementing a counter by 1.

andyHa: right, this would also look like a lazy guy just lovelessly copy+pasted the method above :)

PR #619, opened 2019-11-25
Base for VFS Uplinks (FTP, SFTP)

andyHa:


src/main/java/sirius/biz/storage/layer3/uplink/util/RelativePath.java
```diff
@@ -0,0 +1,57 @@
...
+
+package sirius.biz.storage.layer3.uplink.util;
+
+import sirius.biz.storage.layer3.MutableVirtualFile;
+import sirius.kernel.commons.Strings;
+
+/**
+ * Wraps a path to be attached to a {@link MutableVirtualFile} when providing an uplink.
+ */
+public class RelativePath {
```
jmuscireum: What is the benefit over using https://docs.oracle.com/javase/8/docs/api/java/io/File.html#File-java.lang.String-java.lang.String-

src/main/java/sirius/biz/storage/layer3/downlink/ssh/sftp/BridgeSeekableByteChannel.java
```diff
@@ -21,9 +21,10 @@
 /**
  * Provides a simple implementation of the seekable channel.
  * <p>
- * This implementation doesn't actually support seeking, as this is not supported by the
- * {@link sirius.biz.storage.layer3.VirtualFileSystem}. Nevertheless, as long as all reads are sequential and don't
- * skip any data, we fullfill all calls.
+ * This implementation doesn't actually support much seeking, as this is not supported by the
+ * {@link sirius.biz.storage.layer3.VirtualFileSystem}. Nevertheless, as long as all writes are sequential and don't
+ * skip any data, we fullfill all calls. Also we can skip some bytes when reading as long as it is in the forward
```
idlira:
```suggestion
 * skip any data, we fulfill all calls. Also we can skip some bytes when reading as long as it is in the forward
```

src/main/java/sirius/biz/storage/layer3/uplink/util/RelativePath.java
idlira: +1

src/main/java/sirius/biz/storage/layer3/uplink/util/UplinkConnectorPool.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    private GenericObjectPool<UplinkConnector<?>> createPool(UplinkConnectorConfig<?> uplinkConnectorConfig) {
+        UplinkConnectorFactory uplinkConnectorFactory = new UplinkConnectorFactory(uplinkConnectorConfig);
+        GenericObjectPool<UplinkConnector<?>> pool = new GenericObjectPool<>(uplinkConnectorFactory);
+        uplinkConnectorFactory.linkToPool(pool);
+        pool.setMaxIdle(uplinkConnectorConfig.maxIdle);
+        pool.setMaxTotal(uplinkConnectorConfig.maxActive);
+        pool.setTimeBetweenEvictionRunsMillis(30_000);
```
sabieber: Extract into constants?

src/main/java/sirius/biz/storage/layer3/uplink/util/RelativePath.java
andyHa: that this simply wraps a string describing a path on a remote server which doesn't share any api with a File in the local file system

idlira: As the JavaDoc states: An abstract representation of file and directory pathnames.
A java.io.File object does not necessarily means a real file in a local file system.

sabieber: Or a Path and its [resolve](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.nio.file.Path)) method could be used

PR #620, opened 2019-11-25
Only enables the legacy versions framework if an SQLBlobStorage is ac…

andyHa:
…tive.

Only then we know that a SQL database is around.

### DIFF
--- a/src/main/java/sirius/biz/storage/versions/VersionedFile.java
+++ b/src/main/java/sirius/biz/storage/versions/VersionedFile.java
@@ -1,7 +1,7 @@
 package sirius.biz.storage.versions;
 
-import sirius.biz.storage.Storage;
 import sirius.biz.storage.StoredObjectRef;
+import sirius.biz.storage.layer2.jdbc.SQLBlobStorage;
 import sirius.biz.tenants.jdbc.SQLTenantAware;
 import sirius.db.mixing.Mapping;
 import sirius.db.mixing.annotations.Length;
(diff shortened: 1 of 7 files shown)


src/main/java/sirius/biz/storage/versions/VersionedFile.java
```diff
@@ -15,7 +15,7 @@
 /**
  * Entity holding meta information about a versioned file.
  */
-@Framework(Storage.FRAMEWORK_STORAGE)
+@Framework(SQLBlobStorage.FRAMEWORK_JDBC_BLOB_STORAGE)
```
sabieber: Is there no Mongo implementation for this or in other words is no mongo version planned as this has a generic name and lives in a generic package?
This does not match the other biz/db stuff.

src/main/java/sirius/biz/storage/versions/VersionedFiles.java
```diff
@@ -26,7 +27,7 @@
 /**
  * Small helper class to access and create {@link VersionedFile versioned files}.
  */
-@Register(classes = VersionedFiles.class, framework = Storage.FRAMEWORK_STORAGE)
+@Register(classes = VersionedFiles.class, framework = SQLBlobStorage.FRAMEWORK_JDBC_BLOB_STORAGE)
```
sabieber: same as above

src/main/java/sirius/biz/storage/versions/VersionedFile.java
andyHa: This is a legacy framework which was only built for SQL - this will be refactored, in the meantime we bind this to the sql storage.

src/main/java/sirius/biz/storage/versions/VersionedFiles.java
andyHa: yep

PR #634, opened 2019-12-02
Aha/vfs simplify

andyHa:


src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -170,6 +174,42 @@ public String path() {
...
+     * @param rootParent one of the parent directories of <tt>this</tt>
+     * @return the relative path from the given root parent to <tt>this</tt>
+     * @throws IllegalArgumentException if the given root parent isn't part of the path of <tt>this</tt>
+     */
+    public String relativePath(VirtualFile rootParent) {
+        return relativePathList(rootParent).stream().map(VirtualFile::name).collect(Collectors.joining("/"));
+    }
+
+    /**
+     * Returns the path a list of virtual files from the given root parent to this.
```
idlira:
```suggestion
     * Returns the path as a list of virtual files from the given root parent to this.
```

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -517,21 +557,88 @@ public void children(FileSearch search) {
         }
     }
 
+    /**
+     * Lists all childrens of this file.
```
idlira:
```suggestion
     * Lists all children of this file.
```

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -517,21 +557,88 @@ public void children(FileSearch search) {
...
+     * <p>
+     * Note that this will load all children into a list. In case of very large directories this might be slow
+     * and memory consuming. Consider using {@link #children(FileSearch)} which supports filters an pagination.
```
sabieber:
```suggestion
     * and memory consuming. Consider using {@link #children(FileSearch)} which supports filters and pagination.
```

src/main/java/sirius/biz/storage/layer2/L3Uplink.java
```diff
@@ -118,18 +118,16 @@ Blob createAsBlob() {
          * @param name            the name of the placeholder
          * @return a placeholder representing a non existent file or directory
          */
-        protected Optional<VirtualFile> createPlaceholder(@Nullable Directory parentDirectory,
-                                                          VirtualFile parent,
-                                                          String name) {
+        protected VirtualFile createPlaceholder(@Nullable Directory parentDirectory, VirtualFile parent, String name) {
```
sabieber: Could we annotate the public and protected functions with Nullable and Nonnull when we remove the usage of Optional? This would be helpful for callers of the API and ovveriders of the protected methods by an extending API.

Also its generally good codestyle and indicates that thought was put into wether or not the method can return null or not.

PR #636, opened 2019-12-03
Adds a transient field in login to be able to skip the password creation

meg1502:
Fixes: OX-5555


src/main/java/sirius/biz/model/LoginData.java
```diff
@@ -170,28 +170,42 @@
...
+    /**
+     * Contains a cleartext password that has just been created by a user or generated by the system.
+     * <p>
+     * This only exists in memory and will be hashed into {@link #PASSWORD_HASH} with a {@link #SALT} when this entity is saved.
+     */
     @Transient
     private String cleartextPassword;
 
+    /**
+     * Skips the creation of {@link #generatedPassword} on save which wis normally used to fill in {@link #passwordHash} with a generated {@link #salt}.
```
sabieber:
```suggestion
     * Skips the creation of {@link #generatedPassword} on save which was normally used to fill in {@link #passwordHash} with a generated {@link #salt}.
```

src/main/java/sirius/biz/model/LoginData.java
```diff
@@ -170,28 +170,42 @@
...
+     * <p>
+     * If set to true, this will allow for users without a valid password login to exists in the database.
```
sabieber:
```suggestion
     * If set to true, this will allow for users without a valid password login to exist in the database.
```

src/main/java/sirius/biz/model/LoginData.java
idlira:
```suggestion
     * Skips the creation of {@link #generatedPassword} on save which is normally used to fill in {@link #passwordHash} with a generated {@link #salt}.
```

src/main/java/sirius/biz/model/LoginData.java
idlira:
```suggestion
     * If set to true, this will allow for users without a valid password login to exist in the database.
```

src/main/java/sirius/biz/model/LoginData.java
meg1502:
```suggestion
     * Skips the creation of {@link #generatedPassword} on save which is normally used to fill in {@link #passwordHash} with a generated {@link #salt}.
```

src/main/java/sirius/biz/model/LoginData.java
```diff
@@ -287,15 +302,16 @@ public static String hashPassword(@Nonnull String salt, @Nonnull String password
      * @return a tuple where the first parameter determines if the password is valid
      */
     public PasswordVerificationResult checkPassword(String username, String password) {
-        for (PasswordHashFunction hashFunction : hashFunctions) {
-            String givenPasswordHash = hashFunction.computeHash(username, salt, password);
-            if (Strings.isFilled(givenPasswordHash) && Strings.areEqual(givenPasswordHash, passwordHash)) {
-                return hashFunction.isOutdated() ?
-                       PasswordVerificationResult.VALID_NEEDS_RE_HASH :
-                       PasswordVerificationResult.VALID;
+        if (Strings.isFilled(passwordHash)) {
```
andyHa: invert and abort early

src/main/java/sirius/biz/model/LoginData.java
```diff
@@ -402,6 +418,10 @@ public String getPasswordHash() {
         return passwordHash;
     }
 
+    public void skipPasswordCreation() {
```
andyHa: javadoc

PR #637, opened 2019-12-04
Checks if the field salutation is a valid entry codelist entry

idlira:
This validates the against the raw value (SIR, MADAM) and not a translated version. The goal is to avoid invalid entries to be stored in the database.

For export/import purposes, the raw value shall be used. Handling translated values requires a higher effort to be evaluated at a later stage.

This is the evaluation in action during an import:

image


src/main/java/sirius/biz/model/PersonData.java
```diff
@@ -153,6 +154,14 @@ public int hashCode() {
         return Objects.hash(title, salutation, firstname, lastname);
     }
 
+    @BeforeSave
+    protected void checkSalutation() {
+        if (Strings.isFilled(salutation)) {
+            // a HandledException will be thrown if the value is not contained in the codelist
+            codeLists.getRequiredValue("salutations", salutation);
```
andyHa: Wouldn’t it be nicer to create a verify value in codelists so that an exception is thown on purpose and not as side effect? The code should already be available anyway(?)

src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -191,6 +191,33 @@ public String getRequiredValue(@Nonnull String codeList, @Nonnull String code) {
...
 
+    /**
+     * Checks if the given code exists inside the given code list or throws an exception if no matching entry exists
+     *
+     * @param codeList the code list to search in
+     * @param code     the code to lookup
+     * @throws sirius.kernel.health.HandledException if no entry exists for the given code or code list
+     */
+    public void verifyValue(@Nonnull String codeList, @Nonnull String code) {
+        Optional<L> cl = findCodelist(codeList);
```
sabieber: Pls use a more descriptive variable name. Parameter Name could be `codeListName` so this could be called `codeList` maybe

src/main/resources/biz_de.properties
```diff
@@ -53,6 +53,7 @@ CodeListController.codeMustBeFilled = Das Feld "Code" darf nicht leer sein.
 CodeListEntry.additionalValue = Zusatzwert
 CodeListEntry.plural = Einträge
 CodeListEntry.value = Wert
+CodeLists.missingList = Der Codeliste '${codeList}' ist nicht vorhanden.
```
meg1502:
```suggestion
CodeLists.missingList = Die Codeliste '${codeList}' existiert nicht.
```

src/main/java/sirius/biz/model/PersonData.java
idlira: Done

PR #642, opened 2019-12-10
Makes sure that the parent directory also exists before trying to cre…

tbiScireum:
…ate the actual directory

- Fixes: SE-9101


src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -738,6 +738,9 @@ public boolean tryCreateAsDirectory() {
                 return false;
             }
 
+            if (parent() != null) {
+                parent().createAsDirectory();
```
andyHa:
```suggestion
                if (!parent().createAsDirectory()) {
                    return false;
                }
```

PR #652, opened 2020-01-02
Aha/timeouts

andyHa:


src/main/java/sirius/biz/storage/layer3/uplink/sftp/SFTPUplinkConnectorConfig.java
```diff
@@ -44,6 +45,8 @@ protected int getDefaultPort() {
     private SshClient getClient() {
         if (sshClient == null) {
             sshClient = SshClient.setUpDefaultClient();
+            sshClient.getProperties().putIfAbsent(FactoryManager.IDLE_TIMEOUT, idleTimeoutMillis);
+            sshClient.getProperties().putIfAbsent(FactoryManager.NIO2_READ_TIMEOUT, readTimeoutMillis);
```
meg1502: should this really be `putIfAbsent` ?
If setUpDefaultClient changes and provides a idle timeout by default, this will no longer apply but it should still apply?

PR #653, opened 2020-01-02
Uses createNowOrUpdateInBatch as default way of updating entities.

andyHa:
Otherwise "unique" checks might fail (at the database level) as the inserts are queued in the batch
and therefore not enforced by the framework. This adds too much trouble for outweight the performance improvements
of a batched insert.

If an entity is known to not have any unique constraints / such conflicts - which is probably very rare - this method
can be overwritten and changed back to a batched insert.

### DIFF
--- a/src/main/java/sirius/biz/importer/Importer.java
+++ b/src/main/java/sirius/biz/importer/Importer.java
@@ -238,7 +238,7 @@ public class Importer implements Closeable {
     }
 
     /**
-     * Creates the entity immediatelly or updates the given entity in the underlying database using a batch update.
+     * Creates the entity immediately or updates the given entity in the underlying database using a batch update.
      * <p>
      * Uses the appropriate {@link ImportHandler} to determine if the entity already exists and then either updates
      * the entity or creates a new one.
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/importer/Importer.java
```diff
@@ -219,15 +222,39 @@ public ImportDictionary getExportDictionary(Class<? extends BaseEntity<?>> type)
...
+     * @see #createNowOrUpdateInBatch(BaseEntity)
+     * @see ImportHandler#createOrUpdateInBatch(BaseEntity)
      */
     @SuppressWarnings("unchecked")
     public <E extends BaseEntity<?>> void createOrUpdateInBatch(E entity) {
         context.findHandler((Class<E>) entity.getClass()).createOrUpdateInBatch(entity);
     }
 
+    /**
+     * Creates the entity immediatelly or updates the given entity in the underlying database using a batch update.
```
meg1502:
```suggestion
     * Creates the entity immediately or updates the given entity in the underlying database using a batch update.
```

PR #654, opened 2020-01-02
Hides parameters marked as hidden but preserving its context

idlira:


src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -44,11 +44,14 @@
         <div class="row">
             <div class="col-md-8">
                 <div class="row">
-                    <i:for type="sirius.biz.jobs.params.Parameter" var="param" items="job.getParameters()">
+                    <i:for type="sirius.biz.jobs.params.Parameter" var="param" items="job.getVisibleParameters(context)">
                         <i:dynamicInvoke template="@param.getTemplateName()" param="@param" context="@context"/>
```
meg1502: Wouldn't it be easier to create a new template file 'bla/bla/hidden.html.pasta' with
```
<input type="hidden" name="@param.getName()" value="@context.getOrDefault(param.getName(), null)"/>
```
and add some logic to the Parameter class that wraps `@param.getTemplateName()` and will return the 'bla/bla/hidden.html.pasta' if the parameter is hidden instead of the template name implemented in the parameter subclasses?

As an alternative, this could also be done using <i:if> right here in the template. 
These solutions would have the benefit of only iterating the parameter list once instead of three times.

src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -44,11 +44,14 @@
...
                     </i:for>
                 </div>
             </div>
+            <i:for type="sirius.biz.jobs.params.Parameter" var="param" items="job.getInvisibleParameters(context)">
```
andyHa: add this block in the interactive-job html.pasta as well?

src/main/resources/default/templates/biz/jobs/job.html.pasta
idlira: New template hidden.html.pasta returned by computeTemplateName which is now used by the pasta templates.

src/main/resources/default/templates/biz/jobs/job.html.pasta
idlira: Covered with new redesign

src/main/java/sirius/biz/jobs/params/Parameter.java
```diff
@@ -127,6 +129,21 @@ public boolean isVisible(Map<String, String> context) {
...
 
+    /**
+     * Returns the name of the template used to render the parameter in the UI.
+     * <p>
+     * Similar to {@link #getTemplateName()}, but this method considers the visibility
+     * of the parameter and delivers an alternative template in case the parameter should be hidden.
+     *
+     * @return the name or path of the template used to render the parameter
+     */
+    public String computeTemplateName() {
```
andyHa: getEffectiveTemplateName / fetchEffectiveTemplateName?

src/main/java/sirius/biz/jobs/params/Parameter.java
```diff
@@ -127,6 +129,21 @@ public boolean isVisible(Map<String, String> context) {
...
+        if (this.visibility == Visibility.HIDDEN) {
```
andyHa: shouldn't this also take "ONLY_WITH_VALUE" into account?

andyHa: i.e. invoke isVisible

src/main/java/sirius/biz/jobs/params/Parameter.java
idlira: Refactored to getEffectiveTemplateName

src/main/java/sirius/biz/jobs/params/Parameter.java
idlira: Validating with isVisible()

PR #655, opened 2020-01-03
Aha/checks

andyHa:


src/main/java/sirius/biz/tenants/UserAccountData.java
```diff
@@ -110,28 +109,35 @@ public UserAccountData(BaseEntity<?> userObject) {
...
-                            .handle();
-        }
 
-        userObject.assertUnique(UserAccount.USER_ACCOUNT_DATA.inner(LOGIN).inner(LoginData.USERNAME),
-                                getLogin().getUsername());
+        // Ensure that the username is is filled and unique...
+        userObject.verifyIfChangedFailIfEmpty(UserAccount.USER_ACCOUNT_DATA.inner(LOGIN).inner(LoginData.USERNAME),
+                                              () -> {
+                                                  // Make it lowercase...
+                                                  getLogin().setUsername(getLogin().getUsername().toLowerCase());
```
idlira: Requires a check if already saved data is lowercase (and a batch-convert) when not?

src/main/java/sirius/biz/tenants/UserAccountData.java
```diff
@@ -110,28 +109,35 @@ public UserAccountData(BaseEntity<?> userObject) {
...
-        } else {
-            throw Exceptions.createHandled()
-                            .withNLSKey("Property.fieldNotNullable")
-                            .set("field", NLS.get("LoginData.username"))
-                            .handle();
-        }
 
-        userObject.assertUnique(UserAccount.USER_ACCOUNT_DATA.inner(LOGIN).inner(LoginData.USERNAME),
-                                getLogin().getUsername());
+        // Ensure that the username is is filled and unique...
```
idlira:
```suggestion
        // Ensure that the username is filled and unique...
```

src/main/java/sirius/biz/model/ContactData.java
```diff
@@ -24,21 +22,24 @@
...
     /**
-     * Validates a phone numner.
+     * Validates a phone number.
      */
-    public static final Pattern VALID_PHONE_NUMBER =
-            Pattern.compile("\\+?\\d+( \\d+)*( */( *\\d+)+)?( *\\-( *\\d+)+)?");
-
-    @Transient
-    private boolean validatePhoneNumbers;
+    public static final Pattern VALID_PHONE_NUMBER = Pattern.compile("\\+?\\d+( \\d+)*( */( *\\d+)+)?( *-( *\\d+)+)?");
```
idlira: Is this format valid internationally?

src/main/java/sirius/biz/tenants/UserAccountData.java
andyHa: this code isn't new, it was just moved to another place...

src/main/java/sirius/biz/model/ContactData.java
andyHa: This regex isn't new / has not changed semantically. The \\- was a redundant escape which has been removed (as suggested by IntelliJ) - hopefully the regex was/is valid :)

PR #662, opened 2020-01-16
New DateRange and CodeListEntry job parameters

idlira:
Introduces new job parameter types for DateRange and CodeListEntry objects.

Fixes: OX-5522

### DIFF
--- a/src/main/java/sirius/biz/codelists/CodeLists.java
+++ b/src/main/java/sirius/biz/codelists/CodeLists.java
@@ -395,6 +395,18 @@ public abstract class CodeLists<I, L extends BaseEntity<I> & CodeList, E extends
     }
 
     /**
+     * Returns the {@link CodeListEntry} from the given code list associated with the given code.
+     *
+     * @param codeList the code list to search in
+     * @param code     the code to lookup
+     * @return the entry associated with the code or an empty optional otherwise
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -394,6 +394,18 @@ public void verifyValue(@Nonnull String codeList, @Nullable String code) {
...
 
+    /**
+     * Returns the {@link CodeListEntry} from the given code list associated with the given code.
+     *
+     * @param codeList the code list to search in
+     * @param code     the code to lookup
+     * @return the entry associated with the code or an empty optional otherwise
+     */
+    public Optional<E> getEntry(@Nonnull String codeList, String code) {
+        L cl = findOrCreateCodelist(codeList);
```
sabieber: Maybe call list either `list` or `codeList` and the Param above `codeListName`?

src/main/resources/default/templates/biz/jobs/params/codelistentry.html.pasta
```diff
@@ -0,0 +1,14 @@
+<i:arg type="sirius.biz.jobs.params.CodeListEntryParameter" name="param"/>
```
sabieber: I think its better to mention the param name before listing the type as this information is more important to see at a glance.

PR #664, opened 2020-01-17
Allows for a BooleanParameter to be nullable/optional

idlira:
There are cases where we need to know if the parameter is selected at all and not always default to false.
Default value is false unless markNullable is used.

Fixes: OX-5522

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/BooleanParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/BooleanParameter.java
@@ -82,7 +82,7 @@ public class BooleanParameter extends Parameter<Boolean, BooleanParameter> {
 
     @Override
     protected Optional<Boolean> resolveFromString(@Nonnull Value input) {
-        if (nullable && !("true".equalsIgnoreCase(input.asString()) || "false".equalsIgnoreCase(input.asString()))) {
+        if (nullable && !input.isFilled()) {
             return Optional.empty();
         }
         return Optional.of(input.asBoolean());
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/jobs/params/BooleanParameter.java
```diff
@@ -47,18 +50,41 @@ protected String checkAndTransformValue(Value input) {
...
 
     @Override
-    protected Optional<Boolean> resolveFromString(Value input) {
+    public boolean isRequired() {
+        return !nullable;
+    }
+
+    @Override
+    protected Optional<Boolean> resolveFromString(@Nonnull Value input) {
+        if (nullable && !("true".equalsIgnoreCase(input.asString()) || "false".equalsIgnoreCase(input.asString()))) {
```
meg1502: this does not match the parsing of input.asBoolean which is used when the parameter is required, as $NLS.yes is also considerd as true (and $NLS.no as false)
->
if the parameter is required i will get true for "yes"
if the parameter is required i will get null for "yes"

meg1502: my suggestion:
```suggestion
        if (nullable && !input.isFilled()) {
```

idlira: Cannot use input.asBoolean here (is actually the whole point of it), because it evaluates anything else as false. In this usage case (and the actual values of the selectBoolean template) only allow (and expect) true or false.

meg1502: but "yes" and "no" should still be parsed as booleans, and null should possibly only be returned if the value is not filled at all?

PR #665, opened 2020-01-18
Job parameter enhancements

idlira:
- Adds the ability for StringParameters to be nullable
- Adds a new StringList Parameter which can be used to render a single select of key-value string pairs

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/StringListParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/StringListParameter.java
@@ -5,6 +5,7 @@ import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Value;
 import sirius.kernel.nls.NLS;
 
+import javax.annotation.Nonnull;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -32,11 +33,11 @@ public class StringListParameter extends Parameter<String, StringListParameter> 
      * Adds an entry to the list.
      *
      * @param key   the entry key
-     * @param value the display value
+     * @param value the display value, which will be {@link NLS#smartGet(String) auto translated}
      * @return the parameter itself for fluent method calls
      */
     public StringListParameter withEntry(String key, String value) {
-        this.entries.put(key, value);
+        this.entries.put(key, NLS.smartGet(value));
         return self();
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/biz/jobs/params/StringParameter.java
```diff
@@ -28,8 +30,35 @@ public StringParameter(String name, String label) {
...
+     */
+    @Override
+    public StringParameter markRequired() {
+        this.nullable = false;
+        return super.markRequired();
+    }
+
     @Override
     protected String checkAndTransformValue(Value input) {
+        if (nullable && input.isEmptyString()) {
```
andyHa: shouldn't this be the default? null instead of "" - even ifthe parameter isn marked as nullable it still might yield null as Value.getString() does filter null - asString would do this?

idlira: I like that! If we agree on returning null instead of "" for a zero-length input, then there is no need for a markNullable method.

src/main/java/sirius/biz/jobs/params/StringListParameter.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        return input.asString();
+    }
+
+    @Override
+    protected Optional<String> resolveFromString(@Nonnull Value input) {
+        if (!entries.containsKey(input.asString())) {
+            return Optional.empty();
+        }
+        return Optional.of(entries.get(input.asString()));
```
meg1502: shouldn't i be receiving the whole Tuple in the job? There should be a lot of instance where the display value is never used in the job

src/main/java/sirius/biz/jobs/params/StringListParameter.java
```diff
@@ -0,0 +1,76 @@
...
+    protected String checkAndTransformValue(Value input) {
+        if (Strings.isEmpty(input) || !entries.containsKey(input.asString())) {
+            return null;
+        }
+        return input.asString();
+    }
+
+    @Override
+    protected Optional<String> resolveFromString(@Nonnull Value input) {
+        if (!entries.containsKey(input.asString())) {
```
meg1502: shouldnt this also return empty for null/"" ?

idlira: null.asString resolves to "" and there will likely be no key inside the map matching an empty string.

src/main/java/sirius/biz/jobs/params/StringListParameter.java
idlira: Refactored to SingleSelectParameter

src/main/java/sirius/biz/jobs/params/StringParameter.java
idlira: After analysis, makeNullable made obsolete and defaults to null

PR #667, opened 2020-01-21
Adds a method to retrieve the query from the page helper

idlira:
Useful to be used by Batch Jobs to reuse the current query/filters from a page.

### DIFF
--- a/src/main/java/sirius/biz/web/BasePageHelper.java
+++ b/src/main/java/sirius/biz/web/BasePageHelper.java
@@ -392,6 +392,19 @@ public abstract class BasePageHelper<E extends BaseEntity<?>, C extends Constrai
         });
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -392,6 +388,19 @@ public void iterateAllTotalItems(Consumer<E> consumer) {
         });
     }
 
+    /**
+     * Returns the underlying query for this page helper including all given facets.
+     *
+     * @return the {@link Query} object
+     */
+    public Q buildUnderlyingQuery() {
```
andyHa: couldn't we then kill iterate and iterateAll as these as 100% redundant to query.iteratAll etc...

idlira: Sound logical. Will rip'em off...

PR #669, opened 2020-01-21
Adds new creation methods for MutableVirtualFile to offer a safe variant 

meg1502:
OX-5616


src/main/java/sirius/biz/storage/layer3/MutableVirtualFile.java
```diff
@@ -40,15 +42,44 @@ protected MutableVirtualFile() {
...
+     * @param parent the parent of the file
+     * @param name   the name of the file
+     * @return a new instance of this class
+     */
+    @Nullable
+    public static MutableVirtualFile safeCreate(@Nonnull VirtualFile parent, @Nullable String name) {
+        if (Strings.isEmpty(name)) {
+            return null;
+        }
+        return new MutableVirtualFile(parent, name.replace('/', '_'));
```
idlira: Do we need to make sure these file names are also Windows-compatible? If so, we should include a few more characters to the list.

- Linux/Unix: / (forward slash)
- Windows: < (less than) > (greater than) : (colon - sometimes works, but is actually NTFS Alternate Data Streams) " (double quote) / (forward slash) \ (backslash) | (vertical bar or pipe) ? ( question mark) * (asterisk)

sabieber: I think we only have to regard the SFTP restrictions for filenames

andyHa: Maybe blocking some more restricted characters would be a sane approach and wouldn't hurt too much(?)

https://en.wikipedia.org/wiki/Filename has a good table - and everythign except . and " " seems dangerous. Maybe we provide three methods one that crashes, one that replaces and one that accepts everything (while trusting the caller) to properly check and format the string(?)

src/main/java/sirius/biz/storage/layer3/MutableVirtualFile.java
```diff
@@ -34,19 +37,66 @@
...
+        super(parent, name);
+    }
+
+    /**
+     * Creates a new file with the given name in the given directory.
      *
      * @param parent the parent of the file
      * @param name   the name of the file
+     * @return a new instance of this class
+     * @throws IllegalArgumentException if the given name is empty or contains an illegal char
```
andyHa:
```suggestion
     * @throws IllegalArgumentException if the given name is empty or contains an {@link #REGEX_ILLEGAL_FILE_CHARS illegal} char
```

src/main/java/sirius/biz/storage/layer3/MutableVirtualFile.java
```diff
@@ -34,19 +37,66 @@
...
+        if (Strings.isEmpty(name) || Pattern.compile(REGEX_ILLEGAL_FILE_CHARS).split(name).length > 1) {
+            throw new IllegalArgumentException("A filename must be filled and must not contain illegal charachters.");
+        }
+        return new MutableVirtualFile(parent, name);
+    }
+
+    /**
+     * Creates a new file with the given name in the given directory.
+     * <p>
+     * Replaces any illegal chars in the given String and if the given String is empty, <tt>null</tt> is returned.
```
andyHa:
```suggestion
     * Replaces any {@link #REGEX_ILLEGAL_FILE_CHARS illegal} chars in the given String and if the given String is empty, <tt>null</tt> is returned.
```

src/main/java/sirius/biz/storage/layer3/MutableVirtualFile.java
```diff
@@ -34,19 +37,66 @@
...
+            return null;
+        }
+        return new MutableVirtualFile(parent, name.replaceAll(REGEX_ILLEGAL_FILE_CHARS, "_"));
+    }
+
+    /**
+     * Creates a new file with the given name in the given directory.
+     * <p>
+     * This method should only be used if you know what you are doing. In most cases use either {@link #checkedCreate(VirtualFile, String)}
+     * or {@link #safeCreate(VirtualFile, String)} to account for illegal characters.
```
andyHa:
```suggestion
     * or {@link #safeCreate(VirtualFile, String)} to account for {@link #REGEX_ILLEGAL_FILE_CHARS illegal} characters.
```

src/main/java/sirius/biz/storage/layer3/MutableVirtualFile.java
```diff
@@ -34,19 +37,66 @@
      * Provides a prediacte which is always <tt>false</tt>.
      */
     public static final Predicate<VirtualFile> CONSTANT_FALSE = ignored -> false;
+    private static final String REGEX_ILLEGAL_FILE_CHARS = "[/\\\\?%*:|\"<>]";
```
andyHa: Add javadoc here and explain here these characters came from (e.g. link to https://en.wikipedia.org/wiki/Filename)

PR #670, opened 2020-01-23
Provides a helper to conveniently check permissions on an entity with…

andyHa:
…out relying on a UserManager or the like to transform them.

In some cases we only have the database entity at hands and need to reason about granted permissions. We still want to
utilize the profiles and other facilities to conveniently work with permissions. This is now provided by getEffectivePermissions
and hasPermission.

### DIFF
--- a/src/main/java/sirius/biz/model/PermissionData.java
+++ b/src/main/java/sirius/biz/model/PermissionData.java
@@ -115,7 +115,7 @@ public class PermissionData extends Composite {
      * <p>
      * Also note that this is a helper method for rare cases, a call to
      * {@link sirius.web.security.UserInfo#hasPermission(String)} should always be preferred (if appropriate) as this
-     * will most probably way more efficient.
+     * will most probably be way more efficient.
      *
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/model/PermissionData.java
```diff
@@ -82,6 +89,43 @@ public StringList getPermissions() {
...
+    /**
+     * Checks if the requested permission expression is satisfied by the permissions granted in
+     * {@link #getPermissions()}.
+     * <p>
+     * Note that this uses {@link #getEffectivePermissions()} which applies all profiles to the granted permissions,
+     * which will be cached in this entity and will not reflect changes to {@link #getPermissions()}.
+     * <p>
+     * Also note that this is a helper method for rare cases, a call to
+     * {@link sirius.web.security.UserInfo#hasPermission(String)} should always be preferred (if appropriate) as this
+     * will most probably way more efficient.
```
sabieber:
```suggestion
     * will most probably be way more efficient.
```

PR #673, opened 2020-01-24
Improves the detection of broken FTP connections

andyHa:
...also provides some tooling to inspect the pool of uplink connectors


src/main/java/sirius/biz/storage/layer3/uplink/util/UplinkConnectorPool.java
```diff
@@ -8,9 +8,11 @@
 
 package sirius.biz.storage.layer3.uplink.util;
 
+import org.apache.commons.pool2.impl.DefaultPooledObject;
```
tbiScireum: unnecessary imports

PR #675, opened 2020-01-27
Makes the required check aware of the automatic trim performed by mos…

andyHa:
…t of the frameworks around it.

### DIFF
--- a/src/main/java/sirius/biz/importer/format/RequiredCheck.java
+++ b/src/main/java/sirius/biz/importer/format/RequiredCheck.java
@@ -8,19 +8,43 @@
 
 package sirius.biz.importer.format;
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/importer/format/RequiredCheck.java
```diff
@@ -8,21 +8,45 @@
...
+ * Use {@link #checkUntrimmed()} in order to suppress this behaviour.
  */
 public class RequiredCheck implements ValueCheck {
 
+    private boolean trim = true;
+
     @Override
     public void perform(Value value) {
-        if (value.isEmptyString()) {
+        String effectiveValue = value.asString();
```
sabieber: fetch the raw value

PR #688, opened 2020-02-20
Mongo Searchable Prefixes via BiConsumer

idlira:
Offers the ability to create Transformers to offer an alternative way to collect values to index from Properties, where the standard `getValue()` is not applicable.

Also adds the _Optimus Prime_ , _Megatron_ and _Bumblebee_ transformers for MultiLanguageString, StringList and StringMap respectively 😉


src/main/java/sirius/biz/mongo/MultiLanguageStringPropertyTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+    public Class<PrefixSearchableContentSupplier> getTargetClass() {
+        return PrefixSearchableContentSupplier.class;
+    }
+
+    @Nullable
+    @Override
+    @SuppressWarnings("unchecked")
+    public PrefixSearchableContentSupplier make(@Nonnull MultiLanguageStringProperty source) {
+        return entity -> {
+            StringList tokens = new StringList();
```
andyHa: why this intermediate step, couldn't we transform into a BiConsumer which accepts an entity and a StringBuilder or even BiConsumer<PrefixSearchableEntity, Consumer<String>> and then invoke with (entity, this::addContentString) or the like

andyHa: or simply transform to String

idlira: We index not only the words, but the phrase as well (see test case)
Indigo blue becomes: ["indigo blue", "indigo", "blue"]

For Multi-Language scenario we want each translation "phrase" to be processed independently and not only its loose words...

idlira: Implemented as `BiConsumer`. Hopefully nailed it now 😬

src/main/java/sirius/biz/mongo/PrefixSearchableContentConsumer.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+/**
+ * Used as target class to transform {@link sirius.db.mixing.Property properties} into a {@link BiConsumer}.
+ * <p>
+ * The consumer takes the {@link PrefixSearchableEntity} containing the values to extract which are then
+ * passed as the {@link Consumer} argument.
+ */
+public interface PrefixSearchableContentConsumer extends BiConsumer<PrefixSearchableEntity, Consumer<Object>> {
```
andyHa:
```suggestion
public interface PrefixSearchableContentConsumer extends BiConsumer<PrefixSearchableEntity, Consumer<String>> {
```

idlira: +1

PR #692, opened 2020-02-21
Revert "Set length of tennantId and userId to 26"

meg1502:
This reverts commit 6cdf803d

As the user_id for users created by MongoTenantUserManager has a length of 43 this composite did not work.
This composite should provide a more loose length for ids and not restrict it as its has no connection to a specific UserManager Implementation and therefore should handle a variety of sensible lengths for ids.

### DIFF
--- a/src/main/java/sirius/biz/analytics/events/UserData.java
+++ b/src/main/java/sirius/biz/analytics/events/UserData.java
@@ -27,7 +27,7 @@ public class UserData extends Composite {
      * Stores the ID of the current user.
      */
     public static final Mapping USER_ID = Mapping.named("userId");
-    @Length(26)
+    @Length(255)
     @NullAllowed
     private String userId;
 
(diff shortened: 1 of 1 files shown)


andyHa:
Müssen wir mit S2 abklären, da ist ne Migration fällig

src/main/java/sirius/biz/analytics/events/UserData.java
```diff
@@ -27,15 +27,15 @@
      * Stores the ID of the current user.
      */
     public static final Mapping USER_ID = Mapping.named("userId");
-    @Length(26)
+    @Length(255)
```
andyHa: remove length entirely

src/main/java/sirius/biz/analytics/events/UserData.java
```diff
@@ -27,15 +27,15 @@
...
     @NullAllowed
     private String userId;
 
     /**
      * Stores the ID of the current tenant.
      */
     public static final Mapping TENANT_ID = Mapping.named("tenantId");
-    @Length(26)
+    @Length(255)
```
andyHa: remove length entirely

PR #693, opened 2020-02-25
Defines a default Index including  the tenant and searchPrefixes

idlira:
Which will be created for all MongoTenantAware entities.

⚠️ **MongoTenantAware entities must be checked for indexes including tenant and either dropped (to be recreated by this class) or have the creation suppressed by defining an index without columns.**

Fixes: SIRI-184

### DIFF
--- a/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
+++ b/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
@@ -29,7 +29,7 @@ import java.util.regex.Pattern;
 
 /**
  * Maintains a <tt>prefixSearchField</tt> in which all fields annotated with {@link PrefixSearchContent} are indexed.
- *
+ * <p>
  * Note that the <tt>searchPrefixes</tt> is not automatically indexed as it usually only makes sense if indexed
  * in combination with other field(s). Therefore make sure to include this field when extending this class into
  * other {@link MongoEntity entities} where applicable. {@link sirius.biz.tenants.mongo.MongoTenantAware} is a good
--- a/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
+++ b/src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
@@ -29,8 +29,12 @@ import java.util.regex.Pattern;
 
 /**
  * Maintains a <tt>prefixSearchField</tt> in which all fields annotated with {@link PrefixSearchContent} are indexed.
+ *
+ * Note that the <tt>searchPrefixes</tt> is not automatically indexed as it usually only make sense if indexed
+ * in combination with other field(s). Therefore make sure to include this field when extending this class into
+ * other {@link MongoEntity entities} where applicable. {@link sirius.biz.tenants.mongo.MongoTenantAware} is a good
+ * example.
  */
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/biz/tenants/mongo/MongoTenantAware.java
```diff
@@ -11,14 +11,23 @@
...
 import java.util.Optional;
 
 /**
  * Base class which marks subclasses as aware of their tenant they belong to.
+ * <p>
+ * Note that an index is automatically created containing the tenant itself and the searchPrefixes,
+ * which are added via {@link MongoBizEntity}. You can skip the index creation by defining an {@link Index}
+ * without columns.
  */
+@Index(name = "tenant_searchPrefixes",
```
andyHa: shouldn't this be named prefix_index to overwrite the index defined by PrefixSearchable Entity?

otherwise we end up with two entities where the original one will almost never be used(?)

idlira: The real index name is `searchPrefixes_1` since it wasn't using the name property of the annotation. All indexes must be dropped and recreated anyway, therefore I would include the word **tenant** in the name to better indicate how the index is build up.

andyHa: i meant name it so that the parent index as defined by PrefixSearchableEntity is suppressed - although a misleading name in mongo would be created then - still better than a mostly unused index

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -29,8 +29,12 @@
 
 /**
  * Maintains a <tt>prefixSearchField</tt> in which all fields annotated with {@link PrefixSearchContent} are indexed.
+ *
+ * Note that the <tt>searchPrefixes</tt> is not automatically indexed as it usually only make sense if indexed
```
sabieber:
```suggestion
 * Note that the <tt>searchPrefixes</tt> is not automatically indexed as it usually only makes sense if indexed
```

src/main/java/sirius/biz/tenants/mongo/MongoTenantAware.java
```diff
@@ -11,14 +11,23 @@
...
 import sirius.kernel.di.std.Part;
 
 import java.util.Optional;
 
 /**
  * Base class which marks subclasses as aware of their tenant they belong to.
+ * <p>
+ * Note that an index is automatically created containing the tenant itself and the searchPrefixes,
+ * which are added via {@link MongoBizEntity}. You can skip the index creation by defining an {@link Index}
+ * without columns.
```
sabieber:
```suggestion
 * with the same name and without columns.
```

src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -29,8 +29,12 @@
 
 /**
  * Maintains a <tt>prefixSearchField</tt> in which all fields annotated with {@link PrefixSearchContent} are indexed.
+ *
```
jmuscireum:
```suggestion
 * <p>
```

PR #695, opened 2020-02-27
Completes the list operation after a successful attempt.

andyHa:
As there is no implicit return, we have to perform this manually here
(in contrast to all other operations).

Fixes: SIRI-185


src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPUplink.java
```diff
@@ -99,6 +100,8 @@ private boolean checkForMLSD(FTPClient client) {
     }
 
     @Override
+    @SuppressWarnings("squid:S3626")
+    @Explain("False positive, the return is necessarry to prevent retries if an attempt was successful.")
```
bwiedmann:
```suggestion
    @Explain("False positive, the return is necessary to prevent retries if an attempt was successful.")
```

PR #697, opened 2020-03-04
Adds VirtualFile.download and uses it in the FileImportJob

andyHa:


src/main/java/sirius/biz/storage/layer3/uplink/sftp/SFTPUplink.java
```diff
@@ -8,6 +8,7 @@
 
 package sirius.biz.storage.layer3.uplink.sftp;
 
+import com.google.common.io.BaseEncoding;
```
tbiScireum: not required?

PR #702, opened 2020-03-16
Always add the name and the label as alias

tbiScireum:
We tried to fix this issue in SIRI-136 but we provided the fix only for autoImported fields.
This solution will also work if the FieldDefinition is added manually, e.g. if it was added by an overridden method
of EntityImportJobFactory#enhanceDictionary

This will prevent us from writing code looking like:
`dictionary.addField(FieldDefinition.stringField(FIELD_NAME, 1).addAlias(FIELD_NAME);`

- Fixes: SE-9397

### DIFF
--- a/src/main/java/sirius/biz/importer/format/FieldDefinition.java
+++ b/src/main/java/sirius/biz/importer/format/FieldDefinition.java
@@ -48,6 +48,8 @@ public class FieldDefinition {
     public FieldDefinition(String name, String type) {
         this.name = name;
         this.type = type;
+
+        addAlias(getName());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/importer/format/FieldDefinition.java
```diff
@@ -239,6 +240,7 @@ public FieldDefinition withLabel(String constantLabel) {
      */
     public FieldDefinition withLabel(Supplier<String> labelFunction) {
         this.label = labelFunction;
+        addAlias(getLabel());
```
andyHa: This will call the label function right away instead of when the user context is all set-up and the proper language is known.

If we _always_ create a new FieldDefinition, we can get away with this - but then we can simply accept a translated string where. Otherwise calling getLabel here is too early which is why it was moved into expandAliases...

andyHa:
The provided solution also doesn't take real aliases into account. Originally the idea of `EntityImportJobFactory#enhanceDictionary` was to merge two existing dictionaries which are in itself consistent. Manually building FieldDefinitions for existing properties is most probably an indicator of a design problem.

PR #703, opened 2020-03-18
Fix FTP Filetype and Upload bug

mkeckmkeck:
Remove local file exists check when creating OutputStreams for S/FTP
Set FTP FileType to Binary instead of default ASCII

Fixes: SIRI-194


src/main/java/sirius/biz/storage/layer3/uplink/sftp/SFTPUplink.java
```diff
@@ -135,7 +135,6 @@ private MutableVirtualFile wrap(VirtualFile parent, SftpClient.DirEntry file, St
               .withDeleteHandler(this::deleteHandler)
               .withCanProvideInputStream(this::isExistingFile)
               .withInputStreamSupplier(this::inputStreamSupplier)
-              .withCanProvideOutputStream(this::isExistingFile)
```
andyHa: could you check all other Uplinks as well - just to be sure...

mkeckmkeck: Found one more, please check if this has to be removed.

PR #706, opened 2020-03-25
SE-9475 Adds the country translations from OXOMI

jmuscireum:
These are also needed in SellSite.


src/main/resources/biz_de.properties
```diff
@@ -68,6 +68,58 @@ ContactData.fax = FAX
 ContactData.invalidPhone = Fehler im Feld '${field}': '${value}' is keine gültige Nummer.
```
ymo-sci:
```suggestion
ContactData.invalidPhone = Fehler im Feld '${field}': '${value}' ist keine gültige Nummer.
```

PR #713, opened 2020-04-03
Doesn't update TraceData fields when running a resave job.

andyHa:
Most commonly one doesn't want to change the data in these fields,
as the job is mainly intended to be used for checks rather than for
real updates.


src/main/java/sirius/biz/model/ResaveEntitiesJobFactory.java
```diff
@@ -92,6 +94,15 @@ private void processEntity(BaseEntity<?> entity) {
             Watch watch = Watch.start();
             try {
                 if (executeSave) {
+                    // We do not want to update the tracing infos here as we most probably don't change anything.
+                    //
+                    // We do not suppress the Journal (if there is any) because we do not
+                    // expect any journalled changes. If there still are any, we have to log them
+                    // for traceability reasons.
+                    if (entity instanceof Traced) {
```
sabieber, line 102: Please also do this for journaldata :)

PR #719, opened 2020-04-10
Improvements for the query explorer

andyHa:


src/main/java/sirius/biz/jdbc/DatabaseController.java
```diff
@@ -89,33 +89,39 @@ public void sql(WebContext ctx) {
...
-            if (monoflop.successiveCall()) {
-                out.endArray();
+        try {
+            String database = ctx.get("db").asString(defaultDatabase);
+            Database db = determineDatabase(database);
+            String sqlStatement = ctx.get("query").asString();
+            SQLQuery qry = db.createQuery(sqlStatement);
+
+            if (isDDSStatement(sqlStatement)) {
+                // To prevent accidential damage, we try to filter DDS queries (modifying the database structure) and
```
idlira, line 99: What does DDS means? I know the term DDL (Data Definition Language).

bwiedmann: DDS --> DDL?

andyHa, line 99:
```suggestion
                // To prevent accidential damage, we try to filter DDL queries (modifying the database structure) and
```

PR #727, opened 2020-04-21
SE-9561 save entity with unchanged blob

mkeckmkeck:
- Bugfix for an issue that happened on the dev-17.0.3 tag code base that is currently productive at Arge and will have to be patched outside of regular release cycle. 
- Compares actual entities as entities mapped from DB were always handles as changed in the past.


src/test/java/sirius/biz/storage/layer2/BlobHardRefSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+    private static BlobStorage blobStorage;
+
+    def "store entity with unchanged hard ref blob must not fail"() {
+        given:
+        def testEntity = new BlobHardRefEntity()
+        Blob blob = blobStorage.getSpace("blob-files").createTemporaryBlob()
+        testEntity.getTheBlobRef().setBlob(blob)
+        oma.update(testEntity)
+        when:
+        def loadedEntity = oma.
```
andyHa, line 30: oma.refresh oder or oma.findOrFail

src/main/java/sirius/biz/storage/layer2/BlobRefProperty.java
```diff
@@ -0,0 +1,126 @@
...
+import java.lang.reflect.Field;
+import java.sql.Types;
+
+/**
+ * Holder of common parts of blob reference property handling.
+ * <p>
+ * NOTE: This implementation assumes that BlobHardRef stays the parent class for all blob reference
+ * property classes, which should be valid because hard and soft references should be enough.
+ */
+public abstract class BlobRefProperty extends Property implements SQLPropertyInfo {
```
andyHa, line 32: protected or package visibility?

src/main/java/sirius/biz/storage/layer2/BlobRefProperty.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public void contributeToTable(Table table) {
+        table.getColumns().add(new TableColumn(this, Types.CHAR));
+    }
+
+    /**
+     * Checks if the blob reference on the entity has changed since loading from DB.
+     *
+     * @param entity the entity to check
+     * @return <tt>true</tt> if unchanged, else <tt>false</tt>
```
andyHa, line 120:
```suggestion
     * @return <tt>true</tt> if changed, <tt>false</tt> otherwise
```

andyHa: the first part is the real change - second is just the common way to express the "else" part :)

src/test/java/sirius/biz/storage/layer2/BlobHardRefSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+
+class BlobHardRefSpec extends BaseSpecification {
+
+    @Part
+    private static OMA oma;
+
+    @Part
+    private static BlobStorage blobStorage;
+
+    def "store entity with unchanged hard ref blob must not fail"() {
```
andyHa, line 23: while being at it - maybe add another test for the happy path (meaning the given part of the test below as when: and in then: simply ensure that the blob is no longer temporary)

and then delete the entity and check that the blob is marked as deleted :) (either in one case or in two cases)

src/test/resources/test.conf
```diff
@@ -182,6 +182,13 @@ storage {
             reply-secondary {
                 engine = "s3"
             }
+            blob-files {}
+        }
+    }
+    layer2.spaces {
+        blob-files {
+                readPermission = "disabled"
+                description = "blob-files {}"
```
andyHa, line 191: just for the record - the description isn't necessary in test configs :)

src/main/java/sirius/biz/storage/layer2/BlobRefProperty.java
mkeckmkeck: done

src/test/java/sirius/biz/storage/layer2/BlobHardRefSpec.groovy
mkeckmkeck: done

src/main/java/sirius/biz/storage/layer2/BlobRefProperty.java
mkeckmkeck: done

src/test/resources/test.conf
mkeckmkeck: done

src/test/java/sirius/biz/storage/layer2/BlobHardRefSpec.groovy
mkeckmkeck: done

PR #732, opened 2020-04-23
Adds an event for blob conversion analytics

sabieber:
Fixes: SIRI-206

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/variants/ConversionEvent.java
+++ b/src/main/java/sirius/biz/storage/layer2/variants/ConversionEvent.java
@@ -9,7 +9,6 @@
 package sirius.biz.storage.layer2.variants;
 
 import sirius.biz.analytics.events.Event;
-import sirius.biz.analytics.events.UserData;
 import sirius.db.mixing.Mapping;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/storage/layer2/variants/ConversionEvent.java
```diff
@@ -0,0 +1,179 @@
...
+    private String sourceFileName;
+
+    /**
+     * Contains the file type / extension of the file used as a source for the conversion.
+     */
+    public static final Mapping SOURCE_TYPE = Mapping.named("sourceType");
+    private String sourceType;
+
+    /**
+     * Contains the database ID (if available) of the file used as a source for the conversion.
```
andyHa, line 40:
```suggestion
     * Contains the database ID (if available) of the file used as the target or destination for the conversion.
```

src/main/java/sirius/biz/storage/layer2/variants/ConversionEvent.java
```diff
@@ -0,0 +1,179 @@
...
+     */
+    public static final Mapping DURATION = Mapping.named("duration");
+    private long duration = 0;
+
+    /**
+     * Contains the tenant owning the source and target files being converted.
+     */
+    public static final Mapping TENANT_ID = Mapping.named("tenantId");
+    @NullAllowed
+    private String tenantId;
```
meg1502, line 90: in the other 2 events with tenants already defined in biz, this field is simply called "tenant"  - unify these?

PR #772, opened 2020-06-08
SIRI-224: Added copy path action to file drop down menu

marvinwoop:
- Fixes: SIRI-224


src/main/resources/default/templates/biz/storage/list.html.pasta
```diff
@@ -205,5 +214,18 @@
                 modalElement.find("#select-file-modal-search-filter").focus();
             }).modal('show');
         }
+
+        function copyToClipboard(value) {
+            /*!
+             * inspired by OXOMI´s implementation of oxomi.copyToClipboard()
```
andyHa, line 220: I'd rather put that into a commit comment - the OXOMI repo is closed source anyway..

PR #774, opened 2020-06-10
Greatly improves the SAML login experience in the backend.

andyHa:


andyHa:
> Soll der keystore.jks eingecheckt werden?

was a self signed key for "localhost" - but the config for it resides in
instance.conf - therefore it is now removed again, we already have a
"localhost.jks" in .gitignore which can be used and customized so that
one can support SSL locally.

PR #783, opened 2020-06-16
SE-9717: Added possibility to remove fields from ImportDictionary

marvinwoop:
Corresponding aliases will be deleted too

- Fixes: SE-9717

### DIFF
--- a/src/main/java/sirius/biz/importer/BaseImportHandler.java
+++ b/src/main/java/sirius/biz/importer/BaseImportHandler.java
@@ -333,6 +333,7 @@ public abstract class BaseImportHandler<E extends BaseEntity<?>> implements Impo
      * <p>
      * This method can be overwritten to suppress properties which wear an <tt>AutoImport</tt> annotation
      * from being put into the <tt>ImportDictionary</tt>.
+     * <p>
      * By default, this will constantly return <tt>true</tt>.
      *
      * @param mapping the mapping to evaluate
--- a/src/main/java/sirius/biz/importer/BaseImportHandler.java
+++ b/src/main/java/sirius/biz/importer/BaseImportHandler.java
@@ -329,13 +329,14 @@ public abstract class BaseImportHandler<E extends BaseEntity<?>> implements Impo
     }
 
     /**
-     * Determines if mapping should be considered.
+     * Determines if an auto imported mapping should be put into the <tt>ImportDictionary</tt>.
      * <p>
-     * This method returns true by default and is intended to be overwritten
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -324,9 +324,23 @@ public ImportDictionary getImportDictionary() {
                                                      .orElse(false))
                          .map(Property::getName)
                          .map(Mapping::named)
+                         .filter(this::isAutoImportMappingAccepted)
                          .collect(Collectors.toList());
     }
 
+    /**
+     * Determines if mapping should be considered.
```
andyHa, line 332:
```suggestion
     * Determines if an auto imported mapping should be put into the <tt>ImportDictionary</tt>.
```

src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -324,9 +324,23 @@ public ImportDictionary getImportDictionary() {
...
+     * <p>
+     * This method returns true by default and is intended to be overwritten
```
andyHa, line 334:
```suggestion
     * This method can be overwritten to suppress properties which wear an <tt>AutoImport</tt> annotation
     * from being put into the import dictionary.
```

src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -324,9 +324,23 @@ public ImportDictionary getImportDictionary() {
...
+     * by sub classes if specific mappings should be ignored.
+     *
+     * @param mapping the mapping to evaluate
+     * @return true by default
```
andyHa, line 338:
```suggestion
     * @return <tt>true</tt> to accept the mapping, <tt>false</tt> otherwise
```

src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -324,9 +324,23 @@ public ImportDictionary getImportDictionary() {
...
                          .map(Mapping::named)
+                         .filter(this::isAutoImportMappingAccepted)
                          .collect(Collectors.toList());
     }
 
+    /**
+     * Determines if mapping should be considered.
+     * <p>
+     * This method returns true by default and is intended to be overwritten
+     * by sub classes if specific mappings should be ignored.
```
andyHa, line 335:
```suggestion
     * By default, this will constantly return <tt>true</tt>.
```

PR #784, opened 2020-06-16
Fixes alias handling for ImportHandlers.

andyHa:
We now provide the alias (field name) earlier (when transforming a Property to a FieldDefinition) so that the alias is also shown in the UI.


src/main/java/sirius/biz/storage/layer3/BlockwiseIterator.java
```diff
@@ -0,0 +1,59 @@
...
+package sirius.biz.storage.layer3;
+
+import sirius.kernel.commons.Limit;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+class BlockwiseIterator implements Iterator<VirtualFile> {
```
jmuscireum, line 18: Gehört zu #785

PR #788, opened 2020-06-17
Uses the Entity interface provided by sirius-db and some small fixes.

andyHa:


src/main/java/sirius/biz/tenants/UserAccount.java
```diff
@@ -44,32 +44,6 @@
      */
     UserAccountData getUserAccountData();
 
-    /**
```
jmuscireum, line 47: This interface doesn't extend Entity yet

jmuscireum: Or maybe RateLimitedEntity should extend Entity instead?

mkeckmkeck: Kommt von TenantAware

PR #808, opened 2020-07-02
Introduction of translation entities

cschierle:
Adds a generic translation entity that can be included in any type of entity by implementing the provided Translatable interface. Any field of an entity using this interface can be translated into any of the supported languages.

caches will be added in a next stage to reduce the amount of database requests


src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,105 @@
...
+
+        translation.getTranslationData().setText(text);
+
+        mango.update(translation);
+    }
+
+    @Override
+    public void deleteText(Mapping field, String lang) {
+        Optional<MongoTranslation> translation = fetchTranslation(field, lang);
+        translation.ifPresent(mongoTranslation -> mango.delete(mongoTranslation));
```
qw3ry, line 47:
```suggestion
        fetchTranslation(field, lang).ifPresent(mango::delete);
```

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param lang  language code
+     * @return the resolved translation entity wrapped as optional or an empty optional if the value couldn't be resolved
+     */
+    protected abstract Optional<T> fetchTranslation(Mapping field, String lang);
+
+    /**
+     * Looks up and returns all available {@link Translation translation} entities for the given field
+     *
+     * @param field {@link Mapping} of the translated field
+     * @return {@link Optional} with a {@link java.util.List} of all {@link Translation translations}, or empty if none are found
```
qw3ry, line 171: Why not empty list instead of empty optional? Saves a lot of extra code in the calls to this method

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,174 @@
...
+    public Map<String, String> getAllTexts(Mapping field) {
+        Optional<List<T>> translations = fetchAllTranslations(field);
+        if (!translations.isPresent()) {
+            return new HashMap<>();
+        }
+        Map<String, String> translatedTexts = new HashMap<>();
+        for (T t : translations.get()) {
+            translatedTexts.put(t.getTranslationData().getLang(), t.getTranslationData().getText());
+        }
+        return translatedTexts;
```
qw3ry, line 155: for example here:
```java
Map<String, String> translatedTexts = new HashMap<>();
for (T t : fetchAllTranslations(field)) {
    translatedTexts.put(t.getTranslationData().getLang(), t.getTranslationData().getText());
}
return translatedTexts;
```

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public void deleteAllTexts(Mapping field) {
+        Optional<List<SQLTranslation>> translations = fetchAllTranslations(field);
+        if (translations.isPresent()) {
+            for (SQLTranslation translation : translations.get()) {
+                oma.delete(translation);
+            }
+        }
```
qw3ry, line 57: Or here
```java
for (SQLTranslation translation : fetchAllTranslations(field)) {
    oma.delete(translation);
}
```

src/main/java/sirius/biz/translations/BasicTranslations.java
cschierle: Thanks for the hint! The optional was definitely unnecessary here

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -44,6 +45,12 @@ protected BasicTranslations(BaseEntity<?> owner) {
         this.owner = owner;
     }
 
+    /**
+     * Ensures that if the owner entity is deleted, all of its translations are deleted as well
```
idlira, line 49:
```suggestion
     * Ensures that translations are cascaded-deleted upon deletion of the owner entity.
```

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -29,6 +28,15 @@ public MongoTranslations(BaseEntity<?> owner) {
         super(owner);
     }
 
+    @Override
+    protected void deleteAllTranslations() {
+        for (MongoTranslation t : mango.select(MongoTranslation.class)
```
idlira, line 33: It's possible to delete all in a single command based on filter instead of looping and deleting one by one.

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -29,6 +28,15 @@ public SQLTranslations(BaseEntity<?> owner) {
         super(owner);
     }
 
+    @Override
+    protected void deleteAllTranslations() {
+        for (SQLTranslation t : oma.select(SQLTranslation.class)
```
idlira, line 33: It's possible to delete all in a single command based on filter instead of looping and deleting one by one.

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @param <T> the translation entity type used by a concrete subclass
+ */
+public abstract class BasicTranslations<T extends BaseEntity<?> & Translation> extends Composite {
+
+    @Transient
+    protected BaseEntity<?> owner;
+
+    @Transient
+    @ConfigValue("mixing.multiLanguageStrings.supportedLanguages")
```
andyHa, line 37: maybe use a more generic config name here?

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Fetches the default text by querying {@link sirius.db.mixing.FieldLookupCache} for the owner entity's given field.
+     *
+     * @param field {@link Mapping field} to fetch the the text from
+     * @return the text as String, or null if lookup was unsuccessful
+     */
+    protected String fetchDefaultText(Mapping field) {
+        return fieldLookupCache.lookup(owner.getClass(), owner.getId(), field).asString();
```
andyHa, line 144: why not directly read from the owner? owner.getDescriptor.getProperty(field).getValue(owner) or the like ? maybe check if the value is fetched and else fallback to the fieldLookupCache otherwise(??)

src/main/java/sirius/biz/translations/Translation.java
```diff
@@ -0,0 +1,30 @@
...
+import sirius.db.mixing.Mapping;
+
+/**
+ * Provides constants and methods that have to be present in any Translation entity
+ */
+public interface Translation {
+    /**
+     * Contains the unique name of the entity this translation belongs to.
+     */
+    Mapping OWNER = Mapping.named("owner");
```
andyHa, line 20: why not place this also in TranslationData?

src/main/java/sirius/biz/translations/jdbc/SQLTranslation.java
```diff
@@ -0,0 +1,35 @@
...
+import sirius.biz.translations.Translation;
+import sirius.biz.translations.TranslationData;
+import sirius.db.jdbc.SQLEntity;
+
+/**
+ * Stores translations as a table.
+ * <p>
+ * Note that translations should only be accessed via siblings of {@link sirius.biz.translations.BasicTranslations}.
+ */
+public class SQLTranslation extends SQLEntity implements Translation {
```
andyHa, line 20: index on owner, field, lang ??

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+    public void deleteAllTexts(Mapping field) {
+        for (SQLTranslation translation : fetchAllTranslations(field)) {
+            oma.delete(translation);
+        }
+    }
+
+    @Override
+    protected SQLTranslation findOrCreateTranslation(Mapping field, String lang, String text) {
+        if (!isSupportedLanguage(lang)) {
+            return null;
```
andyHa, line 65: exception(?)

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+            sqlTranslation.setOwner(owner.getUniqueName());
+            sqlTranslation.getTranslationData().setField(field.getName());
+            sqlTranslation.getTranslationData().setLang(lang);
+            return sqlTranslation;
+        }
+    }
+
+    @Override
+    protected Optional<SQLTranslation> fetchTranslation(Mapping field, String lang) {
+        if (field == null) {
```
andyHa, line 83: either mark as Nonnull and don't care or Objects.requireNonNull or the like - there is no sane API case for supporting null values here(?)

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+                                          .eq(Translation.TRANSLATION_DATA.inner(TranslationData.FIELD),
+                                              field.getName())
+                                          .eq(Translation.TRANSLATION_DATA.inner(TranslationData.LANG), lang)
+                                          .queryFirst());
+        }
+    }
+
+    @Override
+    protected List<SQLTranslation> fetchAllTranslations(Mapping field) {
+        if (field == null) {
```
andyHa, line 97: s.a.

src/main/java/sirius/biz/translations/mongo/MongoTranslation.java
```diff
@@ -0,0 +1,35 @@
...
+import sirius.biz.translations.Translation;
+import sirius.biz.translations.TranslationData;
+import sirius.db.mongo.MongoEntity;
+
+/**
+ * Stores translations as a collection.
+ * <p>
+ * Note that translations should only be accessed via siblings of {@link sirius.biz.translations.BasicTranslations}.
+ */
+public class MongoTranslation extends MongoEntity implements Translation {
```
andyHa, line 20: index

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    protected void removeTranslations() {
+        oma.select(SQLTranslation.class).eq(Translation.OWNER, owner.getUniqueName()).delete();
+    }
+
+    @Override
+    public void updateText(Mapping field, String lang, String text) {
+        SQLTranslation translation = findOrCreateTranslation(field, lang, text);
```
andyHa, line 38: handle empty -> delete before this...

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public void deleteText(Mapping field, String lang) {
+        fetchTranslation(field, lang).ifPresent(oma::delete);
+    }
+
+    @Override
+    public void deleteAllTexts(Mapping field) {
+        for (SQLTranslation translation : fetchAllTranslations(field)) {
```
andyHa, line 57: delete per oma.deleteStatement api?

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        translation.getTranslationData().setText(text);
+
+        oma.update(translation);
+    }
+
+    @Override
+    public void deleteText(Mapping field, String lang) {
+        fetchTranslation(field, lang).ifPresent(oma::delete);
```
andyHa, line 52: delete per oma.deleteStatementAPI + write to journal of owner if journaled

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
andyHa: + write to journal of owner if journaled

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
andyHa: and write to journal of owner (could be even done in the superclass)

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    protected void removeTranslations() {
+        mango.select(MongoTranslation.class).eq(Translation.OWNER, owner.getUniqueName()).delete();
+    }
+
+    @Override
+    public void updateText(Mapping field, String lang, String text) {
+        MongoTranslation translation = findOrCreateTranslation(field, lang, text);
```
andyHa, line 38: first delete then check + create otherwise

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+
+        if (Strings.isEmpty(text)) {
+            deleteText(field, lang);
+            return;
+        }
+
+        translation.getTranslationData().setText(text);
```
andyHa, line 45: update journal

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+
+        mango.update(translation);
+    }
+
+    @Override
+    public void deleteText(Mapping field, String lang) {
+        fetchTranslation(field, lang).ifPresent(mango::delete);
```
andyHa, line 52: mongo.delete() + journal

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public void deleteAllTexts(Mapping field) {
+        for (MongoTranslation translation : fetchAllTranslations(field)) {
+            mango.delete(translation);
+        }
+    }
```
andyHa, line 60: s.a.

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+            mongoTranslation.setOwner(owner.getUniqueName());
+            mongoTranslation.getTranslationData().setField(field.getName());
+            mongoTranslation.getTranslationData().setLang(lang);
+            return mongoTranslation;
+        }
+    }
+
+    @Override
+    protected Optional<MongoTranslation> fetchTranslation(Mapping field, String lang) {
+        if (field == null) {
```
andyHa, line 83: s.a.

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+            return Optional.empty();
+        } else {
+            return Optional.ofNullable(mango.select(MongoTranslation.class)
```
andyHa, line 86: thats what Query.first() (in contrast to Query.queryFirst()) does.. :)

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,107 @@
...
+                                            .eq(Translation.TRANSLATION_DATA.inner(TranslationData.FIELD),
+                                                field.getName())
+                                            .eq(Translation.TRANSLATION_DATA.inner(TranslationData.LANG), lang)
+                                            .queryFirst());
+        }
+    }
+
+    @Override
+    protected List<MongoTranslation> fetchAllTranslations(Mapping field) {
+        if (field == null) {
```
andyHa, line 97: s.a.

andyHa:
Should we also consider providing a cache for translated texts?

src/main/java/sirius/biz/translations/BasicTranslations.java
cschierle: this currently uses the existing supported languages config entry from the multi language strings implementation. It will be replaced once the standard codelists for countries/languages are implemented. The existing config should solve the chicken/egg-problem for now

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
cschierle: thanks for the hint!

cschierle:
> Should we also consider providing a cache for translated texts?

Caching would make a lot of sense and is planned for a future update. We will start without it, so the baseline of standardized codelists for countries/languages/currencies can be implemented first while making use of the new translation entities.

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -0,0 +1,131 @@
...
+                  .eq(Translation.TRANSLATION_DATA.inner(TranslationData.LANG), lang)
+                  .first();
+    }
+
+    @Override
+    protected List<SQLTranslation> fetchAllTranslations(@Nonnull Mapping field) {
+        return oma.select(SQLTranslation.class)
+                  .eq(Translation.TRANSLATION_DATA.inner(TranslationData.OWNER), owner.getUniqueName())
+                  .eq(Translation.TRANSLATION_DATA.inner(TranslationData.FIELD), field.getName())
+                  .limit(supportedLanguages.size())
```
andyHa, line 128: I'd skip that - this might silently suppress a bug we'd rather want to surface here....

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Deletes all translations for the given field.
+     * <p>
+     * Note that this will also emit an {@link sirius.biz.protocol.JournalEntry} to
+     * {@link sirius.biz.protocol.Journaled} owners keeping track of the changes made to their translations.
+     *
+     * @param field {@link Mapping} of the translated field
+     */
+    public abstract void deleteAllTexts(@Nonnull Mapping field);
```
andyHa, line 125: shouldn't we also place a @AfterDelete handler here, which wipes all texts (without logging to the journal)?

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    protected void updateTranslation(MongoTranslation translation) {
+        mango.update(translation);
+    }
+
+    @Override
+    public void deleteText(@Nonnull Mapping field, String lang) {
+        fetchTranslation(field, lang).ifPresent(mongoTranslation -> {
```
andyHa, line 46: coundn' t we use mongo.delete().eq....eq...().executeFor(...) here?

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -0,0 +1,102 @@
...
+            if (owner instanceof Journaled) {
+                JournalData.addJournalEntry(owner,
+                                            "Deleted translated text for " + field.getName() + " (" + lang + ")");
+            }
+        });
+    }
+
+    @Override
+    public void deleteAllTexts(@Nonnull Mapping field) {
+        for (MongoTranslation translation : fetchAllTranslations(field)) {
```
andyHa, line 57: s.a.

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -0,0 +1,223 @@
...
+     * {@link sirius.biz.protocol.Journaled} owners keeping track of the changes made to their translations.
+     *
+     * @param field {@link Mapping} of the translated field
+     * @param lang  language code
+     * @param text  translated text for the given language
+     */
+    public void updateText(@Nonnull Mapping field, String lang, String text) {
+        if (Strings.isEmpty(text)) {
+            deleteText(field, lang);
+            return;
```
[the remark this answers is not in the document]
cschierle: The question is if deleting should only be allowed via updateText() with empty text, or rather via explicit calls to deleteText(). Currently, both options are supported so writing to the journal in deleteText() makes more sense.
It could be moved, if only one of the options should be publicly available

src/main/java/sirius/biz/translations/BasicTranslations.java
cschierle: the @AfterDelete is handled by removeTranslations()

andyHa: nvm - totally didn't see that...

PR #824, opened 2020-07-09
Uses the presence marker fields for checkboxes provided in sirius-web.

andyHa:
If the property is a boolean one, it will most probably handled
 by a checkbox. As an unchecked checkbox will not submit any value
 we therefore add a hidden field which is signals the presence of the
 checkbox. The field is named "property"_marker.

### DIFF
--- a/src/main/java/sirius/biz/web/BizController.java
+++ b/src/main/java/sirius/biz/web/BizController.java
@@ -293,7 +293,7 @@ public class BizController extends BasicController {
 
         // If the property is a boolean one, it will most probably handled
         // by a checkbox. As an unchecked checkbox will not submit any value
-        // we therefore add a hidden field which is signals the presence of the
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -287,9 +293,13 @@ private boolean shouldAutoload(WebContext ctx, Property property) {
 
         // If the property is a boolean one, it will most probably handled
         // by a checkbox. As an unchecked checkbox will not submit any value
-        // we still process this property, which is then considered to be
-        // false (matching the unchecked checkbox).
-        return property instanceof BooleanProperty;
+        // we therefore add a hidden field which is signals the presence of the
```
mkeckmkeck, line 296:
```suggestion
        // we therefore add a hidden field which signals the presence of the
```

PR #828, opened 2020-07-14
Fixed Value class not recognizing String values as LocalDate

marvinwoop:
Would have affected the metric-computers by always returning the current date instead of in console entered date.


src/main/java/sirius/biz/analytics/scheduler/AnalyticsBatchExecutor.java
```diff
@@ -30,7 +31,9 @@
     @Override
     public void executeWork(JSONObject context) throws Exception {
         String schedulerName = context.getString(AnalyticalEngine.CONTEXT_SCHEDULER_NAME);
-        LocalDate date = Value.of(context.get(AnalyticalEngine.CONTEXT_DATE)).asLocalDate(LocalDate.now());
+        LocalDate date =
+                Value.of(NLS.parseMachineString(LocalDate.class, context.getString(AnalyticalEngine.CONTEXT_DATE)))
```
qw3ry, line 35: Wirft der nicht ne IllegalArgumentException wenns schief läuft?

PR #829, opened 2020-07-20
Adds a default set of DateRanges to TraceData for time facets

meg1502:


src/main/java/sirius/biz/protocol/TraceData.java
```diff
@@ -99,6 +100,24 @@
     @Transient
     private boolean silent;
 
+    /**
+     * Returns a default selection of {@link DateRange date ranges} for a time aggregation on {@link #CHANGED_AT}.
+     *
+     * @return an array of date ranges to create a time aggregation on the last changed date
+     * @see sirius.biz.web.MongoPageHelper#addTimeAggregation(Mapping, boolean, DateRange...)
```
andyHa, line 107: +1 for providing a reference - maybe also link to the equivalent ES version in ElasticPageHelper(?)

PR #833, opened 2020-07-27
Provides a baseline for simple processcommands with no arguments

meg1502:
Includes boilerplate to run the process as simulation


src/main/java/sirius/biz/process/SimpleProcessCommand.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    protected abstract String createBaseDescription();
+
+    @Override
+    public String getDescription() {
+        return createBaseDescription() + " Only simulates unless '" + EXECUTE + "' is appended.";
+    }
+
+    protected abstract void executeSimpleProcess(ProcessContext context);
```
sabieber, line 44: wouldnt it be nice to pass a boolean parameter whether the execution is a simulation or not?

PR #834, opened 2020-07-28
Adds a facet for system messages in the process log

meg1502:


meg1502:
Screenshot 2020-07-28 at 15 37 31

PR #836, opened 2020-07-29
BREAKING: Only loads fields from requests when matching param is given

sabieber:
We dont want to reset a field to null when no parameter is given in the request that matches the field.
We also use `get` instead of `getParameters` here as it takes the `WebContext::attributes` into account which allows to semi-inject parameters before loading (e.g. for providing a default value when no parameter is given).

Fixes: OX-6055

### DIFF
--- a/src/main/java/sirius/biz/web/BizController.java
+++ b/src/main/java/sirius/biz/web/BizController.java
@@ -28,6 +28,7 @@ import sirius.kernel.async.Tasks;
 import sirius.kernel.commons.Hasher;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
+import sirius.kernel.commons.Value;
 import sirius.kernel.commons.Values;
 import sirius.kernel.di.std.ConfigValue;
 import sirius.kernel.di.std.Part;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/web/BizController.java
```diff
@@ -256,11 +257,19 @@ protected void load(WebContext ctx, BaseEntity<?> entity, List<Mapping> properti
             Property property = entity.getDescriptor().getProperty(columnProperty);
             String propertyName = property.getName();
 
+            // If the parameter is not present in the request we just skip it to prevent resetting the field to null
```
meg1502, line 258: this comment makes more sense inside right before or inside the if

src/main/java/sirius/biz/web/BizController.java
```diff
@@ -256,11 +257,19 @@ protected void load(WebContext ctx, BaseEntity<?> entity, List<Mapping> properti
...
+            final Value parameterValue = ctx.get(propertyName);
+            if (parameterValue.isNull()) {
+                continue;
+            }
+
             try {
-                property.parseValues(entity, Values.of(ctx.getParameters(propertyName)));
+                property.parseValues(entity,
+                                     Values.of(parameterValue.get(List.class,
+                                                                  Collections.singletonList(parameterValue.get()))));
```
meg1502, line 267: this looks weird, maybe long term we should look at properties only parsing a Value and the properties that actually support multiple values to look for a list inside of the value

sabieber: Yes, although this would probably require quite a lot of changes in sirius-db

src/main/java/sirius/biz/web/BizController.java
```diff
@@ -222,46 +221,54 @@ public static String getBaseUrl() {
...
         boolean hasError = false;
 
         for (Mapping columnProperty : properties) {
             Property property = entity.getDescriptor().getProperty(columnProperty);
             String propertyName = property.getName();
 
+            final Value parameterValue = webContext.get(propertyName);
+            if (parameterValue.isNull()) {
+                // If the parameter is not present in the request we just skip it to prevent resetting the field to null
+                continue;
```
andyHa, line 261: couldn't this be handled via a stream + filter or a sub method + return?

src/main/java/sirius/biz/web/BizController.java
```diff
@@ -272,6 +264,28 @@ protected void load(WebContext ctx, BaseEntity<?> entity, List<Mapping> properti
         }
     }
 
+    private boolean tryLoadProperty(WebContext webContext, BaseEntity<?> entity, Property property) {
```
andyHa, line 267: maybe invert here so that true = success and false = failure?

PR #842, opened 2020-08-17
Normalizes Search Prefixes

fhaScireum:
...as MongoFilterFactory#prefix filters out most special characters from the search query before comparing them to the search prefixes.
Also simplifies null-check.

-Fixes: OX-6081


src/main/java/sirius/biz/mongo/PrefixSearchableEntity.java
```diff
@@ -30,6 +30,14 @@
  */
 public abstract class PrefixSearchableEntity extends MongoEntity {
 
+    /**
+     * Represents a regular expression which detects all character (except whitespace)
```
bwiedmann, line 34:
```suggestion
     * Represents a regular expression which detects all characters (except whitespace)
```

fhaScireum: I fixed it.

PR #843, opened 2020-08-19
Adds Import-/Export-Jobs concerning Archives (Only ZIP for now)

fhaScireum:
- Fixes: OX-5983


idlira (approved):

bwiedmann (approved):

src/main/java/sirius/biz/jobs/batch/file/ArchiveExportJob.java
```diff
@@ -0,0 +1,68 @@
...
+    @Nonnull
+    protected OutputStream createNextArchiveEntry(String filename) throws IOException {
+        if (zipOutputStream == null) {
+            zipOutputStream = new ZipOutputStream(createOutputStream());
+        }
+
+        ZipEntry zipEntry = new ZipEntry(filename);
+        zipOutputStream.putNextEntry(zipEntry);
+
+        return new CloseShieldOutputStream(zipOutputStream);
```
andyHa, line 54: Is it really prohibited to invoke close in that stream?

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
```diff
@@ -0,0 +1,149 @@
...
+     * @throws Exception in case of an exception while importing
+     */
+    protected abstract void importFiles() throws Exception;
+
+    protected List<InputStream> getFiles() {
+        return new ArrayList<>(entries.values());
+    }
+
+    @Nullable
+    protected InputStream getFile(String fileName, boolean isRequired) {
```
andyHa, line 95: Why not search the stream on demand here - should greatly simplify the code and male getFile re usable for the same file

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param fileParameter the parameter which is used to derive the import file from
+     * @param process       the process context in which the job is executed
+     */
+    protected ArchiveImportJob(FileParameter fileParameter, ProcessContext process) {
+        super(fileParameter, process);
+    }
+
+    @Override
+    protected void executeForSingleFile(FileHandle fileHandle) throws Exception {
+        entries.put(fileHandle.getFile().getName(), fileHandle.getInputStream());
```
andyHa, line 52: Shouldn‘t we rather throw/abort here - we acutally expect an archive with multiple files answay

andyHa (changes requested):
...

src/main/java/sirius/biz/jobs/batch/file/ArchiveExportJob.java
fhaScireum: Yes, because the output stream is for the whole archive. #putNextEntry does not create a new output stream but only creates a new entry inside the archive that is being written to when writing to the stream.

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
fhaScireum: I'll look into that.

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
fhaScireum: This might generate quite a few changes in the job structure further down, but will simplify the logic here. I'll change it and see how much refactoring is actually needed.

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    protected final boolean canHandleFileExtension(@Nullable String fileExtension) {
+        return "zip".equalsIgnoreCase(fileExtension);
+    }
+
+    @Override
+    protected void executeForStream(String filename, InputStream in) throws Exception {
+        // never called in overwritten execute method
```
qw3ry, line 122:
```suggestion
        throw new UnsupportedOperationException();
```
? Dann knallts auch falls es doch mal aufgerufen wird

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
```diff
@@ -0,0 +1,130 @@
...
+    protected boolean containsEntries(String... fileNamesToCheck) {
+        ArrayList<? extends ZipEntry> zipEntries = Collections.list(zipFile.entries());
+
+        return Arrays.stream(fileNamesToCheck)
+                     .allMatch(fileName -> zipEntries.stream().anyMatch(entry -> entry.getName().equals(fileName)));
+    }
+
+    @Override
+    protected final boolean canHandleFileExtension(@Nullable String fileExtension) {
+        return "zip".equalsIgnoreCase(fileExtension);
```
qw3ry, line 117:
```suggestion
        return FILE_EXTENSION_ZIP.equalsIgnoreCase(fileExtension);
```

src/main/java/sirius/biz/jobs/batch/file/ArchiveImportJob.java
```diff
@@ -0,0 +1,130 @@
...
+                backupInputFile(file.name(), fileHandle);
+                zipFile = new ZipFile(fileHandle.getFile());
+                importEntries();
+            }
+        } else {
+            throw Exceptions.createHandled().withNLSKey("FileImportJob.fileNotSupported").handle();
+        }
+    }
+
+    protected abstract void importEntries() throws Exception;
```
andyHa, line 61: JavaDoc

qw3ry (approved):

qw3ry (approved):

andyHa (approved):

PR #839, opened 2020-08-10
Makes CodeListEntries translatable

cschierle:
CodeListEntry now implements the Translatable interface.
Class usages in Controllers and other classes have been updated to include the additional required generic.

Translation capabilities are not actively used yet, but will be after upcoming UI changes

Fixes: SIRI-239

PR #844, opened 2020-08-21
WIP: CodeLists can be imported in multiple languages and Translations are updated accordingly

cschierle:
uses changes from https://github.com/scireum/sirius-biz/pull/839

Adds a language parameter to the codelist import. 
- If no language is selected, the standard codelist import is executed. 
- If a language is set, the import will create/update Translation entities for the description of imported entries (codeListEntryData_description) with the given language.

Currently only works for MongoDB.

Fixes: SIRI-243


cschierle:
moved changes to a new PR (#881) with cleaner commit history and without unnecessary changes in develop.conf

PR #850, opened 2020-09-03
Allows job parameters to be logged only as system message or not at all

meg1502:
Altough parameters can be restricted by permission within the collectParameters method, once they are logged they can be seen by everyone who has access to these logs. As this could expose sensitive data, parameters can now restrict their logging behaviour.

Fixes: OX-6144


sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/biz/jobs/batch/BatchProcessJobFactory.java
```diff
@@ -213,16 +216,36 @@ protected void executeTask(ProcessContext process) throws Exception {
...
+
+        if (parameters.stream().anyMatch(Parameter::shouldBeSystemLogged)) {
+            logSystemParameters(process, parameters);
+        }
+    }
 
-        getParameters().forEach(param -> {
+    private void logSystemParameters(ProcessContext process, List<Parameter<?, ?>> parameters) {
+        StringBuilder output = new StringBuilder();
+        output.append("Systemparameter:\n\n");
```
andyHa, line 233: i18n

src/main/java/sirius/biz/jobs/batch/BatchProcessJobFactory.java
```diff
@@ -213,16 +216,36 @@ protected void executeTask(ProcessContext process) throws Exception {
     protected abstract BatchJob createJob(ProcessContext process) throws Exception;
 
     protected void logParameters(ProcessContext process) {
+        List<Parameter<?, ?>> parameters = getParameters();
+
         StringBuilder output = new StringBuilder();
         output.append("Parameter:\n\n");
```
andyHa, line 222: i18n

andyHa: couldn't we re-use the code and call it twice with a different heading and predicate (`Parameter::shouldBeSystemLogged` or `p -> !p.shouldBeSystemLogged()`)

src/main/java/sirius/biz/jobs/params/Parameter.java
```diff
@@ -122,6 +148,25 @@ public boolean isVisible(Map<String, String> context) {
         return get(context).isPresent();
     }
 
+    /**
```
andyHa, line 151: wouldn't a getter be enough here? it is only used in another condition, we could even use this as a parameter (s.a.)

src/main/java/sirius/biz/jobs/batch/BatchProcessJobFactory.java
andyHa: scratch that - add a getVisibility() +- parameter and simply pass in the desired visibility type (other normal or system)...

andyHa (changes requested):
...

idlira (approved):

bwiedmann (approved):

qw3ry (approved):

andyHa (approved):

PR #858, opened 2020-09-15
Improves roles setup for user / tenant management...

andyHa:
* A user of the system tenant, which is permitted to manage users can now also manage and switch tenants.
* Only users of the system tenant can be granted the role "system-administrator" - granting this to non-system-tenants is now rejected on two levels to preven any mis-use / mis-configuration of systems
* A user cannot become/select a fellow user of the system tenant unless the role PERMISSION_MANAGE_SYSTEM_USERS is granted to prevent an unwanted privilege escalation.

### DIFF
--- a/src/main/java/sirius/biz/tenants/UserAccountController.java
+++ b/src/main/java/sirius/biz/tenants/UserAccountController.java
@@ -635,7 +635,7 @@ public abstract class UserAccountController<I, T extends BaseEntity<I> & Tenant<
         }
 
         // If the target user belongs to the system tenant, our current user has to have highest user management
-        // permission as otherwise we would perform an unwanted roles delegation (given the current user higher
+        // permission as otherwise we would perform an unwanted roles delegation (giving the current user higher
         // access rights - right up to the system management level...)
         if (Strings.areEqual(tenants.getTenantUserManager().getSystemTenantId(), user.getTenant().getIdAsString())
             && !getUser().hasPermission(PERMISSION_MANAGE_SYSTEM_USERS)) {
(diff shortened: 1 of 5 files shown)


qw3ry (approved):

src/main/java/sirius/biz/tenants/UserAccountController.java
```diff
@@ -633,6 +634,17 @@ public void selectUserAccount(final WebContext webContext, String accountId) {
             return;
         }
 
+        // If the target user belongs to the system tenant, our current user has to have highest user management
+        // permission as otherwise we would perform an unwanted roles delegation (given the current user higher
```
jmuscireum, line 638:
```suggestion
        // permission as otherwise we would perform an unwanted roles delegation (giving the current user higher
```

jmuscireum (changes requested):

jmuscireum (approved):

src/main/java/sirius/biz/tenants/UserAccountController.java
```diff
@@ -633,6 +633,15 @@ public void selectUserAccount(final WebContext webContext, String accountId) {
...
+        // access rights - right up to the system management level...)
+        if (Strings.areEqual(tenants.getTenantUserManager().getSystemTenantId(), user.getTenant().getIdAsString())
```
sabieber, line 640: Können wir da eig ned mal irgendwo an ner zentralen Stelle ne isSystemTenantMember Methode platzieren, also im usermanager oder usercontext oder so?

sabieber (approved):

src/main/java/sirius/biz/tenants/UserAccountController.java
sabieber: Bzw ne isSystemTenant die nur die Tenant ID bekommt zb

sabieber: letzteres braucht man ab und zu glaube ich schon

idlira (approved):

PR #859, opened 2020-09-15
Only warns about a partial address if more than just the country is f…

andyHa:
…illed.

### DIFF
--- a/src/main/java/sirius/biz/model/InternationalAddressData.java
+++ b/src/main/java/sirius/biz/model/InternationalAddressData.java
@@ -73,9 +73,10 @@ public class InternationalAddressData extends AddressData {
      * Note that this excludes the check if only a country if filled and nothing else (as this is commonly
      * provided by a drop-down selector). Therefore we treat an address with only a country as "empty".
      *
-     * @return <tt>true</tt> if there is at leas one field filled and at least one field left empty.
+     * @return <tt>true</tt> if there is at least one field filled and at least one field left empty.
      * <tt>false</tt> otherwise.
      */
+    @Override
     public boolean isPartiallyFilled() {
         return isAnyFieldEmpty() && !super.areAllFieldsEmpty();
     }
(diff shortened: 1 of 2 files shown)


sabieber (approved):

src/main/java/sirius/biz/model/InternationalAddressData.java
```diff
@@ -67,6 +67,19 @@ public boolean isAnyFieldEmpty() {
...
+    /**
+     * Determines if the given address is partially filled.
+     * <p>
+     * Note that this excludes the check if only a country if filled and nothing else (as this is commonly
+     * provided by a drop-down selector). Therefore we treat an address with only a country as "empty".
+     *
+     * @return <tt>true</tt> if there is at leas one field filled and at least one field left empty.
+     * <tt>false</tt> otherwise.
+     */
+    public boolean isPartiallyFilled() {
```
jmuscireum, line 79:
```suggestion
    @Override
    public boolean isPartiallyFilled() {
```

src/main/java/sirius/biz/model/InternationalAddressData.java
```diff
@@ -67,6 +67,19 @@ public boolean isAnyFieldEmpty() {
         return super.isAnyFieldEmpty() || Strings.isEmpty(country);
     }
 
+    /**
+     * Determines if the given address is partially filled.
+     * <p>
+     * Note that this excludes the check if only a country if filled and nothing else (as this is commonly
+     * provided by a drop-down selector). Therefore we treat an address with only a country as "empty".
+     *
+     * @return <tt>true</tt> if there is at leas one field filled and at least one field left empty.
```
jmuscireum, line 76:
```suggestion
     * @return <tt>true</tt> if there is at least one field filled and at least one field left empty.
```

jmuscireum (changes requested):

idlira (approved):

sabieber (approved):

jmuscireum (approved):

PR #870, opened 2020-10-02
Enables XSD validation for XML Export Jobs

idlira:
Fixes: SIRI-255


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -631,6 +633,24 @@ public OutputStream addFile(String processId, String filename) {
         return process.getFiles().findOrCreateAttachedBlobByName(filename).createOutputStream(filename);
     }
 
+    /**
+     * Returns an input stream to a file stored in the process.
+     * <p>
+     * Note that it the responsibility of the caller to close the stream upon usage.
```
bwiedmann, line 639:
```suggestion
     * Note that it is the responsibility of the caller to close the stream upon usage.
```

bwiedmann (changes requested):
...

bwiedmann (approved):

sabieber (approved):

src/main/java/sirius/biz/jobs/batch/file/XMLExportJob.java
```diff
@@ -0,0 +1,107 @@
...
+    protected void initializeXmlFile(String filename) throws IOException {
+        closeOpenStream();
+        xmlOutputStream = createEntry(filename);
+        xml = new XMLStructuredOutput(xmlOutputStream);
+    }
+
+    @Override
+    public void close() throws IOException {
+        closeOpenStream();
+        super.close();
```
qw3ry, line 62: Careful: This may throw an exception, in which case the digestExportedFile wont get executed

idlira: Yup, but if we can't save the target file then there is not much to validate anyway. Besides, the job will also fail.

qw3ry (approved):

src/main/java/sirius/biz/jobs/batch/file/ArchiveExportJob.java
```diff
@@ -79,5 +83,28 @@ protected void collectParameters(Consumer<Parameter<?, ?>> parameterCollector) {
...
+
+    /**
+     * Digests every entry of a fresh created export archive.
+     * <p>
+     * Override this method in order to perform validations on contents of the final archive.
+     *
+     * @param digester a consumer receiving the name and the {@link InputStream} of each entry
+     */
+    protected void digestExportedFile(BiConsumer<String, InputStream> digester) {
+        digestExportedFile(inputStream -> processExportedArchive(inputStream, digester));
```
andyHa, line 95: isn't

src/main/java/sirius/biz/jobs/batch/file/ArchiveExportJob.java
```diff
@@ -79,5 +83,28 @@ protected void collectParameters(Consumer<Parameter<?, ?>> parameterCollector) {
             super.collectParameters(parameterCollector);
         }
     }
+
+    /**
+     * Digests every entry of a fresh created export archive.
+     * <p>
+     * Override this method in order to perform validations on contents of the final archive.
```
andyHa, line 90: should this be overwritten? this would remove the logic here?

andyHa: shouldn't we separate validaten logic from store logic? it its no clear when what has to be invoked..

src/main/java/sirius/biz/jobs/batch/file/ArchiveExportJob.java
```diff
@@ -79,5 +83,28 @@ protected void collectParameters(Consumer<Parameter<?, ?>> parameterCollector) {
...
+     *
+     * @param digester a consumer receiving the name and the {@link InputStream} of each entry
+     */
+    protected void digestExportedFile(BiConsumer<String, InputStream> digester) {
```
andyHa, line 94: why should this be overwritten?
when is this invoked where?

idlira: As the method name says, this method provides access to the InputStream of the fresh exported file (does not matter which type: could be a CSV, or a ZIP containing several files). Since getting access to this stream is not always needed, I find it more elegant that the caller explicitly does it if needed. For this there are different methods depending on the parent class (FileExportJob or ArchiveExportJob). The wording validation could be somewhat misleading I agree... is just the typical use case why one needs this InputStream.

XMLExportJob is one place that makes usage of this digester, in order to run the XSD validation on the export file.

andyHa (approved):

PR #879, opened 2020-10-14
Allows to provide multiple XSDs for validation in XML import jobs

sabieber:
Also skips validation entirely when a specific parameter value is selected.
We need to allow the user to switch between multiple schemas for formats that have variants.
Example: ARGE XML also has a DGH variant with some of the required fields being missing or optional.

Fixes: OX-6257


src/main/java/sirius/biz/jobs/batch/file/XMLImportJobFactory.java
```diff
@@ -42,12 +49,14 @@ public String getIcon() {
...
+     * This allows to provide more than one XSD for validation when multiple variants of a data format are supported.
      *
-     * @return the path to the XSD file if the XML file should be validated, null otherwise
+     * @return a map of paths to XSD files if the XML file should be validated, an empty map otherwise
      */
-    @Nullable
-    protected String getXsdResourcePath() {
-        return null;
+    @Nonnull
+    protected Map<String, String> getXsdResourcePaths() {
```
andyHa, line 59: create a map externally and offer it here as BiConsumer<String,String> here? this also permits to enforce a LinkedHashMap which guarantees order

andyHa (commented):
...

idlira (approved):

idlira (approved):

marvinwoop (approved):

src/main/java/sirius/biz/jobs/batch/file/XMLImportJobFactory.java
```diff
@@ -42,12 +46,11 @@ public String getIcon() {
...
-     *
-     * @return the path to the XSD file if the XML file should be validated, null otherwise
+     * Adds xsd schema paths to the entryConsumer if the XML file should and can be validated.
+     * <p>
+     * This allows to provide more than one XSD for validation when multiple variants of a data format are supported.
      */
-    @Nullable
-    protected String getXsdResourcePath() {
-        return null;
+    protected void getXsdResourcePaths(BiConsumer<String, String> entryConsumer) {
```
mkeckmkeck, line 53: Rename to fillXsdResourcePaths. Get* returning void smells a little.

mkeckmkeck (approved):

bwiedmann (approved):

PR #881, opened 2020-10-15
CodeLists can be im-/exported in multiple languages

cschierle:
Fixes: SIRI-243

CodeLists can now be imported with translations:
- files with additional columns holding translations (e.g. "de_codeListEntryData_value", "fr_codeListEntryData_value",...) are supported (see example in [TestDE_EN.txt](https://github.com/scireum/sirius-biz/files/5428518/TestDE_EN.txt))
- if the language parameter is not set, the translation columns are ignored and the basic import is executed
- basic import for files without translation columns is still supported
- an error message is emitted, if the imported file has no columns matching the selected language

CodeLists can now be exported with translations:
- selecting a language for the export job emits translations for CodeListEntries in additional columns next to the standard output
- a standard export without a selected language produces the standard output


src/main/java/sirius/biz/codelists/CodeListExportJob.java
```diff
@@ -19,14 +20,19 @@
...
 import sirius.kernel.commons.Strings;
+import sirius.kernel.nls.NLS;
 
 import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
+import static sirius.biz.codelists.CodeListImportJob.*;
```
sabieber, line 34: Static imports find ich nicht so schön

sabieber (changes requested):

src/main/java/sirius/biz/codelists/CodeListExportJob.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Nullable
+    @Override
+    protected Function<E, Object> customFieldExtractor(String field) {
+        if (!field.startsWith(languagePrefix) || !languageParameter.isPresent()) {
+            return null;
+        }
+
+        String realField = field.replaceFirst(languagePrefix, "");
```
andyHa, line 124: doesn't this belong in the import handler?

src/main/java/sirius/biz/codelists/CodeListImportJob.java
```diff
@@ -0,0 +1,158 @@
...
+                             Class<E> type,
+                             ImportDictionary dictionary,
+                             ProcessContext process,
+                             String factoryName) {
+        super(fileParameter, ignoreEmptyParameter, importModeParameter, type, dictionary, process, factoryName);
+        this.codeList = process.require(codeListParameter);
+        this.languageParameter = languageParameter;
+        languageParameter.ifPresent(lang -> {
+            this.languagePrefix = lang + "_";
+            this.dictionary.addField(FieldDefinition.stringField(languagePrefix + CLE_VALUE)
```
andyHa, line 86: this feels like it belongs in the import handler

src/main/java/sirius/biz/codelists/CodeListImportJob.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @param entity  the entity to persist
+     * @param context the row represented as context
+     */
+    @Override
+    protected void createOrUpdate(E entity, Context context) {
+        super.createOrUpdate(entity, context);
+
+        // update the translation texts for the selected language
+        languageParameter.ifPresent(lang -> {
```
andyHa, line 139: importhandler?

src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
```diff
@@ -32,7 +36,10 @@
 public class SQLCodeListExportJobFactory
         extends EntityExportJobFactory<SQLCodeListEntry, SmartQuery<SQLCodeListEntry>> {
 
-    private CodeListParameter codeListParameter = new CodeListParameter("codeList", "$CodeList").markRequired();
+    private static final CodeListParameter CODE_LIST_PARAMETER =
```
andyHa, line 39: wasn't there a reason to make parameters non static? internal translations or the like?

src/main/java/sirius/biz/model/Constants.java
```diff
@@ -0,0 +1,42 @@
...
+ * Provides helpful constants and methods regarding language configuration data
+ * (and known currencies, countries, ... coming soon™)
+ */
+public class Constants {
+
+    private Constants() {
+    }
+
+    public static List<String> getLanguages() {
+        return Sirius.getSettings().getStringList("nls.languages");
```
andyHa, line 27: doesn't NLS provides this already? if not, why not add the method there?

andyHa (changes requested):
...

src/main/java/sirius/biz/model/Constants.java
cschierle: I looked at how language parameters are handled in Oxomi and that's how it's done there 🤷‍♂ The Constants Class holds a lot more constants and functions there though...

src/main/java/sirius/biz/codelists/CodeListImportJob.java
cschierle: If we move it to the handler, we would have to duplicate the code for SQL and Mongo. That would be unnecessary, since there are no database dependent instructions needed to update translation texts. The Translation entity handles that by itself

src/main/java/sirius/biz/codelists/CodeListImportJob.java
cschierle: This also prevents unnecessary code duplication (see below)

src/main/java/sirius/biz/codelists/CodeListExportJob.java
cschierle: This also prevents unnecessary code duplication (see below)

src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
cschierle: there are also a lot of static parameters in other projects' import/export jobs, so I don't think that is an issue?

src/main/resources/biz_en.properties
```diff
@@ -703,6 +704,9 @@ TraceData.created = Created
 TraceData.createdAt = Created
 TraceData.createdBy = Created by
 TraceData.infos = Information
+Translations.export.lang.help = Leave this field empty to only export the standard values of the code list. Select a specific language to export the corresponding translations for the code list entries as well.
+Translations.import.lang.help = Leave this field empty to only import the standard values of the code list. Select a specific language to import the corresponding translations for the code list entries as well.
+Translations.import.languageNotFound = Cannot find selected language in the imported file.
```
idlira, line 709:
```suggestion
Translations.import.languageNotFound = Cannot find the selected language in the imported file.
```

idlira (approved):

idlira (approved):

idlira (approved):

src/main/java/sirius/biz/codelists/CodeListExportJob.java
idlira: To prevent code duplication, it has been agreed to keep the code in this class.

src/main/java/sirius/biz/codelists/CodeListImportJob.java
idlira: To prevent code duplication, it has been agreed to keep the code in this class.

src/main/java/sirius/biz/codelists/CodeListImportJob.java
idlira: To prevent code duplication, it has been agreed to keep the code in this class.

src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
idlira: Parameters on all jobs are now constants.

src/main/java/sirius/biz/codelists/CodeListExportJob.java
andyHa: EntityImportHandlerExtender could hande both - see PerformanceDataXXX or PermissionDataXXX

cschierle: That is a valid suggestion, but after looking into it not feasible at the moment. We need access to parameters from the jobs to build things like exportable mappings dynamically, which is currently not possible with HandlerExtenders. 
I suggest we do a larger rework similar to the universal import as a follow-up

idlira:
Closing since the framework was deprecated.

PR #898, opened 2020-11-05
Adds errorneous facet to processes page

fhaScireum:
- Fixes: SIRI-274


sabieber (approved):

andyHa:
Why do we add an errorneos facet? :-P

andyHa (approved):

mkeckmkeck (approved):

src/main/resources/biz_en.properties
```diff
@@ -508,6 +508,7 @@ PersonData.title = Title
 Process = Process
 Process.completed = Completed
 Process.counters = Performance
+Process.errorneous = Errorneous
```
idlira, line 511:
```suggestion
Process.errorneous = Erroneous
```

src/main/java/sirius/biz/process/ProcessController.java
```diff
@@ -96,6 +96,7 @@ public void processes(WebContext ctx) {
         ElasticPageHelper<Process> pageHelper = ElasticPageHelper.withQuery(query);
         pageHelper.withContext(ctx);
         pageHelper.addTermAggregation(Process.STATE, ProcessState.class);
+        pageHelper.addBooleanAggregation(Process.ERRORNEOUS);
```
idlira, line 99: Actually ERRONEOUS ist the right word.

src/main/resources/biz_en.properties
fhaScireum: Huh, seems like the field is named wrongly

src/main/java/sirius/biz/process/ProcessController.java
fhaScireum: :-)

fhaScireum:
> Why do we add an errorneos facet? :-P

So it is easier to find erroneous processes :-P (requested by OXOMI support)

PR #900, opened 2020-11-06
BREAKING: Valid languages for Translations are now customizable

cschierle:
Translations used the same property value for supportedLanguages as MLS did. This value is removed in https://github.com/scireum/sirius-db/pull/377.

Similar to the changes in that PR, Translations now also provide a constructor with a custom set of valid language codes. If that set is empty, all language codes are accepted.

NOTE: marked as breaking due to the changed behavior with an empty set. Translations are not actively used yet, so no other areas are currently affected.

Fixes: SIRI-240


idlira (approved):

idlira (approved):

sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -142,14 +178,8 @@ public void updateText(@Nonnull Mapping field, String lang, String text) {
      * @param field {@link Mapping} of the translated field
      * @param lang  language code
      * @return {@link Optional} with translated text for the given language, or empty if none is found
-     * @throws IllegalArgumentException if lang is not supported by the system
      */
     public Optional<String> getText(@Nonnull Mapping field, String lang) {
-        if (!isSupportedLanguage(lang)) {
-            throw new IllegalArgumentException(
-                    "lang must be a language code supported by the system (supportedLanguages)!");
-        }
```
cschierle, line 151: Validity check for languages is already performed during the creation of a Translation entity.

Double-checking when getting texts currently leads to exceptions making affected pages unusable. Removed the exception, since returning an empty Optional is sufficient here and keeps pages functional.

idlira (approved):

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -111,15 +112,23 @@ public void updateText(@Nonnull Mapping field, String lang, String text) {
...
+            updateTranslation(translation);
+            if (owner instanceof Journaled) {
+                JournalData.addJournalEntry(owner,
+                                            String.format("Updated translated text for %s (%s): '%s'",
+                                                          field.getName(),
+                                                          lang,
+                                                          text));
+            }
+        } else if (owner instanceof Journaled) {
             JournalData.addJournalEntry(owner,
```
idlira, line 126: Humm, JournalData is not really the place to log errors.

idlira (changes requested):
JournalData vs Logger

fhaScireum (approved):

src/main/java/sirius/biz/translations/BasicTranslations.java
cschierle: the isValidLanguage() check in findOrCreateTranslation already produces a Handled exception (and returns null) if the language is invalid. So yeah, maybe the journal entry isn't needed then?

idlira (approved):

src/main/java/sirius/biz/translations/BasicTranslations.java
```diff
@@ -58,11 +74,23 @@ protected BasicTranslations(BaseEntity<?> owner) {
...
+     * @param field the name of the field to be translated
      * @return true if language is supported by the system, false otherwise
+     * @throws sirius.kernel.health.HandledException if lang is not supported by the system
      */
-    protected boolean isSupportedLanguage(String lang) {
-        return supportedLanguages.contains(lang);
+    protected boolean isValidLanguage(String lang, String field) {
+        boolean isSupported = validLanguages.isEmpty() || validLanguages.contains(lang);
+
+        if (!isSupported) {
```
andyHa, line 85: this is very unexpected behavior - isXXX returning true/false should never throw instead of returning false.

Rename to "assert / ensureValidLanguage" and remove return type?

src/main/java/sirius/biz/translations/jdbc/SQLTranslations.java
```diff
@@ -102,9 +114,8 @@ public void deleteAllTexts(@Nonnull Mapping field) {
 
     @Override
     protected SQLTranslation findOrCreateTranslation(@Nonnull Mapping field, String lang, String text) {
-        if (!isSupportedLanguage(lang)) {
-            throw new IllegalArgumentException(
-                    "lang must be a language code supported by the system (supportedLanguages)!");
+        if (!isValidLanguage(lang, field.getName())) {
+            return null;
```
andyHa, line 118: unreachable code

src/main/java/sirius/biz/translations/mongo/MongoTranslations.java
```diff
@@ -71,9 +83,8 @@ public void deleteAllTexts(@Nonnull Mapping field) {
 
     @Override
     protected MongoTranslation findOrCreateTranslation(@Nonnull Mapping field, String lang, String text) {
-        if (!isSupportedLanguage(lang)) {
-            throw new IllegalArgumentException(
-                    "lang must be a language code supported by the system (supportedLanguages)!");
+        if (!isValidLanguage(lang, field.getName())) {
+            return null;
```
andyHa, line 87: s.a.

andyHa (changes requested):

idlira (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #902, opened 2020-11-10
SE-10138: Cross-lingual handling of salutations

marvinwoop:
Since it is possible in S2 to send e-mails to several addressees in different languages, they must be addressable in their respective language.

### DIFF
--- a/src/main/java/sirius/biz/codelists/CodeLists.java
+++ b/src/main/java/sirius/biz/codelists/CodeLists.java
@@ -428,7 +428,7 @@ public abstract class CodeLists<I, L extends BaseEntity<I> & CodeList, E extends
     }
 
     /**
-     * Returns the value translated from the given code list associated with the given code.
+     * Returns the value translated from the given code list associated with the given code for the current language.
      * <p>
      * If no matching entry exists, the code itself will be returned.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/model/PersonData.java
```diff
@@ -212,6 +225,16 @@ public String getTranslatedSalutation() {
         return codeLists.getTranslatedValue(CODE_LIST_SALUTATIONS, salutation);
     }
 
+    /**
+     * Returns the value (by langCode specified translated name) of the salutation.
+     *
+     * @param langCode the language code to translate to
+     * @return the value for <tt>salutation</tt> from the <tt>salutations</tt> code list
+     */
+    public String getUserSpecificTranslatedSalutation(String langCode) {
```
mkeckmkeck, line 234:
```suggestion
    public String getTranslatedSalutation(String langCode) {
```

src/main/java/sirius/biz/codelists/CodeLists.java
```diff
@@ -438,7 +438,21 @@ public String getValue(@Nonnull String codeListName, @Nullable String code, @Nul
...
+     *
+     * @param codeListName the code list to search in
+     * @param code         the code to lookup
+     * @param langCode     the language code to translate into
+     * @return the translated value associated with the code or the code itself if no value exists
+     */
+    public String getTranslatedValue(@Nonnull String codeListName, @Nullable String code, String langCode) {
+        String value = getValue(codeListName, code, langCode);
         if (value != null && value.startsWith("$")) {
             return Value.of(value).translate().getString();
```
mkeckmkeck, line 457: Don't we need the langcode for Value.of(value).translate() too?

mkeckmkeck (changes requested):
Make naming more consistent

sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/model/PersonData.java
marvinwoop: Renamed method.

src/main/java/sirius/biz/codelists/CodeLists.java
marvinwoop: Yes we do actually need that code in that method. Added it.

idlira (approved):

mkeckmkeck (approved):

qw3ry (approved):

PR #905, opened 2020-11-11
Provides a streamlined API to access 7-ZIP

andyHa:
Using native code brings quite some risks with it as it might crash the JVM. Therefore
the bindings have to be enabled manually. Also, we always use the Java API to handle
ZIP files as this is faster (no disk buffer required) and safer - 7-ZIP has quite some
bugs when decoding paths in ZIP files created by Microsoft Windows


src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+import java.util.zip.ZipFile;
+
+/**
+ * Utility to handle and extract archive files.
+ * <p>
+ * If the framework <tt>biz.seven-zip</tt> is enabled, this will load the native 7-ZIP bindings and support a
+ * multitude of file formats like <tt>7z, tar, rar etc.</tt> otherwise "just" ZIP files are supported.
+ * <p>
+ * Note that using native code always bears a certain risk as an error there crashed the whole JVM. Thereore enabling
+ * this is only advised on clusters which have dedicated worker servers.
```
jakobvogel, line 50:
```suggestion
 * Note that using native code always bears a certain risk as an error there crashed the whole JVM. Therefore enabling
 * this is only advised on clusters which have dedicated worker servers.
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+        if (Sirius.isFrameworkEnabled(FRAMEWORK_SEVEN_ZIP)) {
+            try {
+                Log.SYSTEM.INFO("Loading 7-ZIP native libraries to support all breeds of archive formats...");
+                SevenZip.initSevenZipFromPlatformJAR();
+                sevenZipEnabled = true;
+            } catch (Throwable e) {
+                Exceptions.handle()
+                          .to(Log.SYSTEM)
+                          .error(e)
+                          .withSystemErrorMessage("Failed to inizialize 7-Zip: %s (%s)")
```
jakobvogel, line 132:
```suggestion
                          .withSystemErrorMessage("Failed to initialize 7-Zip: %s (%s)")
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+                throw Exceptions.createHandled()
+                                .error(sevenZipException)
+                                .withSystemErrorMessage("7-ZIP failed (the archive is probably corrupted): %s (%s)")
+                                .handle();
+            }
+        }
+    }
+
+    /**
+     * XIterates over all items of an archive file.
```
jakobvogel, line 257: Is `XIterates` deliberately chosen to mark extraction?

src/main/java/sirius/biz/util/ExtractedFile.java
```diff
@@ -0,0 +1,57 @@
...
+     * Returns the path of the file.
+     *
+     * @return the path as indicated by the archive
+     */
+    String getFilePath();
+
+    /**
+     * Returns the progress in percent based on the number of files processed.
+     *
+     * @return the ration between the number of processed files and the total number of files in the archive
```
jakobvogel, line 54:
```suggestion
     * @return the ratio between the number of processed files and the total number of files in the archive
```

src/main/java/sirius/biz/util/ExtractedFile.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Returns the path of the file.
+     *
+     * @return the path as indicated by the archive
+     */
+    String getFilePath();
+
+    /**
+     * Returns the progress in percent based on the number of files processed.
```
jakobvogel, line 52: The name implies a factor of 100 (percent), the description of the returned value a simple ratio, i.e. between `0` and `1`. Clarify?

src/main/java/sirius/biz/util/SevenZipAdapter.java
```diff
@@ -0,0 +1,192 @@
...
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.function.Predicate;
+
+/**
+ * Acts as a callback between the 7-ZIP binding and our API as defined by {@link ExtractedFile}.
+ * <p>
+ * Note that due to the <i>interesting</i> nature of the 7-ZIP API, this contains a whole of stateful variables which
+ * are setup in each of the callbacks and processed later (mainly in {@link #setOperationResult(ExtractOperationResult)}.
```
jakobvogel, line 35:
```suggestion
 * are set up in each of the callbacks and processed later (mainly in {@link #setOperationResult(ExtractOperationResult)}.
```

src/main/resources/component-biz.conf
```diff
@@ -64,6 +64,9 @@ sirius.frameworks {
     # metadata repository.
     biz.storage-replication-mongo = false
 
+    # Determines if the 7-ZIP native bindings will be enabled or not.
+    biz.seve-zip = false
```
jakobvogel, line 68: Probably this?
```suggestion
    biz.seven-zip = false
```

jakobvogel (changes requested):

jakobvogel (approved):

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+    private Set<String> supportedExtensions;
+    private Boolean sevenZipEnabled = null;
+
+    /**
+     * Defines if and how existing files are overwritten.
+     *
+     * @see #updateFile(ExtractedFile, VirtualFile, OverrideMode)
+     */
+    public enum OverrideMode {
+        ON_CHANGE, ALWAYS, NEVER;
```
sabieber, line 75: Would also love some JavaDoc on public enum values

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+        return fileExtension != null && getSupportedFileExtensions().contains(fileExtension.toLowerCase());
+    }
+
+    /**
+     * Determines if the given archive is a ZIP file.
+     *
+     * @param fileExtension the file extension of the file to check
+     * @return <tt>true</tt> if the given file is a ZIP file, <tt>false</tt> otherwise
+     */
+    public boolean isZIPFile(@Nullable String fileExtension) {
```
sabieber, line 157:
```suggestion
    public boolean isZipFile(@Nullable String fileExtension) {
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+     * @param filter                determines which files will be processed
+     * @param extractedFileConsumer invoked for each extracted file. May return <tt>false</tt> to abort processing or
+     *                              <tt>true</tt> to continue
+     */
+    public void extract(String filename,
+                        File archiveFile,
+                        @Nullable Predicate<String> filter,
+                        Processor<ExtractedFile, Boolean> extractedFileConsumer) {
+        try {
+            if (isZIPFile(Files.getFileExtension(filename))) {
```
sabieber, line 175:
```suggestion
            if (isZipFile(Files.getFileExtension(filename))) {
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+                extractZIP(archiveFile, enhanceFileFilter(filter), extractedFileConsumer);
```
sabieber, line 176:
```suggestion
                extractZip(archiveFile, enhanceFileFilter(filter), extractedFileConsumer);
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+            return this::ignoreHiddenFiles;
+        }
+    }
+
+    private boolean ignoreHiddenFiles(String path) {
+        String filename = Files.getFilenameAndExtension(path);
+        return !filename.startsWith(".") && !filename.startsWith("__MACOSX");
+    }
+
+    private void extractZIP(File archiveFile,
```
sabieber, line 202:
```suggestion
    private void extractZip(File archiveFile,
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+                            Predicate<String> filter,
+                            Processor<ExtractedFile, Boolean> extractedFileConsumer) throws Exception {
+        try (ZipFile zipFile = new ZipFile(archiveFile)) {
+            extractZIPEntriesFromZipFile(filter, extractedFileConsumer, zipFile);
```
sabieber, line 206:
```suggestion
            extractZipEntriesFromZipFile(filter, extractedFileConsumer, zipFile);
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+            // This is most probably an error indicating an inconsistent ZIP archive. We therefore directly throw
+            // a handled exception to avoid jamming the syslog...
+            throw Exceptions.createHandled()
+                            .error(zipException)
+                            .withSystemErrorMessage("Failed to unzip the given archive: %s (%s)")
+                            .handle();
+        }
+    }
+
+    private void extractZIPEntriesFromZipFile(Predicate<String> filter,
```
sabieber, line 217:
```suggestion
    private void extractZipEntriesFromZipFile(Predicate<String> filter,
```

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+    /**
+     * Iterates over the items of an archive file.
+     *
+     * @param filename              the filename of the archive to extract
+     * @param archiveFile           the archive file to extract
+     * @param filter                determines which files will be processed
+     * @param extractedFileConsumer invoked for each extracted file. May return <tt>false</tt> to abort processing or
+     *                              <tt>true</tt> to continue
+     */
+    public void extract(String filename,
```
mkeckmkeck, line 170: Wouldn't it be better to pass the file extension as parameter and get the proper filename from the java.io File when needed?

src/main/java/sirius/biz/util/ArchiveExtractor.java
```diff
@@ -0,0 +1,313 @@
...
+ * Also note that ZIP files are always processed using the Java APIs as this is way simpler to process (can be
+ * done in memory without allocating large buffers or resorting to disk - therefore this is also faster (as disk
+ * access is always slower than executing Java-Code). Also 7-ZIP has some severe bugs when handling paths in ZIP
+ * files which were produced by Microsoft Windows.
+ * <p>
+ * Note that {@link #getSupportedFileExtensions()} honors the framework selection and always returns an appropriate
+ * list.
+ */
+@Register(classes = ArchiveExtractor.class)
+public class ArchiveExtractor {
```
sabieber, line 61: Is this a replacement for the ArchiveHelper class?

sabieber (changes requested):

src/main/java/sirius/biz/util/ArchiveExtractor.java
mkeckmkeck, line 175: ist't isZipFile our naming convention?

mkeckmkeck (changes requested):
Should we maybe double check for sevenZipEnabled in actual extraction method?

src/main/java/sirius/biz/util/ArchiveExtractor.java
andyHa: yes this will be deleted in a subsequent PR

src/main/java/sirius/biz/util/ArchiveExtractor.java
andyHa: this is a temporary file and will most probably has a random filename

idlira (approved):

src/main/java/sirius/biz/util/ArchiveExtractor.java
andyHa: I added an appropriate comment in the JavaDoc. (One could argue that most probably the file extension would match, but this approach seems a bit cleaner.

sabieber (approved):

PR #910, opened 2020-11-16
Lazily initializes a parameter.

andyHa:
As we need to access an injected field to properly create the parameter
we cannot perform this in the constructor. We therefore resort to a lazy
initialization pattern.

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/ExtractArchiveJob.java
+++ b/src/main/java/sirius/biz/storage/layer3/ExtractArchiveJob.java
@@ -48,7 +48,7 @@ public class ExtractArchiveJob extends SimpleBatchProcessJobFactory {
     @Part
     private ArchiveExtractor extractor;
 
-    private final Parameter<VirtualFile> sourceParameter;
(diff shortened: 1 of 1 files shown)


meg1502 (approved):

sabieber (approved):

idlira (changes requested):
...

src/main/java/sirius/biz/storage/layer3/ExtractArchiveJob.java
```diff
@@ -84,9 +76,22 @@ public ExtractArchiveJob() {
                         "$ExtractArchiveJob.deleteArchiveParameter.help").withDefaultTrue().build();
     }
 
+    private Parameter<VirtualFile> getSourceParameter() {
```
idlira, line 79: Shall we rename this method? It does more than a simple getter.

idlira (approved):

sabieber (approved):

jakobvogel (approved):

mkeckmkeck (approved):

PR #904, opened 2020-11-11
BREAKING CHANGE: Makes Job Parameters immutable.

andyHa:
As we more and more tend to use parameters as globally shared constants, it is neccesarry to make them immutable once shared.

Therefore we differentiate between "ParameterBuilder", which is the mutable way of setting up a parameter as wanted and "Parameter" itself, which is then an immutable object which can safely be shared.

This leads to some refactorings and simplifications, hence the amount of changed files. The actual changes are quite small.

PR #913, opened 2020-11-18
Fixes for Jobs and Parameters

idlira:
Fixes some bugs found with the breaking changes introduced with https://github.com/scireum/sirius-biz/pull/904

- FileImportJob considers archives
- XSD validation re-enabled in XML exports+imports
- Enhancement: ability to run more than one pass over an XML file with a different set of handlers

Fixes: OX-6341

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
@@ -20,16 +20,16 @@ import java.util.function.Consumer;
  */
 public abstract class XMLExportJobFactory extends FileExportJobFactory {
 
-    protected final Parameter<Boolean> requireValidFile =
+    protected final BooleanParameter requireValidFileBuilder =
             new BooleanParameter("requireValidFile", "$XMLExportJobFactory.requireValidFile").hidden()
                                                                                              .withDescription(
-                                                                                                     "$XMLExportJobFactory.requireValidFile.help")
-                                                                                             .build();
+                                                                                                     "$XMLExportJobFactory.requireValidFile.help");
 
     @Override
     protected void collectParameters(Consumer<Parameter<?>> parameterCollector) {
         if (getXsdResourcePath() != null) {
-            parameterCollector.accept(requireValidFile);
+            requireValidFileBuilder.visible();
+            parameterCollector.accept(requireValidFileBuilder.build());
         }
         super.collectParameters(parameterCollector);
     }
--- a/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
@@ -20,16 +20,15 @@ import java.util.function.Consumer;
  */
 public abstract class XMLExportJobFactory extends FileExportJobFactory {
 
-    protected final BooleanParameter requireValidFileBuilder =
-            new BooleanParameter("requireValidFile", "$XMLExportJobFactory.requireValidFile").hidden()
-                                                                                             .withDescription(
-                                                                                                     "$XMLExportJobFactory.requireValidFile.help");
+    public static final String VALID_FILE_PARAMETER_NAME = "requireValidFile";
 
     @Override
     protected void collectParameters(Consumer<Parameter<?>> parameterCollector) {
         if (getXsdResourcePath() != null) {
-            requireValidFileBuilder.visible();
-            parameterCollector.accept(requireValidFileBuilder.build());
+            Parameter<Boolean> validFileParameter = new BooleanParameter(VALID_FILE_PARAMETER_NAME,
+                                                                         "$XMLExportJobFactory.requireValidFile").withDescription(
+                    "$XMLExportJobFactory.requireValidFile.help").build();
+            parameterCollector.accept(validFileParameter);
         }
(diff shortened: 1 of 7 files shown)


src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
```diff
@@ -20,16 +20,16 @@
  */
 public abstract class XMLExportJobFactory extends FileExportJobFactory {
 
-    protected final Parameter<Boolean> requireValidFile =
+    protected final BooleanParameter requireValidFileBuilder =
```
andyHa, line 23: make a method and return a fresh instance each time? this avoids accidentally modifying a shared object

andyHa: maybe also define the name as reusable constant or a placeholder parameter as was there before.

src/main/java/sirius/biz/jobs/batch/file/XMLImportJob.java
```diff
@@ -31,9 +30,9 @@
  */
 public abstract class XMLImportJob extends FileImportJob {
 
-    public static final Parameter<String> XSD_SCHEMA_PARAMETER =
+    protected static final SelectStringParameter XSD_SCHEMA_PARAMETER_BUILDER =
```
andyHa, line 33: provide a factory method and or the name as shared constant.

maybe also prevent a lookup in the string list for when reading the parameter as the verification is done before and the value can be assumed to be correct. in other words a SelectStringParameter should behave like a StringParameter when it comes to reading the value from the job...

src/main/java/sirius/biz/jobs/params/ParameterBuilder.java
```diff
@@ -261,7 +261,7 @@ protected V require(Map<String, String> context) {
      *
      * @return the name of the parameter
      */
-    protected String getName() {
```
andyHa, line 264: i think its better to share the name as a constant than sharing a mutable builder - as this is very easy to get wrong...(modify global state unintentionally)

src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
idlira: Added a shared name. Builder removed completely as this is handled internally by this class.

src/main/java/sirius/biz/jobs/batch/file/XMLImportJob.java
idlira: Builder delivered by static method.

sabieber (approved):

src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
```diff
@@ -20,16 +20,15 @@
  */
 public abstract class XMLExportJobFactory extends FileExportJobFactory {
 
-    protected final Parameter<Boolean> requireValidFile =
```
andyHa, line 23: why not keep this parameter this is a simple boolean and commonly needs no customization..

src/main/java/sirius/biz/jobs/batch/file/XMLImportJob.java
```diff
@@ -47,14 +49,21 @@
...
+        return new SelectStringParameter(XSD_SCHEMA_PARAMETER_NAME, "$XMLImportJobFactory.xsdSchema").withDescription(
+                "$XMLImportJobFactory.xsdSchema.help");
     }
 
     @Override
     protected void executeForStream(String filename, Producer<InputStream> inputSupplier) throws Exception {
         if (isValid(inputSupplier)) {
-            try (InputStream inputStream = inputSupplier.create()) {
-                executeForValidStream(inputStream);
+            for (Consumer<BiConsumer<String, NodeHandler>> handlerConsumer : fetchHandlersList()) {
```
andyHa, line 63:
```suggestion
            for (Consumer<BiConsumer<String, NodeHandler>> processingStage : fetchStages()) {
```

src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
idlira: parameter still exist, just the scope changes. It's created in L28 if xsd paths were provided.
https://github.com/scireum/sirius-biz/blob/d66daa6a906c18b1ce3ef23c0d65e178233ebe6b/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java#L28

src/main/java/sirius/biz/jobs/batch/file/XMLImportJob.java
```diff
@@ -47,14 +49,21 @@
...
+                try (InputStream inputStream = inputSupplier.create()) {
+                    executeForValidStream(inputStream, handlerConsumer);
```
andyHa, line 65:
```suggestion
                    executeProcessingStage(inputStream, stage);
```

andyHa (commented):
Some minor renamings and one minor remark.

qw3ry (approved):

meg1502 (approved):

jakobvogel (approved):

sabieber (approved):

PR #927, opened 2020-11-26
SE-10166: Add SystemHelper to access environment variables

marvinwoop:
- Fixes SE-10166

### DIFF
--- a/src/main/java/sirius/biz/util/SystemHelper.java
+++ b/src/main/java/sirius/biz/util/SystemHelper.java
@@ -25,7 +25,7 @@ public class SystemHelper {
 
     /**
      * This environment variable is defined in the docker-compose.yaml.erb template file and will alway be "prod",
-     * "staging" or "test", as our node namings are fixes to that environment definitions.
+     * "staging" or "test", as our node namings are fixed to that environment definitions.
      */
     private static final String ENVIRONMENT = "ENVIRONMENT";
 
--- a/src/main/java/sirius/biz/util/SystemHelper.java
+++ b/src/main/java/sirius/biz/util/SystemHelper.java
@@ -1,0 +1,118 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.util;
+
+import sirius.kernel.di.std.Register;
+import sirius.kernel.health.Exceptions;
+import sirius.kernel.health.Log;
+import sirius.web.security.HelperFactory;
+import sirius.web.security.ScopeInfo;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides static System functionality and system environment detection as helper for in-template usage.
+ */
+public class SystemHelper {
+
+    private static final String SYSTEM_HELPER = "SystemHelper";
+
+    /**
+     * This environment variable is defined in the docker-compose.yaml.erb template file and will alway be "prod",
+     * "staging" or "test", as our node namings are fixes to that environment definitions.
+     */
+    private static final String ENVIRONMENT = "ENVIRONMENT";
+
+    @Register
+    public static class SystemHelperFactory implements HelperFactory<SystemHelper> {
+
+        @Override
+        public Class<SystemHelper> getHelperType() {
+            return SystemHelper.class;
+        }
+
+        @Nonnull
+        @Override
+        public String getName() {
+            return SYSTEM_HELPER;
+        }
+
+        @Override
+        public SystemHelper make(ScopeInfo scopeInfo) {
+            return new SystemHelper();
+        }
+    }
+
+    /**
+     * Returns the value of the given environment variable.
+     *
+     * @param varName the environment variable to look up
+     * @return the value of the environment variable
+     */
+    public String getEnvironmentVariable(String varName) {
+        return System.getenv(varName);
+    }
+
+    /**
+     * Returns the current {@link SystemEnvironment} of the system.
+     *
+     * @return the current {@link SystemEnvironment} of the system.
+     */
+    public SystemEnvironment getCurrentSystemEnvironment() {
+        try {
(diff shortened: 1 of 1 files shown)


andyHa (approved):

src/main/java/sirius/biz/util/SystemHelper.java
```diff
@@ -0,0 +1,118 @@
...
+         * Defines the testing system environment
+         */
+        TEST,
+
+        /**
+         * Defines a not defined system environment. This can be used to indicate that a faulty environment variable
+         * has been set. If a system environment is undefined, it is strongly recommended to check the environment
+         * variables.
+         */
+        UNDEFINED
```
idlira, line 116: Shoudn't we make this a breaking change and force a variable to be set?

sabieber (approved):

src/main/java/sirius/biz/util/SystemHelper.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return <tt>true</tt> if this is a productive system
+     */
+    public boolean isProductiveSystem() {
+        return getCurrentSystemEnvironment() == SystemEnvironment.PROD;
+    }
+
+    /**
+     * Defines the systems current environment
+     */
+    public enum SystemEnvironment {
```
idlira, line 94: Is an enum a good idea? PROD, STAGING and TEST are strongly our terms. If people elsewhere is supposed to use this framework, I would rather move this into config settings. Something like:

```
system.environment {
   name = "STAGING"
   isProductive = false
}
```

src/main/java/sirius/biz/util/SystemHelper.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Returns the current {@link SystemEnvironment} of the system.
+     *
+     * @return the current {@link SystemEnvironment} of the system.
+     */
+    public SystemEnvironment getCurrentSystemEnvironment() {
+        try {
+            return SystemEnvironment.valueOf(System.getenv(ENVIRONMENT).toUpperCase());
```
idlira, line 69: I would rather choose a not so generic name. SIRIUS_ENVIRONMENT perhaps?

idlira (changes requested):
Some conceptual suggestions. But let's see what the others have to say!

src/main/java/sirius/biz/util/SystemHelper.java
marvinwoop: That's a good point. If using the config settings, shouldn't we rather name it sirius.environment as mentioned below?

src/main/java/sirius/biz/util/SystemHelper.java
marvinwoop: Yes, i agree on that.

src/main/java/sirius/biz/util/SystemHelper.java
idlira: SiriusEnvironment is good (Class and config name)... Actually, using config settings would kill the need for an environment variable.

qw3ry: prod/staging/test are very generic IMO. For most use cases this will suffice. OTOH configuration via config file sounds reasonable; after all, we have all our configuration in the config files

src/main/java/sirius/biz/util/SystemHelper.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class SystemHelper {
+
+    private static final String SYSTEM_HELPER = "SystemHelper";
+
+    /**
+     * This environment variable is defined in the docker-compose.yaml.erb template file and will alway be "prod",
+     * "staging" or "test", as our node namings are fixed to that environment definitions.
+     */
+    private static final String ENVIRONMENT = "ENVIRONMENT";
```
mkeckmkeck, line 30: Change key + value here

src/main/java/sirius/biz/util/SystemHelper.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Determines if the current System is a productive system.
+     *
+     * @return <tt>true</tt> if this is a productive system
+     */
+    public boolean isProductiveSystem() {
+        return getCurrentSystemEnvironment() == SystemEnvironment.PROD;
+    }
```
mkeckmkeck, line 89: Is this required? I would drop it.

src/main/java/sirius/biz/util/SystemHelper.java
mkeckmkeck: I would keep it simple for the beginning and so stick to a few defined enums. Add DEV and it should be fine for the beginning. We can then still make it more sophisticated (aka configurable) later. Also this could probably lead to environment specific config files. But as there is no need for this now, we should keep it simple

src/main/java/sirius/biz/util/SystemHelper.java
mkeckmkeck: I would keep the usage of the "environment" optional and so use an undefined.

mkeckmkeck (changes requested):

src/main/java/sirius/biz/util/SystemHelper.java
idlira: There is already a Sirius.isDev() - which tells you are under the IDE...
AFAIK, we already have and maintain environment-specific configs anyway (your local instance.conf and the ones rolled out via puppet).
Besides, reading a config entry is a one liner...

sabieber: In my opinion the config suggestion is actually quite good as we already have system dependent configs in most cases as we want to use different storage configurations, endpoints, etc. Moving this setting to the config would mean all system dependent config can be done via config files and will not be split up between config and environment.

marvinwoop:
According to internal consultation, this will be split into a sirius-kernel part and a sirius-web part. Therefore, this PR will be closed.

PR #934, opened 2020-12-01
Provides a documentation for the replicationSpace parameter.

andyHa:


idlira (approved):

src/main/resources/component-biz.conf
```diff
@@ -699,6 +699,11 @@ storage {
                 # Specifies the passphrase used to initialize the cipher provider
                 # (e.g. used by the AES256CipherFactory).
                 passphrase = ""
+
+                # Can be used to configure the backup storage space for this space. This has to be the
+                # name of another configured layer1 storage space. The replication manager will then
+                # pick this up and transfer all objects to the backup.
```
sabieber, line 705: Could you please include that the name of the used target space will also be used as bucket name in the replication storage - so most probably different spaces should have different replicationSpaces configured?

mkeckmkeck (approved):

src/main/resources/component-biz.conf
andyHa: this has nothing to do with the replication but is inherent to the way layer1 spaces are configured when using s3 as storage system. We always use the space name as bucket name unless a "bucketName" is given.

PR #935, opened 2020-12-01
Permits to specify an explicit bucket name to use.

andyHa:


sabieber (approved):

src/main/java/sirius/biz/storage/layer1/S3ObjectStorageSpace.java
```diff
@@ -73,7 +81,7 @@ private ObjectStore resolveObjectStore(Extension extension) {
     }
 
     private BucketName bucketName() {
-        return store.getBucketName(name);
+        return store.getBucketName(bucketName);
```
meg1502, line 84: This still adds the suffix to the bucket name, wasn't the intention of this change to also supress the suffix?

meg1502 (changes requested):

meg1502 (approved):

idlira (approved):

fhaScireum (approved):

PR #939, opened 2020-12-07
SIRI-286 Sanitize paths in the storage engine

jmuscireum:


src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -173,6 +176,39 @@ public static String normalizePath(@Nullable String path) {
...
+    @Nonnull
+    public String sanitizePath(@Nullable String path) {
+        path = Strings.trim(path);
+
+        if (Strings.isEmpty(path)) {
+            return "";
+        }
+
+        if (path.contains("\\")) {
+            path = path.replace("\\", "/");
```
qw3ry, line 194: ReplaceAll?

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -373,36 +375,30 @@ protected Blob fetchByPath(String tenantId, @Nonnull String path) {
...
-
     @Override
     public Optional<? extends Blob> findByPath(String path) {
         return findByPath(UserContext.getCurrentUser().getTenantId(), path);
     }
 
     @Override
     public Blob findOrCreateByPath(String tenantId, String path) {
-        if (Strings.isEmpty(path)) {
+        String sanitizedPath = utils.sanitizePath(path);
```
sabieber, line 385: I think this call has to be added to Blob::updateContent as well as we could otherwise overwrite a sanitized path with an unsanitized one by accident when updating the content of a blob and passing a file name

sabieber (changes requested):

src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -173,6 +176,39 @@ public static String normalizePath(@Nullable String path) {
...
+        }
+
+        if (path.contains("//")) {
+            path = path.replaceAll("/+", "/");
+        }
+
+        if (path.startsWith("/")) {
+            path = path.substring(1);
```
qw3ry, line 202: Ist absolut = relativ? War vorher ja das gleiche Verhalten, aber das verdient IMO nen erklärenden Kommentar

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
jmuscireum: updateContent only updates the file name. The only way to "unsanitize" the path is to pass an invalid file name, which is a separate problem?

src/main/java/sirius/biz/storage/util/StorageUtils.java
jmuscireum: replace already replaces each occurrence. The only difference is that replaceAll expects a regex pattern.

src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -173,6 +176,42 @@ public static String normalizePath(@Nullable String path) {
...
+     */
+    @Nonnull
+    public String sanitizePath(@Nullable String path) {
+        path = Strings.trim(path);
+
+        if (Strings.isEmpty(path)) {
+            return "";
+        }
+
+        if (path.contains("\\")) {
```
andyHa, line 196: why check for a contains?

src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -173,6 +176,42 @@ public static String normalizePath(@Nullable String path) {
...
+            path = path.replace("\\", "/");
+        }
+
+        if (path.contains("//")) {
```
andyHa, line 200: s.a.

src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -173,6 +176,42 @@ public static String normalizePath(@Nullable String path) {
...
+     *
+     * @param path the path to cleanup
+     * @return the sanitized path without backslashes, successive slashes, and without leading and trailing slashes
+     */
+    @Nonnull
+    public String sanitizePath(@Nullable String path) {
+        path = Strings.trim(path);
+
+        if (Strings.isEmpty(path)) {
+            return "";
```
andyHa, line 193: return "/" as all other cases return a leading slash

andyHa (commented):
...

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
sabieber: Yes thats exactly the problem, which when done will kill most parts of the storage system (vfs roots cant list any files of a space containing a single file with broken filename anymore)

We at least should create a ticket, my thought was that its at least related to the path sanitization (as the name seems to be created from the sanitized path in most cases, only the updateContent method somehow bypasses this logic.

jmuscireum: The ticket SIRI-251 should cover this

sabieber: Ah thx 👍

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -373,36 +375,30 @@ protected Blob fetchByPath(String tenantId, @Nonnull String path) {
...
+
+        if (Strings.isEmpty(sanitizedPath)) {
             throw new IllegalArgumentException("An empty path was provided!");
         }
 
-        String key = determinePathCacheKey(tenantId, path);
+        String key = determinePathCacheKey(tenantId, sanitizedPath);
         Blob blob = blobByPathCache.get(key);
         if (blob == null) {
-            blobByPathCache.remove(key);
```
andyHa, line 398: wasn't this put here intentionally as removed is distributed over the intercom and put isn't? @idlira

idlira: Yup, put is not distributed

andyHa (approved):

idlira (approved):

PR #942, opened 2020-12-10
Dependency Upgrade

jakobvogel:
> SIRI-288

- Updates pom.xml


jakobvogel:
Merged myself, dependency upgrade only.

PR #943, opened 2020-12-17
Misc Fixes: Replication, Conversion Engine, Processes

andyHa:
* The replication task storages now properly compute the executable tasks.
* The conversion engine runs more tasks in parallel and also measures the actual conversion performance properly.
* Some cleanup in the default configs.
* If a Process get terminated due to a system restart, we now mark it as erroneous.

Fixes: SIRI-294

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -28,7 +28,6 @@ import sirius.kernel.commons.Producer;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Wait;
-import sirius.kernel.commons.Watch;
 import sirius.kernel.di.std.ConfigValue;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.health.Exceptions;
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1408,12 +1407,14 @@ private V awaitConversionResultAndRetryToFindVariant(B blob, String variantName,
...
-            try (FileHandle automaticHandle = handle) {
+        conversionEngine.performConversion(blob, variant.getVariantName()).onSuccess(handleAndWatch -> {
+            try (FileHandle automaticHandle = handleAndWatch.getFirst()) {
                 String physicalKey = keyGenerator.generateId();
                 getPhysicalSpace().upload(physicalKey, automaticHandle.getFile());
-                markConversionSuccess(variant, physicalKey, automaticHandle.getFile().length(), watch.elapsedMillis());
+                markConversionSuccess(variant,
+                                      physicalKey,
+                                      automaticHandle.getFile().length(),
+                                      handleAndWatch.getSecond().elapsedMillis());
```
sabieber, line 1417: Is this enough added precision for now or do we want to track download, conversion and upload duration separately right away?

sabieber (approved):
Thanks for the fixes

idlira (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
andyHa: Well we could add a 2nd measurement for the "overall time" - this would als include the queuing time...

qw3ry (approved):

mkeckmkeck (approved):

PR #959, opened 2021-01-11
Fixes the output of a not required SelectStringParameter

idlira:
Selecting no value at all was actually delivering an Optional of a blank string and so isPresent() would always return true.

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/SelectStringParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/SelectStringParameter.java
@@ -64,9 +64,6 @@ public class SelectStringParameter extends SelectParameter<String, SelectStringP
 
     @Override
     protected Optional<String> resolveFromString(@Nonnull Value input) {
-        if (input.isEmptyString()) {
-            return Optional.empty();
-        }
-        return Optional.of(input.asString());
+        return input.asOptionalString();
     }
 }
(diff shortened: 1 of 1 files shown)


bwiedmann (approved):

sabieber (approved):

andyHa (approved):

src/main/java/sirius/biz/jobs/params/SelectStringParameter.java
```diff
@@ -64,6 +64,9 @@ protected String checkAndTransformValue(Value input) {
 
     @Override
     protected Optional<String> resolveFromString(@Nonnull Value input) {
+        if (input.isEmptyString()) {
+            return Optional.empty();
+        }
         return Optional.of(input.asString());
```
qw3ry, line 70:
```suggestion
        return Optional.of(input.asString()).filter(Strings::isFilled);
```
Ist ein bisschen ideomatischer (evtl auch `ofNullable`)

idlira: Making a pre-check would avoid running an unnecessary `Value.asString`, which validates the underlying object against several conditions defined by NLS.toMachineString.

qw3ry:
```suggestion
        return input.asOptionalString();
```
Checked again: What you want is this. It does exactly what you did yourself, as far as i can tell

idlira: It does! Thx for checking.

PR #970, opened 2021-01-28
Reports unused NLS keys and minor misc fixes...

andyHa:
* Using a new job, we can determine unused NLS keys within the application
* Provides some helpers to run code as system tenant
* Prevents an NPE when querying the BlobStorageSpaces without file types
* Skips adding the default port to S3-URLs

### DIFF
--- a/src/main/java/sirius/biz/cluster/NLSController.java
+++ b/src/main/java/sirius/biz/cluster/NLSController.java
@@ -36,21 +36,21 @@ public class NLSController extends BasicController {
     private InterconnectClusterManager clusterManager;
 
     @Override
-    public void onError(WebContext ctx, HandledException error) {
-        ctx.respondWith().error(HttpResponseStatus.INTERNAL_SERVER_ERROR, error);
+    public void onError(WebContext webContext, HandledException error) {
+        webContext.respondWith().error(HttpResponseStatus.INTERNAL_SERVER_ERROR, error);
     }
 
     /**
      * Reports all unused NLS keys.
      *
-     * @param ctx the request to handle
+     * @param webContext the request to handle
      * @param out the output to write the JSON to
      * @param token the cluster authentication token
      */
     @Routed(value = "/system/nls/unused/:1", jsonCall = true)
-    public void unused(WebContext ctx, JSONStructuredOutput out, String token) {
+    public void unused(WebContext webContext, JSONStructuredOutput out, String token) {
         if (!clusterManager.isClusterAPIToken(token)) {
-            ctx.respondWith().error(HttpResponseStatus.UNAUTHORIZED);
+            webContext.respondWith().error(HttpResponseStatus.UNAUTHORIZED);
             return;
         }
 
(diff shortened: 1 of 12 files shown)


src/main/java/sirius/biz/cluster/NLSController.java
```diff
@@ -0,0 +1,86 @@
...
+import sirius.kernel.nls.Translation;
+import sirius.web.controller.BasicController;
+import sirius.web.controller.Routed;
+import sirius.web.http.WebContext;
+import sirius.web.services.JSONStructuredOutput;
+
+/**
+ * Provides services to fetch unused or autocreated NLS keys.
+ * <p>
+ * This is mainly used by {@link sirius.biz.util.ReportUnusedNLSKeysJob} to identify NLS keys which are unused accross
```
idlira, line 25:
```suggestion
 * This is mainly used by {@link sirius.biz.util.ReportUnusedNLSKeysJob} to identify NLS keys which are unused across
```

src/main/java/sirius/biz/util/ReportUnusedNLSKeysJob.java
```diff
@@ -0,0 +1,105 @@
...
+                          } else {
+                              missingKeys.retainAll(keysOfNode);
+                          }
+                      });
+
+        if (missingKeys.isEmpty()) {
+            process.log("No unused keys found...");
+        } else {
+            process.log(ProcessLog.info().withFormattedMessage("Exporting %s unused keys...", missingKeys.size()));
+            try (PrintStream out = new PrintStream(process.addFile("unused_keys.txt"))) {
```
idlira, line 94: Why not use process.addTable? One can just visualize or export the table to a local file...

idlira (changes requested):
...

src/main/java/sirius/biz/util/ReportUnusedNLSKeysJob.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public String getDescription() {
+        return "Queries all unused NLS keys on each node in the cluster, computes the intersection and reports this as textfile within the created process.";
+    }
+
+    @Override
+    protected void execute(ProcessContext process) throws Exception {
+        process.log("Fetching unused keys from all nodes...");
+        Set<String> missingKeys = new HashSet<>();
+        Monoflop mf = Monoflop.create();
```
sabieber, line 69: more descriptive name what the monoflip does pls

sabieber (changes requested):

qw3ry (approved):

mkeckmkeck (approved):

src/main/java/sirius/biz/util/ReportUnusedNLSKeysJob.java
andyHa: This data is processed most probably any way (it is just a single key per line nothing more). Also, running this right after a patch might generate large datasets which will further fille the processlogs table with many more entries.

idlira (approved):

sabieber (approved):

src/main/java/sirius/biz/util/ReportUnusedNLSKeysJob.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.annotation.Nullable;
+import java.io.PrintStream;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+/**
+ * Collects all unused NLS keys accross all cluster nodes and reports the intersection as text file.
```
bwiedmann, line 33:
```suggestion
 * Collects all unused NLS keys across all cluster nodes and reports the intersection as text file.
```

bwiedmann (changes requested):

PR #974, opened 2021-02-01
do not index the login page for search engines

qw3ry:
... except it is the root page


qw3ry:
Depends on https://github.com/scireum/sirius-web/pull/800

sabieber (approved):

src/main/resources/default/templates/biz/login.html.pasta
```diff
@@ -1,6 +1,8 @@
 <i:arg type="String" name="originalUrl"/>
+<i:arg type="boolean" name="noindex"
+       default="@('/' != CallContext.getCurrent().get(WebContext.class).getRawRequestedURI())"/>
```
andyHa, line 3: is this also true for "http://host.tld"

andyHa (commented):
See question inline.

meg1502 (approved):

src/main/resources/default/templates/biz/login.html.pasta
qw3ry: At least in S2 the URI is `http://host.tld`, while the `WebContext` gives me `/`

mkeckmkeck (approved):

src/main/resources/default/templates/biz/login.html.pasta
andyHa: +1

andyHa (approved):

PR #993, opened 2021-02-15
Implement a handler to replace blob:// URIs in PDFs

jmuscireum:
This can be used instead of referencing the blobs via http in case the server is having problems resolving its own domain.

### DIFF
--- a/src/main/java/sirius/biz/web/BlobPdfReplaceHandler.java
+++ b/src/main/java/sirius/biz/web/BlobPdfReplaceHandler.java
@@ -12,6 +12,7 @@ import org.xhtmlrenderer.extend.FSImage;
 import org.xhtmlrenderer.extend.UserAgentCallback;
 import sirius.biz.storage.layer1.FileHandle;
 import sirius.biz.storage.layer2.BlobStorage;
+import sirius.biz.storage.layer2.URLBuilder;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Register;
@@ -23,7 +24,7 @@ import java.util.Optional;
 /**
  * Resolves blob:// URIs to resized images while maintaining the image ratios.
  * <p>
- * The format of the URI needs to match blob://space/variant/blobKey.
+ * The format of the URI needs to match blob://space/blobKey or blob://space/blobKey/variant.
  */
 @Register
 public class BlobPdfReplaceHandler extends PdfReplaceHandler {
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/biz/web/BlobPdfReplaceHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import sirius.kernel.commons.Strings;
+import sirius.kernel.di.std.Part;
+import sirius.kernel.di.std.Register;
+import sirius.web.templates.pdf.handlers.PdfReplaceHandler;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
+ * Responsible for resolving blob:// URIs and resizing the image while maintaining the image ratio.
```
sabieber, line 24: Resolves ...

sabieber (changes requested):

mkeckmkeck (approved):

bwiedmann (approved):

qw3ry (approved):

src/main/java/sirius/biz/web/BlobPdfReplaceHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import sirius.kernel.di.std.Register;
+import sirius.web.templates.pdf.handlers.PdfReplaceHandler;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
+ * Resolves blob:// URIs to resized images while maintaining the image ratios.
+ * <p>
+ * The format of the URI needs to match blob://space/variant/blobKey.
```
andyHa, line 26: shouldn't this be blob://space/blobKey/variant?

Also do we plan to support readable URLs like /pub/my_magic_image.jpg?

jmuscireum: This is how line 49 implements it. I can switch it around if you'd like.

Readable URLs would currently only make sense for the public space, which is not part of biz.

src/main/java/sirius/biz/web/BlobPdfReplaceHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public FSImage resolveUri(String uri, UserAgentCallback userAgentCallback, int cssWidth, int cssHeight)
+            throws Exception {
+        String[] blobInfo = Strings.split(uri, "://").getSecond().split("/");
+
+        if (blobInfo.length != 3) {
+            throw new IllegalArgumentException("The URI is required to match the format 'blob://space/variant/blobKey'");
+        }
+
+        Optional<FileHandle> fileHandle = storage.getSpace(blobInfo[0]).download(blobInfo[2], blobInfo[1]);
```
andyHa, line 49: with `blob://space/blobKey/variant` this would become:
`storage.getSpace(blobInfo[0]).download(blobInfo[1], blobInfo[2]);` right?

Thats what I meant...

bwiedmann (approved):

andyHa (approved):

sabieber (approved):

mkeckmkeck (approved):

cschierle (approved):

PR #1001, opened 2021-02-19
MLS: Right aligns dropdown items so they won't be cut on smaller screens.

bwiedmann:
Bildschirmfoto 2021-02-19 um 16 49 31
Bildschirmfoto 2021-02-19 um 16 49 37

Fixes: SIRI-319

### DIFF
--- a/src/main/resources/default/assets/javascript/multiLanguageField.js
+++ b/src/main/resources/default/assets/javascript/multiLanguageField.js
@@ -19,6 +19,7 @@ function MultiLanguageField(options) {
     this.languageManagementEnabled = options.languageManagementEnabled;
     this.rows = options.rows;
     this.multiline = options.multiline || false;
+    this.mobileOrSmallScreen = window.matchMedia('(max-width: 600px)').matches;
 
     this.validLanguages = {};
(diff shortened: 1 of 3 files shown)


sabieber (approved):

src/main/resources/default/assets/javascript/multiLanguageField.js
```diff
@@ -19,6 +19,7 @@ function MultiLanguageField(options) {
     this.languageManagementEnabled = options.languageManagementEnabled;
     this.rows = options.rows;
     this.multiline = options.multiline || false;
+    this.mobileOrSmallScreen = window.matchMedia('(max-width: 600px)').matches;
```
andyHa, line 22: shouldnt we put such methods into a central place like the core.js in sirius-web - if we ever change page breaks or the like no one will ever find this...

bwiedmann: A better approach instead of only checking the viewport width would be to check for the label length and the tab width of all visible tabs in combination. Therefore more than 3 tabs could be displayed depending on the screen size. (thx @mbrsci for the idea)

PR #1022, opened 2021-03-11
VFS Modal Bugfixing

qw3ry:
ymo found some bugs:
- pagination was only 5 items
- search does work with enter key, but not with the button
- when clicking on breadcrumbs, a js error occurred

Furthermore, the search field was not reset when changing directory, which doesn't make much sense

### DIFF
--- a/src/main/resources/default/extensions/wondergem-page/storage-vfs-modals.html.pasta
+++ b/src/main/resources/default/extensions/wondergem-page/storage-vfs-modals.html.pasta
@@ -53,15 +53,14 @@
                     filter: function (child) {
                         return true;
                     },
-                    createRow: function (child, pagination, _modal, resolve) {
+                    createRow: function (child, _modal, resolve, cd) {
                         const self = this;
 
                         const _tr = document.createElement("tr");
                         _tr.innerHTML = Mustache.render("<td><a class='file-link' href='#' data-path='{{path}}'>" +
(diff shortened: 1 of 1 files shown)


src/main/resources/default/extensions/wondergem-page/storage-vfs-modals.html.pasta
```diff
@@ -53,15 +53,14 @@
                     filter: function (child) {
                         return true;
                     },
-                    createRow: function (child, pagination, _modal, resolve) {
+                    createRow: function (child, _modal, resolve, cd) {
```
sabieber, line 56: could we choose a more descriptive name?

sabieber (changes requested):

src/main/resources/default/extensions/wondergem-page/storage-vfs-modals.html.pasta
sabieber: roughly 5 mins after merging this we have forgot what cd stands for ;)

idlira (approved):

ymo-sci (approved):

mkeckmkeck (approved):

jmuscireum (approved):

sabieber (approved):

andyHa (approved):

PR #1025, opened 2021-03-16
Outputs the translated value for CodeListEntries again

cschierle:
to prevent list entries like `$Country.de`

Fixes: OX-6731

### DIFF
--- a/src/main/java/sirius/biz/translations/MultiLanguageStringProperty.java
+++ b/src/main/java/sirius/biz/translations/MultiLanguageStringProperty.java
@@ -205,20 +205,15 @@ public class MultiLanguageStringProperty extends BaseMapProperty
     @SuppressWarnings("unchecked")
     protected Object transformFromMongo(Value object) {
         Map<String, String> texts = new LinkedHashMap<>();
-
         Object valueObject = object.get();
-        if (valueObject == null) {
-            return texts;
-        }
         if (valueObject instanceof String) {
             texts.put(MultiLanguageString.FALLBACK_KEY, valueObject.toString());
-            return texts;
-        }
-
-        for (Document document : (List<Document>) valueObject) {
-            Object textValue = document.get(TEXT_PROPERTY);
-            if (textValue != null) {
-                texts.put(document.get(LANGUAGE_PROPERTY).toString(), textValue.toString());
+        } else if (valueObject instanceof List) {
+            for (Document document : (List<Document>) valueObject) {
+                Object textValue = document.get(TEXT_PROPERTY);
+                if (textValue != null) {
+                    texts.put(document.get(LANGUAGE_PROPERTY).toString(), textValue.toString());
+                }
             }
         }
         return texts;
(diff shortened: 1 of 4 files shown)


src/main/resources/default/taglib/w/codelistMultiSelect.html.pasta
```diff
@@ -21,7 +21,7 @@
            items="@Injector.context().getPart(CodeLists.class).getEntries(list)">
         <option value="@entry.getCodeListEntryData().getCode()"
                 @selected="value.contains(entry.getCodeListEntryData().getCode())">
-            @entry.getCodeListEntryData().getValue()
+            @entry.getCodeListEntryData().getTranslatedValue(NLS.getCurrentLang())
```
andyHa, line 24: the same probably also applis to code-list-entries.html.pasta (?)

andyHa (commented):
...

src/main/resources/default/taglib/w/codelistMultiSelect.html.pasta
cschierle: if so, it would rather be in multiLanguageTextField.html.pasta. Will check, if we need it there

idlira (approved):

sabieber (approved):

src/main/resources/default/taglib/w/codelistMultiSelect.html.pasta
cschierle: My bad, was thinking about the code-list-entry.html.pasta. But I think in both cases (when viewing the list of entries for a codelist, or editing an individual entry) it makes sense to display the clear text.

We still needed it for the job param template though

mkeckmkeck (approved):

src/main/java/sirius/biz/translations/MultiLanguageStringProperty.java
```diff
@@ -205,7 +205,17 @@ public MultiLanguageString getMultiLanguageString(Object target) {
     @SuppressWarnings("unchecked")
     protected Object transformFromMongo(Value object) {
         Map<String, String> texts = new LinkedHashMap<>();
-        for (Document document : (List<Document>) object.get()) {
+
+        Object valueObject = object.get();
+        if (valueObject == null) {
```
andyHa, line 210: remove

src/main/java/sirius/biz/translations/MultiLanguageStringProperty.java
```diff
@@ -205,7 +205,17 @@ public MultiLanguageString getMultiLanguageString(Object target) {
...
+            return texts;
+        }
+        if (valueObject instanceof String) {
+            texts.put(MultiLanguageString.FALLBACK_KEY, valueObject.toString());
+            return texts;
+        }
+
```
andyHa, line 217:
```suggestion
        } else if (valueObject instanceof List) {

```

andyHa (commented):
could be simplified a bit

sabieber (approved):

idlira (approved):

PR #1027, opened 2021-03-17
Fixes OOM (Java Heap Space) when Unzipping by Writing to File if necessary + Refactoring

fhaScireum:
(non breaking)

- Fixes: OX-6679

### DIFF
--- a/src/main/java/sirius/biz/util/ExtractedFileBuffer.java
+++ b/src/main/java/sirius/biz/util/ExtractedFileBuffer.java
@@ -45,9 +45,9 @@ public class ExtractedFileBuffer {
      * @param tempFilePrefix    the prefix of the temporary file which will be created if the memory threshold is reached
      */
     public ExtractedFileBuffer(int memoryThreshold, int initialBufferSize, String tempFilePrefix) {
-        buffer = new DeferredFileOutputStream(DEFAULT_IN_MEMORY_THRESHOLD,
-                                              DEFAULT_INITIAL_BUFFER_SIZE,
-                                              DEFAULT_TMP_FILE_PREFIX,
+        buffer = new DeferredFileOutputStream(memoryThreshold,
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/biz/util/ExtractedFileBuffer.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Wraps a {@link DeferredFileOutputStream stream} that contains an extracted file either retained in memory or in
+ * a temporary file.
+ */
+public class ExtractedFileBuffer {
```
andyHa, line 25: implement Closable?

idlira (approved):

src/main/java/sirius/biz/util/ExtractedFileBuffer.java
fhaScireum: I thought about that, but we only close the output stream but no input streams. I'm not sure if that would confuse people, but i can change that if you'd like.

andyHa (approved):
Looks generally good, maybe implement Closable to earn extra checks by Java/inspections. Or make this class package private as it shouldn't be used outside....

src/main/java/sirius/biz/util/ExtractedFileBuffer.java
andyHa: make this class package private then (?)

fhaScireum: Done

sabieber (approved):

mkeckmkeck (approved):

cschierle (approved):

PR #1028, opened 2021-03-18
Fixes Im- and Export of CodeLists

cschierle:
- Fixes some errors induced by existing MongoCodeLists
- Exports only the fallback of value and additionalValue instead of dumping the entire MultiLanguageString, so we output `value` in the export instead of `{"fallback" : "value"}`.
-> we should probably revisit supporting multiple languages soon (see #881)

Fixes: OX-6731

### DIFF
--- a/src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
+++ b/src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
@@ -18,7 +18,6 @@ import sirius.biz.jobs.params.Parameter;
 import sirius.biz.process.ProcessContext;
 import sirius.biz.process.ProcessLink;
 import sirius.db.jdbc.SmartQuery;
-import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Register;
 import sirius.web.http.QueryString;
 import sirius.web.security.Permission;
@@ -50,12 +49,17 @@ public class SQLCodeListExportJobFactory
     }
 
     @Override
+    protected String createProcessTitle(Map<String, String> context) {
+        return super.createProcessTitle(context) + codeListParameter.get(context)
+                                                                    .map(codeList -> " - " + codeList.getCodeListData()
+                                                                                                     .getName())
+                                                                    .orElse("");
+    }
+
+    @Override
     protected void executeTask(ProcessContext process) throws Exception {
         process.addLink(new ProcessLink().withLabel("$CodeList")
                                          .withUri("/code-list/" + process.require(codeListParameter).getIdAsString()));
-        process.updateTitle(Strings.apply("%s - %s",
-                                          process.getTitle(),
-                                          process.require(codeListParameter).getCodeListData().getName()));
         super.executeTask(process);
     }
 
--- a/src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
+++ b/src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
@@ -16,7 +16,9 @@ import sirius.biz.jobs.batch.file.EntityExportJobFactory;
 import sirius.biz.jobs.params.CodeListParameter;
 import sirius.biz.jobs.params.Parameter;
 import sirius.biz.process.ProcessContext;
+import sirius.biz.process.ProcessLink;
 import sirius.db.jdbc.SmartQuery;
+import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Register;
 import sirius.web.http.QueryString;
 import sirius.web.security.Permission;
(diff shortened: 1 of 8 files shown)


src/main/java/sirius/biz/codelists/mongo/MongoCodeListEntryImportHandler.java
```diff
@@ -53,6 +55,18 @@ protected MongoCodeListEntryImportHandler(Class<?> clazz, ImporterContext contex
...
+    public Function<? super MongoCodeListEntry, ?> createExtractor(String fieldToExport) {
+        if (fieldToExport.equals(CodeListEntry.CODE_LIST_ENTRY_DATA.inner(CodeListEntryData.VALUE).getName())) {
+            return codeListEntry -> codeListEntry.getCodeListEntryData().getValue().getFallback();
+        }
+        if (fieldToExport.equals(CodeListEntry.CODE_LIST_ENTRY_DATA.inner(CodeListEntryData.ADDITIONAL_VALUE)
+                                                                   .getName())) {
+            return codeListEntry -> codeListEntry.getCodeListEntryData().getAdditionalValue().getFallback();
+        }
+        return super.createExtractor(fieldToExport);
+    }
```
cschierle, line 68: do we need this for the SQL side as well? If so, I can add it there too

meg1502 (approved):

mkeckmkeck (approved):

src/main/java/sirius/biz/translations/MultiLanguageString.java
```diff
@@ -324,7 +324,9 @@ public boolean isWithFallback() {
     @Override
     public void setData(Map<String, String> newData) {
         // remove keys with null values first
-        super.setData(newData.entrySet()
+        super.setData(newData == null ?
+                      Collections.emptyMap() :
+                      newData.entrySet()
```
sabieber, line 329:
```suggestion
        if (newData == null) {
            super.setData(Collections.emptyMap());
            return:
        }
        
        super.setData(newData.entrySet()
```
I really dont like a ternary inside of a method call :)

sabieber (changes requested):

src/main/java/sirius/biz/translations/MultiLanguageString.java
cschierle: Actually had it like this at first and then "optimized" it :D

But a separate if is actually cleaner I guess

idlira (approved):

src/main/java/sirius/biz/translations/MultiLanguageString.java
```diff
@@ -323,6 +323,11 @@ public boolean isWithFallback() {
 
     @Override
     public void setData(Map<String, String> newData) {
+        if (newData == null) {
+            super.setData(Collections.emptyMap());
```
andyHa, line 327: shouldn't we simply clear our map here - Collections.emptyMap() returns an unmodifyable map(?) which would lead to successive failures...

cschierle: Good point 👍  Didn't consider that the emptyMap is unmodifiable

andyHa (changes requested):
...

fhaScireum (approved):

src/main/java/sirius/biz/codelists/jdbc/SQLCodeListExportJobFactory.java
```diff
@@ -47,6 +49,16 @@ public String getName() {
         return SQLCodeListEntry.class;
     }
 
+    @Override
+    protected void executeTask(ProcessContext process) throws Exception {
+        process.addLink(new ProcessLink().withLabel("$CodeList")
+                                         .withUri("/code-list/" + process.require(codeListParameter).getIdAsString()));
+        process.updateTitle(Strings.apply("%s - %s",
```
andyHa, line 56: rather use ```
@Override
    protected String createProcessTitle(Map<String, String> context) {
        return super.createProcessTitle(context);
    }
    ```?

src/main/java/sirius/biz/codelists/jdbc/SQLCodeListImportJobFactory.java
```diff
@@ -46,6 +48,16 @@ public String getName() {
         return SQLCodeListEntry.class;
     }
 
+    @Override
+    protected void executeTask(ProcessContext process) throws Exception {
+        process.addLink(new ProcessLink().withLabel("$CodeList")
+                                         .withUri("/code-list/" + process.require(codeListParameter).getIdAsString()));
+        process.updateTitle(Strings.apply("%s - %s",
```
andyHa, line 55: s.a.

sabieber (approved):

PR #1030, opened 2021-03-19
Allow duplicate file names over FTP & Support UTF-8 file names in FTP uplinks

jmuscireum:


sabieber (approved):

src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPUplinkConnectorConfig.java
```diff
@@ -25,8 +26,12 @@
 
     private static final int DEFAULT_FTP_PORT = 21;
 
+    private final String encoding;
+
     protected FTPUplinkConnectorConfig(Extension config) {
         super(config);
+
+        this.encoding = config.get("encoding").getString();
```
andyHa, line 34:
```suggestion
        this.encoding = config.get("encoding").asString(StandardCharsets.UTF_8);
```
Note that the import is missing and I'm not too sure about the actual name, but this would enforce a default if no value is given (and still "" could be used to suppress a value). Or is not setting a value == UTF-8 anyway?

andyHa (commented):
see inline comment/question..

src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPUplinkConnectorConfig.java
jmuscireum: The FTPClient uses ISO-8859-1 by default, this made be believe it is the de-facto standard. Actually it seems to be UTF-8, so this change is a good idea

mkeckmkeck (approved):

idlira (approved):

sabieber (approved):

PR #1031, opened 2021-03-19
Reduces the number of identical positive audit logs.

andyHa:


jmuscireum (changes requested):
This would also swallow multiple positive entries on the same date but with different messages? Do we want this? If a user would login and then change their password, the second message would get lost.

sabieber (approved):

sabieber (approved):

idlira (approved):

jmuscireum (approved):

PR #1034, opened 2021-03-22
Adds missing texts

cschierle:
Adds a missing text for `VFSController.createFile`, which was reported by babelfish.

Fixes: OX-6711


src/main/resources/biz_de.properties
```diff
@@ -287,7 +287,9 @@ ImportMode.UPDATE_ONLY = Nur bestehende Datensätze aktualisieren
 IncidentController.totalIncidents = Anzahl Einträge
 InteractiveJobFactory.additionalMetrics = Weitere Werte
 InternationalAddressData.country = Land
+Isenguard.import = Importe
```
sabieber, line 290: The isenguard keys should be added in OX instead of biz, as the limits whith these names are only introduced in OX

sabieber (changes requested):

idlira (approved):

fhaScireum (approved):

PR #1046, opened 2021-03-29
Fixes a resource leak when exracting large files from a 7-zip archive.

andyHa:


src/main/java/sirius/biz/util/ExtractedFileBuffer.java
```diff
@@ -90,11 +91,14 @@ public long getSize() {
     }
 
     /**
-     * Closes the underlying output stream.
+     * Releases all internally acquired resources.
      */
-    public void closeOutputStream() {
+    public void cleanup() {
```
qw3ry, line 96: Why not implement (Auto)Closeable? Then it can be used with try-with-resources etc.

fhaScireum (approved):

src/main/java/sirius/biz/util/ExtractedFileBuffer.java
andyHa: It is an internal class. And the way the 7-ZIP binding works, we woudln't really be able to use the pattern anyway. However, using Closable is generally a good choice :)

PR #1047, opened 2021-04-06
Removes mandatory UI flag on field that actually is not mandatory

mkeckmkeck:
- it can be empty in case the login is set


andyHa (approved):

mkeckmkeck:
special logic around

PR #1055, opened 2021-04-20
Introduces the DelegateJournalData

idlira:
This class should be used instead of JournalData when there is a need to record changes under another entity (usually the parent entity), like recording changes to product's features under the product itself instead of logging such changes directly at the child entity.

It can be constructed using an entity ref (of the entity which shall own the record) or using the id + type.

Fixes: SIRI-298

### DIFF
--- a/src/main/java/sirius/biz/protocol/DelegateJournalData.java
+++ b/src/main/java/sirius/biz/protocol/DelegateJournalData.java
@@ -1,0 +1,148 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.protocol;
+
+import sirius.db.mixing.BaseEntity;
+import sirius.db.mixing.Composite;
+import sirius.db.mixing.EntityDescriptor;
+import sirius.db.mixing.Mixing;
+import sirius.db.mixing.annotations.AfterDelete;
+import sirius.db.mixing.annotations.AfterSave;
+import sirius.db.mixing.annotations.Transient;
+import sirius.db.mixing.types.BaseEntityRef;
+import sirius.kernel.Sirius;
+import sirius.kernel.commons.Strings;
+import sirius.kernel.health.Exceptions;
+
+import javax.annotation.Nonnull;
+import java.util.function.Supplier;
+
+/**
+ * Provides a hook which records custom messages into the system journal which can be embedded into other entities
+ * or mixins.
+ * <p>
+ * To skip a record entirely, {@link #setSilent(boolean)} can be called before the update or delete.
+ */
+public class DelegateJournalData extends Composite {
+
+    @Transient
+    private volatile boolean silent;
+
+    @Transient
+    private final BaseEntity<?> owner;
+
+    @Transient
+    private final Supplier<String> targetId;
+
+    @Transient
+    private final Supplier<String> targetType;
+
+    @Transient
+    private Supplier<String> changeMessageSupplier;
+
+    @Transient
+    private Supplier<String> deleteMessageSupplier;
+
+    /**
+     * Creates a new instance for the given id and type suppliers.
+     *
+     * @param owner      the entity which triggered the change or delete event.
+     * @param targetId   the supplier which delivers the target id under which the journal message will be created
+     * @param targetType the supplier which delivers the target entity type under which the journal message will be created
+     */
+    public DelegateJournalData(BaseEntity<?> owner, Supplier<String> targetId, Supplier<String> targetType) {
+        this.owner = owner;
+        this.targetId = targetId;
+        this.targetType = targetType;
+    }
+
+    /**
+     * Creates a new instance for the given entity reference.
+     *
+     * @param owner     the entity which triggered the change or delete event.
+     * @param entityRef the entity reference under which the journal message will be created
+     */
+    public DelegateJournalData(BaseEntity<?> owner, @Nonnull BaseEntityRef<?, ?> entityRef) {
+        this(owner, entityRef::getIdAsString, () -> Mixing.getNameForType(entityRef.getType()));
+    }
+
+    /**
+     * Defines the supplier responsible to deliver the message to write in the protocol upon entity changes.
+     * <p>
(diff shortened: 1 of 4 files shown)


sabieber (approved):

src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -0,0 +1,148 @@
...
+    private Supplier<String> deleteMessageSupplier;
+
+    /**
+     * Creates a new instance for the given id and type suppliers.
+     *
+     * @param owner      the entity which triggered the change or delete event.
+     * @param targetId   the supplier which delivers the target id under which the journal message will be created
+     * @param targetType the supplier which delivers the target entity type under which the journal message will be created
+     */
+    public DelegateJournalData(BaseEntity<?> owner, Supplier<String> targetId, Supplier<String> targetType) {
```
andyHa, line 59: we commonly use targetType, targetId as order elsewhere

src/main/java/sirius/biz/protocol/JournalData.java
```diff
@@ -129,18 +130,45 @@ public static void addJournalEntry(BaseEntity<?> entity, String changes) {
...
-        if (!Sirius.isFrameworkEnabled(Protocols.FRAMEWORK_PROTOCOLS) || entity.isNew()) {
+    /**
+     * Adds an entry to the journal of the given entity.
+     *
+     * @param targetId   the id of the entity under which the entity will be written
+     * @param targetType the type of the entity under which the entity will be written
+     * @param targetName the name identifying the entity, which can differ from the owner entity
+     * @param changes    the entry to add to the journal
+     */
+    public static void addJournalEntry(String targetId, String targetType, String targetName, String changes) {
```
andyHa, line 141: s.a.

andyHa: also, should we skip targetName here or force it to be "external" or kill it entirely. At this point there is no way or properly filling this in a consistent manner

andyHa (changes requested):

mkeckmkeck (approved):

src/main/java/sirius/biz/protocol/JournalData.java
idlira: Good question. Currently the targetName ain't used anywhere. For this scenario it could be helpful in the future, especially if make it part of the searchable fields or turn it to be the type+id of the changed entity, so one could search entries based on the id of the child entity.

src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -0,0 +1,148 @@
...
+            || !Sirius.isFrameworkEnabled(Protocols.FRAMEWORK_JOURNAL)
+            || messageSupplier == null
+            || owner.isNew()) {
+            return;
+        }
+
+        try {
+            String message = messageSupplier.get();
+            if (Strings.isFilled(message)) {
+                JournalData.addJournalEntry(targetType.get(), targetId.get(), owner.toString(), message);
```
andyHa, line 142: using owner.toString here is wrong as for consistency reasons we'd have to use target.toString() - I'd rather kill this field entirely and only store the owner.toString() during a delete of the main entity so that we can search for vanished entities. Everything else can be found via the targetType/targetId

andyHa: Therefore JournalData#onDelete must be updated accordingly

andyHa (changes requested):

src/main/java/sirius/biz/protocol/JournalData.java
```diff
@@ -133,12 +133,12 @@ public static void addJournalEntry(BaseEntity<?> entity, String changes) {
...
      * Adds an entry to the journal of the given entity.
      *
-     * @param targetId   the id of the entity under which the entity will be written
      * @param targetType the type of the entity under which the entity will be written
+     * @param targetId   the id of the entity under which the entity will be written
      * @param targetName the name identifying the entity, which can differ from the owner entity
      * @param changes    the entry to add to the journal
      */
-    public static void addJournalEntry(String targetId, String targetType, String targetName, String changes) {
+    public static void addJournalEntry(String targetType, String targetId, String targetName, String changes) {
```
sabieber: This is a breaking change, isnt it?

src/main/java/sirius/biz/protocol/JournalEntry.java
```diff
@@ -67,11 +69,16 @@
...
+     * The identifier is composed of {@link EntityDescriptor#getType()}-{@link BaseEntity#getIdAsString()}
+     * For entities using {@link JournalData}, it is equal to the {@link #TARGET_TYPE}-{@link #TARGET_ID}
+     * For entities using {@link DelegateJournalData}, it will contain the type and if of the entity which
+     * caused the journal entry, since the target fields contains the parent entity owning this record.
      */
-    public static final Mapping TARGET_NAME = Mapping.named("targetName");
+    public static final Mapping CONTENT_IDENTIFIER = Mapping.named("contentIdentifier");
     @SearchContent
-    private String targetName;
+    private String contentIdentifier;
```
sabieber, line 81: This is a breaking change, isnt it?

sabieber (changes requested):
If the marked changes are indeed breaking please mark the PR as such and also describe the breaking changes and how to mitigate them in the description.
Also not that a major dev version should be tagged immediately after merging it.

If they are not breaking please ignore :D

src/main/java/sirius/biz/protocol/JournalEntry.java
idlira: No, the field wasn't used by no one unless a template. (Technically speaking yes, but without effect to us).

src/main/java/sirius/biz/protocol/JournalData.java
idlira: No, because this method is new within this PR.

src/main/java/sirius/biz/protocol/JournalEntry.java
andyHa: Technically we dropped the unused field targetName and introduced the new field contentIdentifier :-P

PR #1056, opened 2021-04-20
Provides an ImportHelper and an EntityImportHandlerExtender to properly im- and export MultiLanguageStrings

andyHa:


src/main/java/sirius/biz/translations/MultiLanguageStringHelper.java
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Creates a new value builder just like {@link #createReplacement()} which is filled with the given fallback value.
+     *
+     * @param fallbackValue the new fallback value to set
+     * @return a new builder which can be directly put into the {@link sirius.biz.importer.ImportContext} to update
+     * a multi language string field.
+     */
+    public static MultiLanguageStringValue createReplacement(String fallbackValue) {
+        return new MultiLanguageStringValue(false).withFallback(fallbackValue);
```
mkeckmkeck, line 107:
```suggestion
        return new MultiLanguageStringValue(true).withFallback(fallbackValue);
```

mkeckmkeck (changes requested):

src/main/java/sirius/biz/translations/MultiLanguageStringExtender.java
```diff
@@ -0,0 +1,111 @@
...
+        // Otherwise we use the value as string and applied as configured by the MultiLanguageStringHelper...
+        MultiLanguageStringHelper multiLanguageStringHelper =
+                context.getImporter().findHelper(MultiLanguageStringHelper.class);
+        if (multiLanguageStringHelper.isReplaceOnImport()) {
+            multiLanguageString.clear();
+        }
+
+        if (Strings.isFilled(multiLanguageStringHelper.getImportLanguage())) {
+            multiLanguageString.put(multiLanguageStringHelper.getImportLanguage(), value.getString());
+        } else if (multiLanguageString.isWithFallback()) {
```
idlira, line 75: I think we must split this if. The provided text could be used both as language and fallback...
if isWithFallback -> setFallback
if getImportLanguage non empty -> put lang

The if in the export does it right!

src/main/java/sirius/biz/translations/MultiLanguageStringHelper.java
```diff
@@ -0,0 +1,232 @@
...
+            data.put(lang, value);
+
+            return this;
+        }
+    }
+
+    private boolean replaceOnImport;
+    private String importLanguage;
+    private String defaultImportLanguage;
+    private String exportLanguage;
```
idlira, line 75: Do we really need to have pairs for import export, or could simply join/share them?

idlira (changes requested):

idlira (approved):

PR #1059, opened 2021-04-22
Adds a method to easily write journal entries

idlira:
For an owning type+id and a target class

Fixes: SIRI-298

### DIFF
--- a/src/main/java/sirius/biz/protocol/DelegateJournalData.java
+++ b/src/main/java/sirius/biz/protocol/DelegateJournalData.java
@@ -125,8 +125,8 @@ public class DelegateJournalData extends Composite {
     /**
      * Adds an entry to the journal of the given entity.
      *
-     * @param targetType             the type of the entity under which the entity will be written
-     * @param targetId               the id of the entity under which the entity will be written
+     * @param targetType             the type of the entity under which the journal entity will be written
+     * @param targetId               the id of the entity under which the journal entity will be written
      * @param contentIdentifierClass the type of the actual entity which was changed
      * @param contentIdentifierId    the id of the actual entity which was changed
      * @param changes                the entry to add to the journal, automatically appended by the class's label
--- a/src/main/java/sirius/biz/protocol/DelegateJournalData.java
+++ b/src/main/java/sirius/biz/protocol/DelegateJournalData.java
@@ -168,12 +168,7 @@ public class DelegateJournalData extends Composite {
         try {
             String message = messageSupplier.get();
             if (Strings.isFilled(message)) {
-                JournalData.addJournalEntry(targetType.get(),
-                                            targetId.get(),
-                                            Strings.apply("%s-%s",
-                                                          Mixing.getNameForType(owner.getClass()),
-                                                          String.valueOf(owner.getId())),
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -118,6 +122,21 @@ public void setSilent(boolean silent) {
         this.silent = silent;
     }
 
+    /**
+     * Adds an entry to the journal of the given entity.
+     *
+     * @param targetType the type of the entity under which the entity will be written
+     * @param targetId   the id of the entity under which the entity will be written
```
cschierle, line 129:
```suggestion
     * @param targetType the type of the entity under which the journal entry will be added
     * @param targetId   the id of the entity under which the journal entry will be added
```
Maybe make the wording here less confusing 😅

cschierle (approved):

src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -118,6 +122,21 @@ public void setSilent(boolean silent) {
...
 
+    /**
+     * Adds an entry to the journal of the given entity.
+     *
+     * @param targetType the type of the entity under which the journal entry will be created
+     * @param targetId   the id of the entity under which the journal entry will be created
+     * @param aClass     the class of the actual entity which was changed
+     * @param changes    the entry to add to the journal, automatically appended by the class's label
+     */
+    public static void createJournalEntry(String targetType, String targetId, Class<?> aClass, String changes) {
```
andyHa, line 133: Class<?> aClass
```suggestion
    public static void createJournalEntry(String targetType, String targetId, Class<?> contentIdentifierClass, String contentIdentifierId, String changes) {
```

andyHa (changes requested):

src/main/java/sirius/biz/protocol/DelegateJournalData.java
idlira: For the direct deletion via (like via Mongo), we have no ID to provide if we don't make a lookup before...

sabieber (approved):

src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -118,6 +122,30 @@ public void setSilent(boolean silent) {
...
+    /**
+     * Adds an entry to the journal of the given entity.
+     *
+     * @param targetType             the type of the entity under which the journal entity will be written
+     * @param targetId               the id of the entity under which the journal entity will be written
+     * @param contentIdentifierClass the type of the actual entity which was changed
+     * @param contentIdentifierId    the id of the actual entity which was changed
+     * @param changes                the entry to add to the journal, automatically appended by the class's label
+     */
+    public static void createJournalEntry(String targetType,
```
andyHa, line 134: use this in  `private void createJournalEntry(Supplier<String> messageSupplier)'  below to removed redundant code?

andyHa (commented):
...

PR #1060, opened 2021-04-23
Adds "mappings" support to LookupTable andprovides/enhances Countries and Languages.

andyHa:


src/main/resources/component-biz.conf
```diff
@@ -614,8 +614,40 @@ lookup-tables {
...
+        mappingsField = "mappings"
+    }
+
+    # Specifies the lookup table which is used by "Countries".
+    # Note that this can be further customized by applications (i.e. store in Jupiter etc.)
+    countries {
+        codeCase = "lower"
+    }
+
+    # Specifies the lookup table to lists all actively enabled countries in "Countries".
```
sabieber, line 629:
```suggestion
    # Specifies the lookup table which lists all actively enabled countries in "Countries".
```

src/main/resources/component-biz.conf
```diff
@@ -614,8 +614,40 @@ lookup-tables {
...
+        codeCase = "lower"
+    }
+
+    # Specifies the lookup table for which is used by "Languages".
+    # Note that this can be further customized by applications (i.e. store in Jupiter etc.)
+    languages {
+        codeCase = "lower"
+    }
+
+    # Specifies the lookup table to lists all actively enabled languages in "Languages".
```
sabieber, line 644:
```suggestion
    # Specifies the lookup table which lists all actively enabled languages in "Languages".
```

sabieber (changes requested):

src/main/resources/component-biz.conf
```diff
@@ -614,8 +614,40 @@ lookup-tables {
...
+    active-languages {
+        # Note that by default we simply redirect this to the "languags" code list. Therefore all languages
```
idlira, line 646:
```suggestion
        # Note that by default we simply redirect this to the "languages" code list. Therefore all languages
```

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -120,6 +137,85 @@ protected String normalizeCodeValue(String code) {
...
+     * This can be used to perform a two stage lookup. E.g. if an <tt>acme 1.0</tt> standard is requested, one could
+     * query <tt>acme-10</tt> as primary mapping and <tt>acme</tt> as secondary. This way, if a version dependent
+     * mapping is present, this will be used. Otherwise the base value from the standard is returned (if present).
+     *
+     * @param code             the code to fetch the mapping for
+     * @param primaryMapping   the more specific mapping to attempt to fetch
+     * @param secondaryMapping the more general mapping to fetch
+     * @return either the mapping for the first or second mapping to use or an empty optional if neither is present
+     */
+    public Optional<String> fetchMappings(String code, String primaryMapping, String secondaryMapping) {
```
idlira, line 188: What is the approach if a value cease to exist after a specific mapping-version? Setting the key as obsolete will still resolve an older mapping? Or do we need to remove the "acme" and set the old value under all previous version of acme-## ?

idlira (changes requested):

src/main/java/sirius/biz/codelists/LookupTable.java
andyHa: let's deal with this once we encounter this problem in the wild.... probably the suggested solution would be the way to go...

sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -120,6 +137,85 @@ protected String normalizeCodeValue(String code) {
...
 
+    /**
+     * Fetches the mapping for the given code.
+     * <p>
+     * Mappings are most probably only supported by IDB backed tables. If <tt>acme</tt> is given, this will
+     * return the value in <tt>mappings.acme</tt>.
+     *
+     * @param code    the code to fetch the mapping for
+     * @param mapping the name of the mapping to fetch
+     * @return the value to use or an empty optional of either no mapping is present or the code is unknown
```
cschierle, line 148:
```suggestion
     * @return the value to use or an empty optional if either no mapping is present or the code is unknown
```

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -120,6 +137,85 @@ protected String normalizeCodeValue(String code) {
...
+    public Optional<String> fetchMapping(String code, String mapping) {
+        if (Strings.isEmpty(code)) {
+            return Optional.empty();
+        }
+
+        return performFetchField(normalizeCodeValue(code), mappingsField + "." + mapping);
+    }
+
+    /**
+     * Fetches the mapping for the give code or the code itself if no mapping is present or if the code is unknown.
```
cschierle, line 160:
```suggestion
     * Fetches the mapping for the given code or the code itself if no mapping is present or if the code is unknown.
```

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -120,6 +137,85 @@ protected String normalizeCodeValue(String code) {
...
+        Optional<String> result = performFetchField(normalizeCodeValue(code), mappingsField + "." + primaryMapping);
+        if (result.isPresent()) {
+            return result;
+        }
+
+        return performFetchField(normalizeCodeValue(code), mappingsField + "." + secondaryMapping);
+    }
+
+    /**
+     * Fetches the mapping for the give code or the code itself if no mapping is present or if the code is unknown.
```
cschierle, line 202:
```suggestion
     * Fetches the mapping for the given code or the code itself if no mapping is present or if the code is unknown.
```

cschierle (approved):

mkeckmkeck (approved):

PR #1063, opened 2021-04-26
Escapes the JSON in the multiLanguageStringField to allow " in the translations

jmuscireum:
If " is used inside one of the translations, it is correctly escaped using `\`. But JSON.parse expects `\` itself to be escaped (`\\"`).

### DIFF
--- a/src/main/resources/default/taglib/w/multiLanguageTextField.html.pasta
+++ b/src/main/resources/default/taglib/w/multiLanguageTextField.html.pasta
@@ -88,7 +88,7 @@
                 fallbackLabel: '@i18n("Language.fallback")',
                 wrapperId: '___fieldId',
                 modalId: '___modalId',
-                values: JSON.parse(<i:raw>'@escapeJS(value.getAsJSON())'</i:raw>),
+                values: <i:raw>@value.getAsJSON()</i:raw>,
                 defaultLanguage: '___defaultLanguage',
                 validLanguages: validLanguages,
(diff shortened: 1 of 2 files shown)


sabieber (approved):

mkeckmkeck (approved):

src/main/resources/default/taglib/w/multiLanguageTextField.html.pasta
```diff
@@ -88,7 +88,7 @@
                 fallbackLabel: '@i18n("Language.fallback")',
                 wrapperId: '___fieldId',
                 modalId: '___modalId',
-                values: JSON.parse(<i:raw>'___value.getAsJSON()'</i:raw>),
+                values: JSON.parse(<i:raw>'@escapeJS(value.getAsJSON())'</i:raw>),
```
andyHa, line 91: Just out of curiosity, why doesn't:
`values: <i:raw>@value.getAsJSON()</i:raw>`
work?

cschierle (approved):

src/main/resources/default/taglib/w/multiLanguageTextField.html.pasta
sabieber, line 91:
```suggestion
                values: <i:raw>@escapeJS(value.getAsJSON())</i:raw>,
```

andyHa (approved):

mkeckmkeck (approved):

PR #1069, opened 2021-04-29
Provides a toString and forceName for LookupValue.

andyHa:
These are common methods and thus factored out here as otherwise
one had to repeat "value.fetchName(value.getValue)" all the time.

### DIFF
--- a/src/main/java/sirius/biz/codelists/LookupValue.java
+++ b/src/main/java/sirius/biz/codelists/LookupValue.java
@@ -122,6 +122,20 @@ public class LookupValue {
     }
 
     /**
+     * Resolves the name for the current value or returns the code itself if the name is unknown.
+     *
+     * @return the name of the current value or the code, if the name cannot be resolved. Note that this will return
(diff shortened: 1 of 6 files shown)


sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/codelists/LookupValue.java
```diff
@@ -121,6 +121,20 @@ public LookupTable getTable() {
         return getTable().resolveName(value);
     }
 
+    /**
+     * Resolves the name for the current value or returns the code itself if the name is unknown.
+     *
+     * @return the name of the current value or the code, if the name cannot be resolved. Note that this will return
+     * an empty string if no value is presnet
```
mkeckmkeck, line 128:
```suggestion
     * an empty string if no value is present
```

mkeckmkeck (approved):
typo

idlira (approved):

idlira (approved):

PR #1071, opened 2021-05-01
Move complexity out of FileImportJob into a designated class.

andyHa:
The FileImportJob and DictionaryBasedImportJob became quite complex
as they were extended to gracefully handle archives which contain
multiple files (e.g. CSVs) with different datasets.

Usually we handle this using the ArchiveImportJob, however, we still
need a lot of functionality from the DictionaryBasedImportJob.

This common code was extracted into DictionaryBasedImport which is now
used by both classes.


src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedImport.java
```diff
@@ -0,0 +1,158 @@
...
+            handleDataRow(index, row);
+        }
+    }
+
+    private void handleHeaderRow(Values row) {
+        if (dictionary.hasIdentityMapping()) {
+            boolean problemsFound = dictionary.detectHeaderProblems(row, (problem, isFatal) -> {
+                if (Boolean.TRUE.equals(isFatal)) {
+                    process.log(ProcessLog.error()
+                                          .withNLSKey(ERROR_IN_ROW_KEY)
```
idlira, line 86: Would be nice to enhance these messages with the current file name. In the process logs one sees: "Error in row #: the error...". But since the jobs can process several files, would be a good hint to show Error in file 'foo', row #: the error... "

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedImport.java
```diff
@@ -0,0 +1,158 @@
...
+
+        Watch watch = Watch.start();
+        try {
+            Context context = filterEmptyValues(dictionary.load(row, false));
+            if (!isEmptyContext(context)) {
+                rowHandler.invoke(Tuple.create(index, context));
+            }
+        } catch (HandledException e) {
+            process.incCounter(NLS.get("LineBasedJob.erroneousRow"));
+            process.handle(Exceptions.createHandled()
```
idlira, line 128: same as above.

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedImport.java
```diff
@@ -0,0 +1,158 @@
...
+                                     .to(Log.BACKGROUND)
+                                     .withNLSKey(ERROR_IN_ROW_KEY)
+                                     .set("row", index)
+                                     .set(MESSAGE_KEY, e.getMessage())
+                                     .handle());
+        } catch (Exception e) {
+            process.incCounter(NLS.get("LineBasedJob.erroneousRow"));
+            process.handle(Exceptions.handle()
```
idlira, line 136: We could expose a method so jobs can easily create process logs for errors/warning with a custom message like "Product XXX not found", respecting the message format (row number + file name)

idlira (changes requested):

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedImport.java
andyHa: this is was the handler above (line 128) should do?

idlira: I mean, a job might want to write in the process log a message like "Error in file my_file, row ##: whatever message", but keep on processing more stuff. (Example with Product not found was not the best I confess). More of a message "Cannot set property x with value y", out of a list of many properties. Or shall the jobs cumulate those and thrown them as a single error?

idlira (approved):

fhaScireum (approved):

sabieber (approved):

cschierle (approved):

PR #1078, opened 2021-05-06
Add a method to fetch a field as a boolean

jmuscireum:
This interprets 1 as true in case the underlying table doesn't know booleans and instead store integers.

### DIFF
--- a/src/main/java/sirius/biz/codelists/LookupTable.java
+++ b/src/main/java/sirius/biz/codelists/LookupTable.java
@@ -197,6 +197,32 @@ public abstract class LookupTable {
     }
 
     /**
+     * Fetches the requested field for the given code.
+     * <p>
+     * This interprets <tt>1</tt> as <tt>true</tt> in case the underlying table doesn't know booleans and instead store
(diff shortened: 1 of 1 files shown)


andyHa (approved):

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -196,6 +196,32 @@ public Value fetchFieldValue(String code, String targetField) {
         return performFetchField(normalizeCodeValue(code), targetField);
     }
 
+    /**
+     * Fetches the requested field for the given code.
+     * <p>
+     * This interprets <tt>1</tt> as <tt>true</tt> in case the underlying table doesn't know booleans and instead store
```
sabieber, line 202:
```suggestion
     * This interprets <tt>1</tt> as <tt>true</tt> in case the underlying table doesn't know booleans and instead stores
```

sabieber (approved):

mkeckmkeck (approved):

PR #1079, opened 2021-05-06
Fixes the fetching of boolean fields if the internal value is a long

jmuscireum:


src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -215,7 +215,7 @@ public Value fetchFieldValue(String code, String targetField) {
 
         Value fieldValue = performFetchField(normalizeCodeValue(code), targetField);
 
-        if (fieldValue.is(Integer.class)) {
+        if (fieldValue.isNumeric()) {
             return fieldValue.map(value -> Integer.valueOf(1).equals(value.getInteger()));
```
andyHa, line 219:
```suggestion
            return fieldValue.map(value -> value.asInt(0) == 1);
```
Feels a bit safer ^^

mkeckmkeck (approved):

andyHa (approved):

PR #1084, opened 2021-05-12
Permit to specify a custom name for the "Row" timing/counter.

andyHa:
This is beneficial when importing data from many files as then one row doesn't equal another.

Therefore we also log the name of each file being imported.


src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
```diff
@@ -88,6 +90,19 @@ public ImportFile ignoreEmptyFields() {
             return this;
         }
 
+        /**
+         * Specifies a custom counter name to be used for each processed row.
+         *
+         * @param rowCounterName the name of the coutner which will be smart translated using
```
sabieber, line 96:
```suggestion
         * @param rowCounterName the name of the counter which will be smart translated using
```

sabieber (approved):

jmuscireum (approved):

qw3ry (approved):

PR #1089, opened 2021-05-18
Prevents the UserAccount-ImportHandlers from failing to loadForFind

qw3ry:
SIRI 355


qw3ry:
This is probably the wrong way to fix the problem

PR #1122, opened 2021-06-28
Fixes for lookupValueField, performs a scan instead of suggest if no query is given 

meg1502:


src/main/resources/default/taglib/w/lookupValue.html.pasta
```diff
@@ -38,6 +38,7 @@
         new TokenAutocomplete({
             name: '@name',
             selectMode: SelectModes.SINGLE,
+            allowCustomEntries: false,
```
andyHa, line 41: shoudln't this be a parameter / bound to the LookupField?

sabieber: Yes i also think this should be read from the lookup field, as it has a constructor param controling this

sabieber (approved):

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

cschierle (approved):

PR #1126, opened 2021-06-29
Adds optional parameter to lookupValue tag & shows multi suggestion earlier

meg1502:
Fixes: OX-6973

### DIFF
--- a/src/main/resources/default/taglib/w/lookupValue.html.pasta
+++ b/src/main/resources/default/taglib/w/lookupValue.html.pasta
@@ -7,6 +7,7 @@
 <i:arg name="helpKey" type="String" default=""/>
 <i:arg name="help" type="String" default="@i18n(helpKey)"/>
 <i:arg name="required" type="boolean" default="false"/>
+<i:arg name="optional" type="boolean" default="false"/>
 <i:arg name="readonly" type="boolean" default="false"/>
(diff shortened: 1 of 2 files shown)


sabieber (approved):

ymo-sci (approved):

src/main/resources/default/taglib/w/lookupValue.html.pasta
```diff
@@ -7,6 +7,7 @@
 <i:arg name="helpKey" type="String" default=""/>
 <i:arg name="help" type="String" default="@i18n(helpKey)"/>
 <i:arg name="required" type="boolean" default="false"/>
+<i:arg name="optional" type="boolean" default="false"/>
```
ymo-sci, line 10: What is the difference between "not optional" and "required"?
Aren't they the same?

sabieber: required only marks the field as such with a * next to the field label. optional actually changes the behaviour. This is quite misleading and can result in confusing configuration but its the way it is done in other wondergem taglibs. For the tycho versions of the fields this will be changed to an `optional` flag that changes behaviour and a classlist parameter that can change the appearance of the field (like required and admin-only)

idlira (approved):

sabieber (approved):

PR #1127, opened 2021-06-29
Improves the footer of Tycho templates.

andyHa:
grafik

We now include the user/tenant name and a deep-link to be sent to support agents. This will perform the proper tenant selection and then redirect to the page the user is currently visiting.

### DIFF
--- a/src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
+++ b/src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
@@ -21,7 +21,7 @@
                         copyToClipboard(window.location.protocol + window.location.host + '/tenants/select/@user().getTenantId()?goto=' + encodeURI(window.location.href));
 
                         clearMessages();
-                        addSuccessMessage('Support-Link in Zwischenablage kopiert');
+                        addSuccessMessage('@i18n("footer.html.supportLinkCopiedSuccessfully")');
                     }
                 </script>
             </i:if>
--- a/src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
+++ b/src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
@@ -14,15 +14,17 @@
         <t:permission permission="flag-logged-in">
             <i:if test="user().is(sirius.biz.tenants.Tenant.class)">
                 <span class="d-none d-md-inline-block text-muted small ml-2 pl-4 pr-0 bl-gray">
-                    <a id="supportLink" href="javascript:copySupportLink()" class="text-sirius-gray-dark"><i class="fa fa-link"></i></a>
+                    <a id="supportLink" class="text-sirius-gray-dark cursor-pointer"><i class="fa fa-link"></i></a>
                 </span>
                 <script type="text/javascript">
-                    function copySupportLink() {
-                        copyToClipboard(window.location.protocol + window.location.host + '/tenants/select/@user().getTenantId()?goto=' + encodeURI(window.location.href));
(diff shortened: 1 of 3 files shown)


src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
```diff
@@ -0,0 +1,30 @@
...
+               <b>@user().getUserName()</b>
+               <i:if test="isFilled(user().getTenantName())">(@user().getTenantName())</i:if>
+            </span>
+        </t:permission>
+    </i:block>
+    <i:block name="footer-right-end">
+        <t:permission permission="flag-logged-in">
+            <i:if test="user().is(sirius.biz.tenants.Tenant.class)">
+                <span class="d-none d-md-inline-block text-muted small ml-2 pl-4 pr-0 bl-gray">
+                    <a id="supportLink" href="javascript:copySupportLink()" class="text-sirius-gray-dark"><i class="fa fa-link"></i></a>
```
meg1502, line 17: can we bind this action in the script portion instead of having javascript call directly in the href attribute?

meg1502 (changes requested):

src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
meg1502: maybe even also add the url itself to the href attribute, this additionaly enables right click --> copy link url which might prevent some misunderstandings for the support staff if both work :)

src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
```diff
@@ -0,0 +1,30 @@
...
+                </span>
+                <script type="text/javascript">
+                    function copySupportLink() {
+                        copyToClipboard(window.location.protocol + window.location.host + '/tenants/select/@user().getTenantId()?goto=' + encodeURI(window.location.href));
+
+                        clearMessages();
+                        addSuccessMessage('Support-Link in Zwischenablage kopiert');
```
meg1502, line 24: @i18n ? :)

sabieber (approved):

src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
andyHa: what's the benefit?

ymo-sci (approved):
i suppose the empty tycho-biz.js.pasta file is for future use?

src/main/resources/default/extensions/tycho-page-menu/footer.html.pasta
meg1502: In short it presents the user with less options, and increases the chances that people will successfully trigger the function.

The presence of this element tells the browser that this is a link with a destination. With that, the browser will show the Open In New Tab / Window function (also triggered when you use shift+click or middle mouse) and the option to copy the link. Doing so will result in opening the same page without the function being called so the actual link is not being copied(resulting in user frustration). Also this shows the ‘ javascript:copySupportLink()’ on hover in the browser status bar, which is confusing and not user friendly imo. Maybe we shouldn’t even use <a> tag but a span/button instead?

meg1502: Also, this goes against the principle of seperating structure(html) from code(script)

andyHa:
> i suppose the empty tycho-biz.js.pasta file is for future use?

bingo

PR #1142, opened 2021-07-12
Adds a way to retrieve all entries matching a lookup of idb tables

sabieber:
Previously there were only ways to easily either get all entries of a table or lookup a single entry.

Fixes: OX-6913

### DIFF
--- a/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
@@ -134,6 +134,8 @@ class CodeListLookupTable extends LookupTable {
 
     @Override
     protected Stream<LookupTableEntry> performQuery(String lang, String lookupPath, String lookupValue) {
+        // This would need a complex caching strategy as always fetching the DB would be too expensive.
+        // Could be implemented when needed.
         throw new UnsupportedOperationException();
     }
 }
--- a/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
@@ -133,7 +133,7 @@ class CodeListLookupTable extends LookupTable {
     }
 
     @Override
-    protected Stream<LookupTableEntry> performLookupScan(String lang, String lookupPath, String lookupValue) {
+    protected Stream<LookupTableEntry> performQuery(String lang, String lookupPath, String lookupValue) {
         throw new UnsupportedOperationException();
     }
 }
--- a/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CodeListLookupTable.java
@@ -131,4 +131,9 @@ class CodeListLookupTable extends LookupTable {
     protected Stream<LookupTableEntry> performScan(String lang) {
         return codeLists.getEntries(codeList).stream().map(entry -> extractEntryData(entry, lang));
     }
+
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/biz/codelists/CodeListLookupTable.java
```diff
@@ -131,4 +131,9 @@ private boolean filter(CodeListEntry<?, ?> entry, String searchTerm, String lang
     protected Stream<LookupTableEntry> performScan(String lang) {
         return codeLists.getEntries(codeList).stream().map(entry -> extractEntryData(entry, lang));
     }
+
+    @Override
+    protected Stream<LookupTableEntry> performLookupScan(String lang, String lookupPath, String lookupValue) {
```
andyHa, line 136: Optional: "performQuery" - scan is used to indicate that "no filter is applied at all"

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -731,4 +730,22 @@ public boolean canScan() {
...
+    /**
+     * Enumerates all entries matching the given lookup in the table using the given language.
+     *
+     * @param lang        the language to translate the name and description to
+     * @param lookupPath  the field to search in
+     * @param lookupValue the value to search for
+     * @return a stream of all entries matching the lookup or an empty stream if scanning isn't supported
+     */
+    public Stream<LookupTableEntry> lookupScan(String lang, String lookupPath, String lookupValue) {
+        if (!canScan()) {
```
andyHa, line 743: IMHO this check can be skipped. "canScan" is rather a "is it smart to scan" as some lists are very big and should not be output completeley. However, as a filter is applied here, we can (hopefully) safely always perform a query

idlira (approved):

src/main/java/sirius/biz/codelists/CodeListLookupTable.java
```diff
@@ -131,4 +131,9 @@ private boolean filter(CodeListEntry<?, ?> entry, String searchTerm, String lang
     protected Stream<LookupTableEntry> performScan(String lang) {
         return codeLists.getEntries(codeList).stream().map(entry -> extractEntryData(entry, lang));
     }
+
+    @Override
+    protected Stream<LookupTableEntry> performQuery(String lang, String lookupPath, String lookupValue) {
+        throw new UnsupportedOperationException();
```
meg1502, line 137: Maybe a a small comment why this is unsupported

mainly to differiantiate between "not yet implemented" and "this method will never work on this class"

meg1502 (approved):

src/main/java/sirius/biz/codelists/IDBLookupTable.java
```diff
@@ -48,6 +48,7 @@
     private static final String CACHE_PREFIX_NORMALIZE_WITH_MAPPING = "normalize-with-mapping";
     private static final String CACHE_PREFIX_REVERSE_LOOKUP = "reverse-lookup-";
     private static final String CACHE_PREFIX_FETCH_OBJECT = "fetch-object-";
+    private static final String CACHE_PREFIX_LOOKUP_SCAN = "lookup-scan-";
```
andyHa, line 51: not used (?)

andyHa (approved):

cschierle (approved):

ymo-sci (approved):

qw3ry (approved):

mkeckmkeck (approved):

PR #1152, opened 2021-07-14
Prevents renaming a file into an invalid file name

sabieber:
... as this would result in the storage space to be no longer accessible.

Fixes: SIRI-382

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
+++ b/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
@@ -405,6 +405,12 @@ public abstract class VirtualFile extends Composable implements Comparable<Virtu
      * @throws HandledException if the file cannot be renamed
      */
     public void rename(String newName) {
+        if (!newName.equals(utils.replaceIllegalFileChars(newName, true))) {
+            throw Exceptions.createHandled()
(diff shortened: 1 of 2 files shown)


ymo-sci (approved):

andyHa (approved):

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -405,6 +405,12 @@ public boolean tryRename(String newName) {
      * @throws HandledException if the file cannot be renamed
      */
     public void rename(String newName) {
+        if (!newName.equals(utils.replaceIllegalFileChars(newName, true))) {
```
andyHa, line 408: or extract and reuse the exact check...:
```
if (Strings.isEmpty(name)
            || StorageUtils.SANITIZE_ILLEGAL_FILE_CHARS.split(name).length > 1
            || StorageUtils.SANITIZE_SLASHES.split(name).length > 1) {
            throw new IllegalArgumentException("A filename must be filled and must not contain illegal characters.");
        }

mkeckmkeck (approved):

andyHa (approved):

idlira (approved):

meg1502 (approved):

PR #1153, opened 2021-07-14
SAML Login: Also uses the "role" claim as a potential source of user permissions.

andyHa:
This field is used by Azure AD to send "readable" roles as the groups itself only contain the internal IDs.

Fixes: SIRI-384


src/main/java/sirius/biz/tenants/SAMLController.java
```diff
@@ -103,10 +104,14 @@ public void samlLogin(WebContext ctx) {
         } else {
             verifyUser(response, user);
         }
+        manager.onExternalLogin(ctx, user);
 
+        // Re-resolve and install the newly created or updated user in the session.
+        // This also flushes all cached again, as we updated some internal data within
```
mkeckmkeck, line 110:
```suggestion
        // This also flushes all caches again, as we updated some internal data within
```

mkeckmkeck (approved):

PR #1155, opened 2021-07-14
Defaults to the base table's code case for filtered lookup tables

cschierle:
This way, we don't need to repeat the code case in the config for each filtered table, where the base table's code case differs from the default setting.

Fixes: SIRI-385

### DIFF
--- a/src/main/java/sirius/biz/codelists/IDBFilteredLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/IDBFilteredLookupTable.java
@@ -36,6 +36,10 @@ class IDBFilteredLookupTable extends LookupTable {
         super(extension);
         this.baseTable = baseTable;
         this.filterSet = filterSet;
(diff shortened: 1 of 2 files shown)


cschierle:
does not work as consistently as I hoped... :(

src/main/java/sirius/biz/codelists/IDBFilteredLookupTable.java
```diff
@@ -36,6 +36,10 @@
         super(extension);
         this.baseTable = baseTable;
         this.filterSet = filterSet;
+        this.codeCase = extension.get(LookupTable.CONFIG_KEY_CODE_CASE_MODE)
```
andyHa, line 39: move to super?

andyHa, line 39: add something similar for CustomLookup table? like, ensure that both tables use the same and use this or complain if different?

PR #1156, opened 2021-07-15
Purges BlobStorageSpace cache when deleting blobVariants

meg1502:


sabieber (approved):

qw3ry (approved):

andyHa (changes requested):
Breaks API boundaries. I'd rather provide a delete method for variants which take care of the cache coherence...

idlira (approved):

andyHa (approved):

PR #1162, opened 2021-07-16
Renames method to better reflect what it does

meg1502:
As we renamed a protected method, this is technically a breaking change, but most probably without any action required.


andyHa (approved):

qw3ry (approved):
Potentially breaking

jmuscireum (approved):

andyHa:
> Potentially breaking

Next release will be breaking due to a web update anyway, but thanks for pointing out..

PR #1166, opened 2021-07-20
Jobs documentation improvements

qw3ry:
- The `@AutoImport`-Annotation supports a flag `hidden` (default `false`). When the flag is set, the property does not show up in the job documentation.
- The `@AutoImport`-Annotation supports a new parameter controlling whether the property is marked as _required_ or not. Default is _AUTO_DETECT_ which uses a now improved framework logic.


src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -335,8 +335,8 @@ public ImportDictionary getImportDictionary() {
         return descriptor.getProperties()
                          .stream()
                          .filter(property -> property.getAnnotation(AutoImport.class)
-                                                     .map(AutoImport::permissions)
-                                                     .map(currentUser::hasPermissions)
+                                                     .map(annotation -> currentUser.hasPermissions(annotation.permissions())
+                                                                        && !annotation.hidden())
```
idlira, line 339: leave .map as is and add a .filter ?

fhaScireum (approved):

idlira (changes requested):

sabieber (approved):

src/main/java/sirius/biz/importer/AutoImport.java
```diff
@@ -28,4 +36,20 @@
...
+    boolean hidden() default false;
+
+    /**
+     * Marks the field as required in the job documentation.
+     * <p>
+     * In most cases, the framework logic will work this out by itself. Use this flag if it doesn't.
+     *
+     * @return true if the field is required in the import
+     */
+    RequiredStatus value() default RequiredStatus.AUTO_DETECT;
```
andyHa, line 54: Optional: name this required() or the like?

andyHa (approved):

src/main/java/sirius/biz/importer/AutoImport.java
qw3ry: Advantage of `value` is that you can write `@AutoImport(RequiredStatus.OPTIONAL)` which reads kind of nicely compared to `@AutoImport(required = RequiredStatus.OPTIONAL)`

idlira (approved):

ymo-sci (approved):

jmuscireum (approved):

mkeckmkeck (approved):

PR #1183, opened 2021-08-03
Ensures that language keys in MultiLanguageStrings cannot be empty

cschierle:
Keys were already expected to be Nonnull, but empty keys could still be added forcefully. This produced fatal errors like persisted MultiLanguageStrings throwing NPEs when being loaded back from the database. Empty keys will now trigger an IllegalArgumentException and thus no longer be accepted! Missing translations should instead be handled via MultiLanguageStrings with fallback enabled.

Fixes: SIRI-375

### DIFF
--- a/src/test/java/sirius/biz/translations/MultiLanguageStringSpec.groovy
+++ b/src/test/java/sirius/biz/translations/MultiLanguageStringSpec.groovy
@@ -1,0 +1,61 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.biz.translations
+
+import sirius.kernel.BaseSpecification
+
+class MultiLanguageStringSpec extends BaseSpecification {
+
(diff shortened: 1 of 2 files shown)


src/test/java/sirius/biz/translations/MultiLanguageStringSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+        mls.addText("en", "some text")
+        mls.addText("de", "irgendein text")
+        Map<String, String> map = new HashMap<>()
+        map.put("en", "some text")
+        map.put("", "empty key")
+        mls.setData(map)
+        then:
+        thrown(IllegalArgumentException)
+    }
+
```
fhaScireum, line 60: Maybe insert simple happy path test here that show that MultiLanguageString doesn't always throw an exception? :)

fhaScireum (approved):

fhaScireum (approved):

andyHa (approved):

sabieber (approved):

ymo-sci (approved):

mkeckmkeck (approved):

PR #1193, opened 2021-08-12
Stops processing entries if a vital error was found

idlira:
When a vital error is detected (eg. csv mismatch against the provided dictionary), we still tried to further process the files defined for processing. Such vital errors cancel the job execution.

Fixes: OX-7286

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
@@ -12,6 +12,7 @@ import sirius.biz.importer.format.ImportDictionary;
 import sirius.biz.process.ProcessContext;
 import sirius.biz.process.logs.ProcessLog;
 import sirius.biz.util.ExtractedFile;
+import sirius.kernel.async.TaskContext;
 import sirius.kernel.commons.Callback;
 import sirius.kernel.commons.Context;
 import sirius.kernel.commons.Tuple;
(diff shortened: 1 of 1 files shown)


meg1502 (approved):

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
```diff
@@ -138,6 +139,9 @@ protected void importEntries() throws Exception {
         handledFiles.clear();
 
         for (ImportFile importFile : importFiles) {
+            if (!TaskContext.get().isActive()) {
```
andyHa, line 142: add same check in L155 or is this already checked by extractAllFiles?

idlira: We return and exit the whole method, not only the for-loop.

andyHa: yes but this could remain true and turn to false in the loop below...

cschierle (approved):

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
```diff
@@ -148,6 +152,9 @@ protected void importEntries() throws Exception {
             handledFiles.add(importFile.filename);
         }
 
+        if (!TaskContext.get().isActive()) {
```
andyHa, line 155: nvm - extractAllFiles already performs this check

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
```diff
@@ -148,6 +152,9 @@ protected void importEntries() throws Exception {
...
+            return;
+        }
```
andyHa, line 157:
```suggestion
```

andyHa (approved):

ymo-sci (approved):

PR #1194, opened 2021-08-12
Automatically drops the dollar sign for timings

idlira:
When adding timings /counters, the stored keys are translated by default. Therefore we store them without the dollar sign. Sometimes a key is already at hand with the sign, so we still can use it for timings, without having to define a new key.

### DIFF
--- a/src/main/java/sirius/biz/process/ProcessEnvironment.java
+++ b/src/main/java/sirius/biz/process/ProcessEnvironment.java
@@ -95,6 +95,11 @@ class ProcessEnvironment implements ProcessContext {
 
     @Override
     public void addTiming(String counter, long millis, boolean adminOnly) {
+        // Drops the dollar sign used in NLS.smartGet, because the counters are always translated.
+        if (counter.startsWith("$")) {
+            counter = counter.substring(1);
+        }
+
         if (adminOnly) {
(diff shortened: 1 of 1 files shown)


cschierle (approved):

src/main/java/sirius/biz/process/ProcessEnvironment.java
```diff
@@ -95,6 +95,11 @@ public void addTiming(String counter, long millis) {
 
     @Override
     public void addTiming(String counter, long millis, boolean adminOnly) {
+        // Drops the dollar sign used in NLS.smartGet, because the counters are always translated.
+        if (counter.startsWith("$")) {
+            counter = counter.substring(1);
```
meg1502, line 100: maybe return first if Strings.isEmpty(counter) to prevent NPE? , but idk throwing an error if counter is null might be the desireable outcome

meg1502 (approved):

src/main/java/sirius/biz/process/ProcessEnvironment.java
idlira: NPE here is pretty much a programming error, so an error is OK.
We could improve the error (instead of NPE) by annotating all addTiming/incCounter with @Nonnull. Not sure I wanna do that now 🤪

PR #1202, opened 2021-08-24
No longer always fills the total items with the current items

meg1502:
If we don't know the total size, we don't want to supply a total item count to the page. This bug was introduced while trying to fix another bug, where fetchTotalItems wasn't applied for non-full pages that weren't at start=0.
This change keeps this fix so total items will only be filled when
 **withTotalCount is enforced**

**In all other cases the totalItems will not be supplied to the Page at all.**

This behavior/bug was introduced with https://github.com/scireum/sirius-biz/pull/923

This filled totalItems produces the following error in the UI:

Screenshot 2021-08-24 at 18 24 51

### DIFF
--- a/src/main/java/sirius/biz/web/BasePageHelper.java
+++ b/src/main/java/sirius/biz/web/BasePageHelper.java
@@ -57,7 +57,7 @@ public abstract class BasePageHelper<E extends BaseEntity<?>, C extends Constrai
     protected List<QueryField> searchFields = Collections.emptyList();
     protected List<Tuple<Facet, BiConsumer<Facet, Q>>> facets = new ArrayList<>();
     protected int pageSize = DEFAULT_PAGE_SIZE;
-    protected boolean fetchTotalCount = false;
+    protected boolean withTotalCount;
     protected boolean debugging;
 
     protected BasePageHelper(Q query) {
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -411,11 +412,12 @@ protected void enforcePaging(Page<E> result, List<E> items) {
...
-        if (fetchTotalCount && (originalSize > pageSize || result.getStart() > pageSize)) {
-            result.withTotalItems((int) baseQuery.count());
-        } else {
-            result.withTotalItems(items.size());
+            if (withTotalCount) {
+                result.withTotalItems((int) baseQuery.count());
+            }
+        } else if (withTotalCount) {
+            // we don't have any more items, so total items is end of the current page
+            result.withTotalItems(result.getEnd());
```
andyHa, line 420: isn't here an off-by-1 if originalSize == pageSize (?)  - i think thats the whole point of keeping originalSize around

meg1502: Not really if originalSize == pageSize we never remove anything from items so it doesnt matter if we use originalSize or items.size()

But i found at this point the items aren't added into the result so we can't use result.getEnd() anyway and spell it out

andyHa (approved):

fhaScireum (approved):

sabieber (approved):

jmuscireum (approved):

qw3ry (approved):

PR #1210, opened 2021-08-29
Misc improvements and fixes

andyHa:
* Hides the description for a lookup table if none is present.
* Marks all JavaScript marker classes with the "-js" suffix.
* BREAKING: Properly enforces an expected type for BizController.findExistingForTenant
* Permits ComplexLoad properties to decide when to perform autoloading.
* Greatly simplifies loading values into a MultiLanguageString.
* Improves / adds helper tags to render languages/countries.
* Permits to properly handle partial creations of entities.
* Repairs the base URLs for process logs.
* Properly parses mongo IDs which are strings and not longs.
* Fixes video per track lookups for the MongoDB onboarding engine.

### DIFF
--- a/src/main/java/sirius/biz/web/ComplexLoadProperty.java
+++ b/src/main/java/sirius/biz/web/ComplexLoadProperty.java
@@ -28,4 +28,17 @@ public interface ComplexLoadProperty {
      * @return <tt>true</tt> if loading succeeded, <tt>false</tt> if an error occurred
      */
     boolean loadFromWebContext(WebContext webContext, BaseEntity<?> entity);
+
+    /**
+     * Determines if "autoloading" should be used for the given request.
+     * <p>
+     * This can inspect the request a check if the required parameters are present in order to successfully perform
+     * an automatic load.
+     * <p>
+     * Note that a default property would check for {@code webContext.hasParameter(this.getPropertyName())}.
+     *
+     * @param webContext the request to inspect
+     * @return <tt>true</tt> if the proper parameters are filled so that the property can be loaded
+     */
+    boolean shouldAutoload(WebContext webContext);
 }
--- a/src/main/java/sirius/biz/web/ComplexLoadProperty.java
+++ b/src/main/java/sirius/biz/web/ComplexLoadProperty.java
@@ -32,7 +32,7 @@ public interface ComplexLoadProperty {
     /**
      * Determines if "autoloading" should be used for the given request.
      * <p>
-     * This can inspect the request a check if the required parameters are present in order to successfully perform
+     * This can inspect the request and check if the required parameters are present in order to successfully perform
      * an automatic load.
      * <p>
      * Note that a default property would check for {@code webContext.hasParameter(this.getPropertyName())}.
--- a/src/main/java/sirius/biz/web/SaveHelper.java
+++ b/src/main/java/sirius/biz/web/SaveHelper.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Consumer;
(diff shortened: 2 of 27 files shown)


src/main/java/sirius/biz/web/ComplexLoadProperty.java
```diff
@@ -28,4 +28,17 @@
...
+     * <p>
+     * This can inspect the request a check if the required parameters are present in order to successfully perform
+     * an automatic load.
+     * <p>
+     * Note that a default property would check for {@code webContext.hasParameter(this.getPropertyName())}.
+     *
+     * @param webContext the request to inspect
+     * @return <tt>true</tt> if the proper parameters are filled so that the property can be loaded
+     */
+    boolean shouldAutoload(WebContext webContext);
```
meg1502, line 43: maybe we could rename this - i think its kind of confusing that it has the same name as the method in the biz controller and from a method called "shouldAutoload" i would expect it to also check the annotation/permission.

I don't have a good suggestion tho, my only is maybe "hasLoadableParameter"

src/main/java/sirius/biz/web/SaveHelper.java
```diff
@@ -67,6 +68,25 @@ public SaveHelper withPostSaveHandler(Consumer<Boolean> postSaveHandler) {
...
+     * process. These are entities where some fields need to be set (like a type) before other fields
+     * can even be rendered (e.g. if their type or appearance depends on the type itself).
+     *
+     * @param skipCreatePredicate the predicate which decides if the entity should be persisted in the database
+     *                            or if the editor should be rendered again. They receive the {@link BaseEntity#isNew()}
+     *                            as if was before loading any data and may return <tt>true</tt> to save the entity
+     *                            or <tt>false</tt> to show the editor again.
+     * @return the helper itself for fluent method calls
+     */
+    public SaveHelper withSkipCreatePredicate(UnaryOperator<Boolean> skipCreatePredicate) {
```
meg1502, line 85: shouldnt this actually be a Predicate Entity -> boolean as also hinted at by the name? how can the predicate inspect the entity if it only recieves BaseEntity#isNew , shouldnt they receive the whole entity ?

src/main/java/sirius/biz/web/ComplexLoadProperty.java
andyHa: the BizController.shouldAutoload calls this method:
```
        if (property instanceof ComplexLoadProperty complexProperty) {
            return complexProperty.shouldAutoload(webContext);
        }
```

src/main/java/sirius/biz/web/SaveHelper.java
andyHa: the entity is normally visible to the predicate / closure passed in. we only forward the isNew flag, as we capture it before any changes are performed (this is the same for all other handlers...)

sabieber (approved):

src/main/java/sirius/biz/web/ComplexLoadProperty.java
```diff
@@ -28,4 +28,17 @@
      * @return <tt>true</tt> if loading succeeded, <tt>false</tt> if an error occurred
      */
     boolean loadFromWebContext(WebContext webContext, BaseEntity<?> entity);
+
+    /**
+     * Determines if "autoloading" should be used for the given request.
+     * <p>
+     * This can inspect the request a check if the required parameters are present in order to successfully perform
```
idlira, line 35:
```suggestion
     * This can inspect the request and check if the required parameters are present in order to successfully perform
```

idlira (approved):
small typo

jmuscireum (approved):

mkeckmkeck (approved):

PR #1222, opened 2021-09-07
fix a bunch of typos

ymo-sci:
fix some occurred typos


sabieber (approved):

fgoerner (approved):

meg1502 (approved):

src/main/java/sirius/biz/storage/layer2/jdbc/SQLDirectory.java
```diff
@@ -33,7 +33,7 @@
 /**
  * Stores the metadata of a {@link Directory} in the underlying JDBC database.
  * <p>
- * Note that all non trivial methods delegate to the associated {@link SQLBlobStorageSpace}.
+ * Note that all nontrivial methods delegate to the associated {@link SQLBlobStorageSpace}.
```
qw3ry, line 36:
```suggestion
 * Note that all non-trivial methods delegate to the associated {@link SQLBlobStorageSpace}.
```

qw3ry (approved):

src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlob.java
```diff
@@ -171,6 +172,7 @@
      * Stores the last modification timestamp.
      */
     public static final Mapping LAST_MODIFIED = Mapping.named("lastModified");
+    @DefaultValue("")
     private LocalDateTime lastModified = LocalDateTime.now();
```
andyHa, line 176: shouldn't we always skip this for LocalDate & LocalDateTime properties? using the given default makes almost no sense anywhere (?)

meg1502: If we want to do this we can overwrite the column method in the properties classes, can copy from MultiLanguageStringProperty

ymo-sci: Is now fixed with https://github.com/scireum/sirius-db/pull/449 as suggested

meg1502 (approved):

PR #1227, opened 2021-09-10
Fixes url after following wrongly encoded redirect headers

meg1502:
By standard, the redirect header should contain the url in ISO-8859-1;
But most browsers and a lot of other clients also support UTF-8 so some servers will use UTF-8. When UTF-8 is used, the wrongly interpreted symbols will always start with the character 'Ã'. As the chance for this symbol in uppercase to be used in a correct url is pretty low, and at this point of the code we are at our last resort anyway, we will re interpret the String as UTF-8 if we come across this character.

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
+++ b/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
@@ -1433,8 +1433,11 @@ public abstract class VirtualFile extends Composable implements Comparable<Virtu
 
             if (Strings.isEmpty(path) && !url.getPath().equals(lastConnectedURL.getPath())) {
                 // We don't have a path yet but we followed redirects so we check the new URL
-                if (headRequest.getResponseCode() == 404 && lastConnectedURL.toString().contains("Ã")) {
-                    // the redirect header contained UTF-8 das was interpreted as ISO-8859-1, try to fix this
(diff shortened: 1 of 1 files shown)


idlira (approved):

sabieber (approved):

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -1433,6 +1433,12 @@ private String parsePathFromUrl(URL url, Predicate<String> fileExtensionVerifier
 
             if (Strings.isEmpty(path) && !url.getPath().equals(lastConnectedURL.getPath())) {
                 // We don't have a path yet but we followed redirects so we check the new URL
+                if (headRequest.getResponseCode() == 404 && lastConnectedURL.toString().contains("Ã")) {
```
andyHa, line 1436: use a constant here ? I think HttpUrlConntections Or HttpResponseState or the like defiens them

andyHa: maybe also add the hint why we check agains "A" in a short inline comment to clarify the intention...

andyHa (changes requested):
...

cschierle (approved):

andyHa (approved):

idlira (approved):

mkeckmkeck (approved):

PR #1235, opened 2021-09-14
Fix the BooleanProperty#tryLoadProperty for checkboxes

qw3ry:
The checkboxes only send an empty parameter `NAME_marker`, and if it is checked an additional parameter `NAME`.
Loading the unchecked version where only the `NAME_marker` is set, sets the property to its `@DefaultValue`, which might not be appropriate


sabieber (approved):

marvinwoop (approved):

qw3ry:
Closing in favor of #1234

PR #1245, opened 2021-09-23
Custom and filtered lookup tables now fall back to title and description of their base table

cschierle:
Fixes: SIRI-436

### DIFF
--- a/src/main/java/sirius/biz/codelists/CustomLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CustomLookupTable.java
@@ -8,6 +8,7 @@
 
 package sirius.biz.codelists;
 
+import com.rometools.utils.Strings;
 import sirius.kernel.commons.Limit;
 import sirius.kernel.commons.Value;
 import sirius.kernel.settings.Extension;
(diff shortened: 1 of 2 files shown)


sabieber (approved):

src/main/java/sirius/biz/codelists/CustomLookupTable.java
```diff
@@ -118,4 +119,14 @@ public int count() {
         return Stream.concat(customTable.performQuery(lang, lookupPath, lookupValue),
                              baseTable.performQuery(lang, lookupPath, lookupValue));
     }
+
+    @Override
+    public String getTitle() {
+        return extension.getId().equals(super.getTitle()) ? baseTable.getTitle() : super.getTitle();
```
andyHa, line 125: shoudln't re rather directly ask NLS.getIfExists or the like here? This seem sort of hacky..

src/main/java/sirius/biz/codelists/CustomLookupTable.java
```diff
@@ -8,6 +8,7 @@
 
 package sirius.biz.codelists;
 
+import com.rometools.utils.Strings;
```
idlira, line 11: Sure this is the right package?

idlira (changes requested):

src/main/java/sirius/biz/codelists/CustomLookupTable.java
cschierle: thank you, code completion fail...

idlira (approved):

src/main/java/sirius/biz/codelists/CustomLookupTable.java
cschierle: Why should we treat the tables differently? I think we should use the same method to find texts for both

cschierle: ah wait, you're right... This would also return the base table's name instead of the name of the custom table. That is an issue... will change to the NLS.getIfExists

sabieber (approved):

PR #1250, opened 2021-09-27
Bugfix: Only applies empty checkbox marker for boolean properties

meg1502:
This fixes a bug where lists filled by checkboxes could not be properly be emptied.

We have to check for a BooleanProperty here, because checkboxes might also be used in templates with different values and the same name to let the user fill/unfill a ListProperty or BaseEntityRefListProperty - and these templates will also send markers for the given propertyName, so this lead to these lists be filled with a single entry 'false' when they were supposed to be empty.

Fixes: OX-7503


qw3ry (approved):

cschierle (approved):

sabieber (approved):
Alternativlösung wäre ja evtl das anhand des wertes des presence marker params zu entscheiden, aber kp ob da für booleans oder listen was anderes im template geschrieben wird

mkeckmkeck (approved):

PR #1268, opened 2021-10-13
Limits the frequency in which VirtualFiles are downloaded from URLs & Enhances the options for Jupiter Sync.

andyHa:
Fixes: OX-7437


src/main/java/sirius/biz/jupiter/JupiterSync.java
```diff
@@ -100,6 +100,10 @@
     @ConfigValue("jupiter.repository.uplink.bucket")
     private String uplinkBucket;
 
+    @ConfigValue("jupiter.repository.uplink.test-bucket")
```
idlira, line 103: Do we really need this config? Why not just set the previous param to another bucket in dev/test/staging?

jmuscireum (approved):

PR #1287, opened 2021-11-02
Allows to mark a process output as system output

cschierle:
which is only visible to admins

Fixes: OX-7702

### DIFF
--- a/src/main/resources/default/templates/biz/process/process.html.pasta
+++ b/src/main/resources/default/templates/biz/process/process.html.pasta
@@ -96,9 +96,11 @@
                                   active="page == 'logs'"/>
 
                     <i:for type="sirius.biz.process.output.ProcessOutput" var="output" items="@process.getOutputs()">
-                        <t:navboxLink icon="@output.getIcon()" label="@output.getLabel()"
-                                      url="@apply('/ps/%s/output/%s', process.getId(), output.getName())"
-                                      active="page == 'output-' + output.getName()"/>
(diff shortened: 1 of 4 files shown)


sabieber (approved):

idlira (approved):

src/main/resources/default/templates/biz/process/process.html.pasta
```diff
@@ -89,16 +89,18 @@
...
                 <t:navbox labelKey="Process.outputs">
                     <t:navboxLink icon="fa fa-bars" labelKey="ProcessLog.plural"
                                   url="@apply('/ps/%s', process.getId())"
                                   active="page == 'logs'"/>
 
                     <i:for type="sirius.biz.process.output.ProcessOutput" var="output" items="@process.getOutputs()">
-                        <t:navboxLink icon="@output.getIcon()" label="@output.getLabel()"
-                                      url="@apply('/ps/%s/output/%s', process.getId(), output.getName())"
-                                      active="page == 'output-' + output.getName()"/>
+                        <i:if test="@!output.isSystemOutput() || user().hasPermission('permission-manage-all-processes')">
```
andyHa, line 99: shouldn't we also enforce this check when actually rendering the output (like in the controller)?

andyHa (changes requested):

andyHa (approved):

sabieber (approved):

ymo-sci (approved):

PR #1288, opened 2021-11-02
Adds a LookupTableEntryParameter

idlira:
which allows to define a job parameter listing entries of the given lookup table.

Inspired in the CodeListEntryParameter.

Fixes: OX-7662

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/LookupTableEntryParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/LookupTableEntryParameter.java
@@ -1,0 +1,93 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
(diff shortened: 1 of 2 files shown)


sabieber (approved):

cschierle (approved):

src/main/java/sirius/biz/jobs/params/LookupTableEntryParameter.java
```diff
@@ -0,0 +1,93 @@
...
+        this.defaultValue = defaultValue;
+        return this;
+    }
+
+    /**
+     * Enumerates all entries provided by the lookup table.
+     *
+     * @return the list of entries defined by the lookup table
+     */
+    public List<LookupTableEntry> getValues() {
```
andyHa, line 62: shouldn't we use an auto-complete approach here? we have tables with 1000s of entries...

andyHa:
Whats the benefit over LookupTableParameter ?

andyHa (changes requested):

PR #1290, opened 2021-11-04
Removes annotation to make pre-authorized articles accessible (and not all without an authKey)

ymo-sci:


src/main/java/sirius/biz/tycho/kb/KnowledgeBaseController.java
```diff
@@ -67,6 +67,7 @@ public void article(WebContext webContext, String articleId) {
      * @param lang       the language to render the article in
      * @param articleId  the id or code of the article to render
      */
+    @LoginRequired
```
andyHa, line 70: This is handled in "resolve" and should not be enforced here, as there might as well be "public" articles.

andyHa (changes requested):

mkeckmkeck (approved):

qw3ry (approved):

sabieber (approved):

PR #1294, opened 2021-11-09
Moves vfs-modal scripts into tycho-biz.js + some codestyle

meg1502:


src/main/resources/default/taglib/t/fileField.html.pasta
```diff
@@ -59,6 +59,7 @@
                 }
                 selectVFSFile({
                     path: value,
+                    allowUpload: true,
```
sabieber, line 62: Was this change intended?

src/main/resources/default/assets/scripts/vfs-modals.js
```diff
@@ -180,7 +180,7 @@ function selectVFSFile(config) {
                         })
                     }
                 } else {
-                    _uploadBox.textContent = '';
+                    _uploadBox.style.display = 'none';
```
sabieber, line 183: Why not via bootstrap marker class?

sabieber (changes requested):

andyHa (approved):

src/main/resources/default/taglib/t/fileField.html.pasta
meg1502: yes, this was previously set inside the script itself to always true. 
i would keep this here in case we might need to move it to a tag parameter or set it to false in another tag

sabieber (approved):

src/main/resources/default/taglib/t/fileField.html.pasta
sabieber: Okay, it just didnt match the commit message at all which hinted at a commit gone wrong

idlira (approved):

PR #1297, opened 2021-11-10
Adds an optional file extension filter for file picker modal

meg1502:
In wondergem, this will only limit the files displayed.
In tycho, this will also restrict the upload box to the given file types.
We use a comma seperated list of file extensions, each prefixed with '.' so we can easily feed the list directly into the dropzone library, and we can easily convert the string into a list(stream) on the server.

Fixes: SIRI-469

### DIFF
--- a/src/main/resources/default/taglib/w/filefield.html.pasta
+++ b/src/main/resources/default/taglib/w/filefield.html.pasta
@@ -17,6 +17,8 @@
 <i:arg name="id" type="String" default=""/>
 <i:arg name="placeholder" type="String" default=""/>
 <i:arg name="tabIndex" type="String" default=""/>
+<i:arg type="String" name="allowedExtensions" default=""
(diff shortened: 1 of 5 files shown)


src/main/resources/default/taglib/w/filefield.html.pasta
```diff
@@ -53,7 +55,7 @@
                     value = value.substr(0, value.lastIndexOf("/"))
                 }
 
-                selectVFSFile(value).then(function (selectedValue) {
+                selectVFSFile(value, null, '___allowedExtensions').then(function (selectedValue) {
```
meg1502, line 58: this is a bit messy, but I didn't want to restructure the wondergem methods, in tycho this is cleaner thanks to the config object

idlira (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #1309, opened 2021-11-18
Logs empty files seperated from skipped files when extracting

meg1502:


sabieber (approved):

src/main/java/sirius/biz/storage/layer3/ExtractArchiveJob.java
```diff
@@ -156,7 +157,9 @@ private void handleExtractedFile(ExtractedFile extractedFile,
         Watch watch = Watch.start();
 
         if (extractedFile == null) {
-            process.addTiming(FILE_SKIPPED_MESSAGE, watch.elapsedMillis());
+            // if this happens we don't know the the file name of this entry, we just log with an empty name
```
ymo-sci, line 160:
```suggestion
            // if this happens we don't know the file name of this entry, we just log with an empty name
```

ymo-sci (approved):

mkeckmkeck (approved):

PR #1310, opened 2021-11-18
No longer shows error if the file upload was canceled by user

meg1502:


src/main/resources/default/assets/scripts/vfs-modals.js
```diff
@@ -124,6 +124,7 @@ function selectVFSFile(config) {
                             },
                             sendFileAsBody: true,
                             parallelUploads: 1,
+                            maxFiles: 1,
```
idlira, line 127: What this does? We can now only drop a single file?

sabieber (approved):

fhaScireum (approved):

idlira (approved):

ymo-sci (approved):

mkeckmkeck (approved):

PR #1312, opened 2021-11-22
Handles exception for non-rfc-urls in location header

meg1502:
currently non-rfc-conforming URLs (e.g. containing unescaped spaces) in the location header when redirecting will break the http client and abort with an IllegalArgumentException. We need to find a better solution for this, but temporarily we just catch it here and handle the exception, so we can provide appropiate logs and we don't abort completly.

Fixes: SIRI-479
Screenshot 2021-11-22 at 13 34 17

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
+++ b/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
@@ -1277,6 +1277,7 @@ public abstract class VirtualFile extends Composable implements Comparable<Virtu
      * @throws HandledException in case of any error while downloading the contents
      */
     public boolean loadFromUrl(URI url, FetchFromUrlMode mode) {
+
         Watch watch = Watch.start();
         if (performLoadFromUrl(url, mode)) {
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -1311,6 +1312,18 @@ private boolean performLoadFromUrl(URI url, FetchFromUrlMode mode) {
                             .hint(ProcessLog.HINT_MESSAGE_KEY, MESSAGE_KEY_LOAD_FROM_URL_FAILED)
                             .hint(ProcessLog.HINT_MESSAGE_COUNT, ProcessLog.MESSAGE_TYPE_COUNT_MEDIUM)
                             .handle();
+        } catch (IllegalArgumentException illegalArgumentException) {
```
qw3ry, line 1315: Not sure if we should "blindly" catch all IllegalArgumentExceptions and just assume that its an error in the location header. But as a quick fix with TODO I'm fine with it.

qw3ry (approved):

idlira (approved):

PR #1314, opened 2021-11-23
Fixes the w:blobSoftRefField taglib

jmuscireum:
The extensions parameter needs to be specified even if the files should not be filtered based on the extension.

### DIFF
--- a/src/main/resources/default/assets/javascript/blobreffield.js
+++ b/src/main/resources/default/assets/javascript/blobreffield.js
@@ -111,7 +111,7 @@ function initBlobSoftRefField(element, blobKeyField, blobStorageSpace, originalU
 
         const currentPath = element.dataset.path || blobStorageSpacePath;
 
-        selectVFSFile(currentPath, blobStorageSpacePath).then(function (selectedValue) {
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/javascript/blobreffield.js
```diff
@@ -111,7 +111,7 @@ function initBlobSoftRefField(element, blobKeyField, blobStorageSpace, originalU
 
         const currentPath = element.dataset.path || blobStorageSpacePath;
 
-        selectVFSFile(currentPath, blobStorageSpacePath).then(function (selectedValue) {
+        selectVFSFile(currentPath, blobStorageSpacePath, '').then(function (selectedValue) {
```
andyHa, line 114: do we need the same fix in tycho?

jmuscireum: no, the method is already called correctly everywhere else

mkeckmkeck (approved):

ymo-sci (approved):

qw3ry (approved):

meg1502 (approved):

sabieber (approved):

PR #1319, opened 2021-11-30
Normalizes VFS to Unicode NFC Form

meg1502:
Fixes SIRI-491

Normalizes to Unicode NFC Form when sanitizing a path. 

This effectively means that all lookups and therefore each new blob creation is normalized to the NFC form. This also means that existing NFD forms an no longer be found, even when querying for exactly the NFD form (as it will get normalized). Therefore all existing DB objects (MongoDirectory,Mongoblob, SQlBlob...) need to be normalized to NFC. An appropriate job for this task will be provided in the next commit.

Additionally sanitizes the filename in beforeSave of vfs db objects 

This makes sure that the data in the db is consistent with the actual names we search for, even if the file or directory was not created by findByPath.
This also means that in order for existing objects to be normalized we can use a simple resave entities job.

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -51,6 +51,7 @@ import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.text.Normalizer;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


sabieber (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -125,6 +126,9 @@
      * <p>
      * Setting this to true makes the file system effectively case-insensitive. Whereas using false
      * makes it case-sensitive.
+     * Note that this only determines case sensitivity:
```
andyHa, line 129: Nice comment / clarification +1

andyHa (approved):

mkeckmkeck (approved):

PR #1322, opened 2021-12-07
Changes visibility of DictionaryBasedImport to public

cschierle:
so it can be directly accessed by custom import jobs.

Fixes: SIRI-499


sabieber (approved):

idlira (approved):

cschierle:
Is needed for now, but might be reverted, if a different solution is found

PR #1325, opened 2021-12-13
VirtualFile new methods: root + isRootWorkingDirectory

idlira:
VirtualFile.root() - determines the root directory of a VirtualFile, handling when the path is resolved from a known space.
```
/work/foo/bar -> /work
/foo/bar -> /
```

VirtualFile.isWootWorkDirectory():
```
/work/foo/bar -> true
/foo/bar -> false
```

Fixes: SIRI-502

### DIFF
--- a/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
+++ b/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
@@ -124,6 +124,19 @@ public class ObjectStorage {
     }
 
     /**
+     * Determines if the storage space with the work-directory space.
+     *
+     * @param space the name of the space to check
+     * @return <tt>true</tt> if the space is known, <tt>false</tt> otherwise
+     */
+    public boolean isWorkDirectory(String space) {
+        if (!isKnown(space)) {
+            return false;
+        }
+        return WORK_SPACE_NAME.equals(getSpaceMap().get(space).getName());
+    }
+
+    /**
      * Returns all known layer 1 spaces.
      *
      * @return all known storage spaces of the layer 1
--- a/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
+++ b/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
@@ -52,6 +52,10 @@ public class ObjectStorage {
      * to use for a space.
      */
     public static final String CONFIG_KEY_LAYER1_CIPHER = "cipher";
+    /**
+     * Contains the name of the storage space used for "work files".
+     */
+    public static final String WORK_SPACE_NAME = "work";
 
     @Part
     private StorageUtils utils;
--- a/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
+++ b/src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
@@ -19,6 +19,7 @@ import sirius.kernel.settings.Extension;
 
 import javax.annotation.Nullable;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
@@ -77,7 +78,7 @@ public class ObjectStorage {
             spaceMap = initializeSpaceMap();
         }
 
-        return spaceMap;
+        return Collections.unmodifiableMap(spaceMap);
     }
 
     private Map<String, ObjectStorageSpace> initializeSpaceMap() {
--- a/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
+++ b/src/main/java/sirius/biz/storage/layer3/VirtualFile.java
@@ -14,6 +14,7 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import sirius.biz.process.logs.ProcessLog;
 import sirius.biz.storage.layer1.FileHandle;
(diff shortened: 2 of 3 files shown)


src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
```diff
@@ -123,6 +123,19 @@ public boolean isKnown(String space) {
         return getSpaceMap().containsKey(space);
     }
 
+    /**
+     * Determines if the storage space with the work-directory space.
+     *
+     * @param space the name of the space to check
+     * @return <tt>true</tt> if the space is known, <tt>false</tt> otherwise
+     */
+    public boolean isWorkDirectory(String space) {
```
sabieber, line 133: The javadoc is kind of confusing, is this a copy+paste leftover?

sabieber (approved):

src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
```diff
@@ -52,6 +53,10 @@
      * to use for a space.
      */
     public static final String CONFIG_KEY_LAYER1_CIPHER = "cipher";
+    /**
+     * Contains the name of the storage space used for "work files".
+     */
+    public static final String WORK_SPACE_NAME = "work";
```
andyHa, line 59: I'd say this either belongs to layer2 or 3 or we move this into some class in utils?

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -247,6 +251,37 @@ public String relativePath(VirtualFile rootParent) {
...
+    /**
+     * Returns the root directory path of the current virtual file.
+     * <p>
+     * If the virtual file belongs to a known space, it is used as root, otherwise <tt>/</tt> is returned.
+     *
+     * @return the root directory path
+     */
+    @Nullable
+    public VirtualFile root() {
+        List<VirtualFile> fileList = pathList();
```
meg1502, line 263: why not call the variable pathList ?

src/main/java/sirius/biz/storage/layer1/ObjectStorage.java
```diff
@@ -119,6 +124,19 @@ public boolean isKnown(String space) {
         return getSpaceMap().containsKey(space);
     }
 
+    /**
+     * Determines if the storage space with the work-directory space.
+     *
+     * @param space the name of the space to check
+     * @return <tt>true</tt> if the space is known, <tt>false</tt> otherwise
+     */
+    public boolean isWorkDirectory(String space) {
```
andyHa, line 133: s.a.

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -247,6 +251,37 @@ public String relativePath(VirtualFile rootParent) {
...
+        if (fileList.isEmpty()) {
+            return null;
+        }
```
meg1502, line 266: Will this ever happen? If yes, a note when null is returned should be included in the javadoc, if not, remove it :)

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -247,6 +251,37 @@ public String relativePath(VirtualFile rootParent) {
...
     }
 
+    /**
+     * Returns the root directory path of the current virtual file.
+     * <p>
+     * If the virtual file belongs to a known space, it is used as root, otherwise <tt>/</tt> is returned.
+     *
+     * @return the root directory path
+     */
+    @Nullable
```
meg1502, line 261: Add to javadoc when null is expected to be returned

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -247,6 +251,37 @@ public String relativePath(VirtualFile rootParent) {
...
+        return fileList.get(0);
+    }
+
+    /**
+     * Indicates if the root directory is considered the working directory.
+     *
+     * @return <tt>true</tt> when the root is the working directory, <tt>false</tt> otherwise
+     * @see ObjectStorage#WORK_SPACE_NAME
+     */
+    public boolean isRootWorkDirectory() {
```
andyHa, line 281: the VirtualFile already has quite a borad API- Shouldn't we move this either into L3Uplink or into a helper in the utils package? We then could also check if this is a wrapper as generated by L3Uplink and extract the space name there - and also provide a constant / another utility method for work...

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -247,6 +251,37 @@ public String relativePath(VirtualFile rootParent) {
...
+        List<VirtualFile> fileList = pathList();
+        if (fileList.isEmpty()) {
+            return null;
+        }
+
+        if (fileList.size() > 1 && objectStorage.isKnown(fileList.get(1).name())) {
+            return fileList.get(1);
+        }
+
+        return fileList.get(0);
```
meg1502, line 272: I'd preface this block of code with a small inline comment, something like
`// index 0 is always '/' -  and known spaces will be at index 1`

just to clarify the structure of the list

meg1502 (changes requested):

PR #1328, opened 2021-12-13
Marries VirtualFile with BlobStorageSpace

idlira:
A new config key allows to mark a storage space as a "work directory", intended for files supposed to have a short life (eg. deleted after processed by a job).

Respectively, this information is made public with `BlobStorageSpace.isWorkDirectory()`

Finally, VirtualFiles initialized from a storage space via _L3UpLink_ will contain the information about their originating storage space. This will work for real files, sitting under a Directory hierarchy, but also placeholders. Therefore:

Both these paths would return the storage space `work`:
* /work/real-directory/real-file and
* /work/foo/bar 

Sample call to verify is a virtual file belongs to a working directory:
``` java
virtualFile.getStorageSpace()
           .filter(BlobStorageSpace::isWorkDirectory)
           .ifPresent(storageSpace -> ...)
```

Fixes: SIRI-502

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
+++ b/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
@@ -117,10 +117,10 @@ public class L3Uplink implements VFSRoot {
          * Creates a placeholder for the given parent and name.
          *
          * @param parentDirectory the directory which (if present) is used to determine if the underlying storage space
-         *                        is readonly. In this case an empty optional is returned
+         *                        is readonly. In this case null is returned
          * @param parent          the parent to pass on
          * @param name            the name of the placeholder
-         * @return a placeholder representing a non existent file or directory
+         * @return a placeholder representing a non-existent file or directory
          */
         @Nullable
         protected VirtualFile createPlaceholder(@Nullable Directory parentDirectory, VirtualFile parent, String name) {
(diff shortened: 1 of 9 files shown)


src/main/java/sirius/biz/storage/layer2/L3Uplink.java
```diff
@@ -132,6 +132,14 @@ protected VirtualFile createPlaceholder(@Nullable Directory parentDirectory, Vir
             file.attach(new Placeholder(parent, name));
```
andyHa, line 132: file.attach(parentDirectory.getStorageSpace());

and also attach the StorageSpace in wrapDirectory and wrapBlob

andyHa: and then add a method either in L3Uplink or VirtualFileSystem or StorageUtils (for better visibitilty) 
isWorkFile(VirtualFile file) { return WORK.equals(file.tryAs(StorageSpace.class).map(StorageSpace::getName).orElse("")); }

andyHa: Or, instead of checking against the name, we could add a method to BlobStorageSpace "isAutoCleanup" which checks for retentionDays > 0 which would be true for work and tmp but not for others...

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer2/L3Uplink.java
idlira: Closing again, since lots will be reverted making reviews more difficult.

PR #1329, opened 2021-12-14
Marries VirtualFiles with BlobStorageSpace - Take 3

idlira:
Now using Composables to attach the blob storage space to virtual files.
Existing files are handled in wrapBlob and wrapDirectory.
Non-existing files inherit the information from their parent (if exists).

In addition, we also detect if a virtual file is a work file using the `retentionDays` config, which controls how many days files are visible before being deleted automatically by the system.

Replaces https://github.com/scireum/sirius-biz/pull/1328

Fixes: SIRI-502

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
+++ b/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
@@ -227,6 +227,7 @@ public class L3Uplink implements VFSRoot {
     public MutableVirtualFile wrapDirectory(VirtualFile parent, Directory directory, boolean forceReadonly) {
         MutableVirtualFile file = MutableVirtualFile.checkedCreate(parent, directory.getName());
         file.attach(Directory.class, directory);
+        file.attach(directory.getStorageSpace());
         attachHandlers(file, forceReadonly);
 
         return file;
@@ -245,6 +246,7 @@ public class L3Uplink implements VFSRoot {
     public MutableVirtualFile wrapBlob(VirtualFile parent, Blob blob, boolean forceReadonly) {
         MutableVirtualFile file = MutableVirtualFile.checkedCreate(parent, blob.getFilename());
         file.attach(Blob.class, blob);
+        file.attach(blob.getStorageSpace());
         attachHandlers(file, forceReadonly);
 
         return file;
--- a/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
+++ b/src/main/java/sirius/biz/storage/layer2/L3Uplink.java
@@ -15,6 +15,7 @@ import sirius.biz.storage.layer3.FileSearch;
 import sirius.biz.storage.layer3.MutableVirtualFile;
 import sirius.biz.storage.layer3.VFSRoot;
 import sirius.biz.storage.layer3.VirtualFile;
+import sirius.biz.storage.layer3.VirtualFileSystem;
 import sirius.biz.storage.util.StorageUtils;
 import sirius.kernel.commons.Files;
 import sirius.kernel.commons.Strings;
@@ -45,6 +46,9 @@ public class L3Uplink implements VFSRoot {
 
     @Part
     private BlobStorage storage;
+
+    @Part
+    private VirtualFileSystem vfs;
 
     /**
      * Represents a non-existent file or directory which might be created by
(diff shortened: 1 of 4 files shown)


sabieber (approved):

jmuscireum (approved):

cschierle (approved):

src/main/java/sirius/biz/storage/layer2/L3Uplink.java
```diff
@@ -238,13 +244,14 @@ public MutableVirtualFile wrapDirectory(VirtualFile parent, Directory directory,
...
      * @param forceReadonly determines if the blob should be forcefully set to readonly (independent of
-     *                      {@link BlobStorageSpace#isReadonly()}. Note that setting this to <tt>false</tt> will not
+     *                      {@link BlobStorageSpace#isReadonly()}). Note that setting this to <tt>false</tt> will not
      *                      make the blob writable if <tt>BlobStorageSpace#isReadonly()</tt> returns <tt>true</tt>
      * @return the resulting mutable virtual file
      */
     public MutableVirtualFile wrapBlob(VirtualFile parent, Blob blob, boolean forceReadonly) {
         MutableVirtualFile file = MutableVirtualFile.checkedCreate(parent, blob.getFilename());
         file.attach(Blob.class, blob);
+        file.attach(blob.getStorageSpace());
```
andyHa, line 254: Attach with BlobStorageSpace.class as first parameter...

src/main/java/sirius/biz/storage/layer3/VirtualFileSystem.java
```diff
@@ -120,4 +124,37 @@ public String makePath(String... parts) {
...
+
+    /**
+     * Returns the storage space associated with the given file.
+     *
+     * @param virtualFile the virtual file to obtain the storage space
+     * @return the {@link BasicBlobStorageSpace} sitting on top of the virtual file
+     */
+    public Optional<? extends BasicBlobStorageSpace<?, ?, ?>> fetchBlobStorageSpace(VirtualFile virtualFile) {
+        Optional<? extends BasicBlobStorageSpace<?, ?, ?>> blobStorageSpace =
+                virtualFile.tryAs(MongoBlobStorageSpace.class);
```
andyHa, line 136: Only use BlobStorageSpace here

andyHa: Do we need this as api anyway, this is just a boilerplate for another single method call then...

src/main/java/sirius/biz/storage/layer3/VirtualFileSystem.java
```diff
@@ -120,4 +124,37 @@ public String makePath(String... parts) {
...
+        return blobStorageSpace;
+    }
+
+    /**
+     * Checks if the file belongs to a work directory, dictated by storage spaces with retention days greater than zero.
+     *
+     * @param virtualFile the virtual file to check
+     * @return <tt>true</tt> when the owner space specifies retention days greater than zero, <tt>false</tt> otherwise
+     */
+    public boolean isWorkFile(VirtualFile virtualFile) {
```
andyHa, line 151: I'd use another name here, as this might imply just "/work". Maybe isTransientFile or isAutoCleanupBlob or the like.

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer3/VirtualFileSystem.java
idlira: The name was your suggestion in the previous PR. Anyways, updated to isAutoCleanupBlob.

src/main/java/sirius/biz/storage/layer2/L3Uplink.java
```diff
@@ -130,6 +134,8 @@ protected VirtualFile createPlaceholder(@Nullable Directory parentDirectory, Vir
 
             MutableVirtualFile file = MutableVirtualFile.checkedCreate(parent, name);
             file.attach(new Placeholder(parent, name));
+            parent.tryAs(BlobStorageSpace.class)
+                  .ifPresent(blobStorageSpace -> file.attach(BlobStorageSpace.class, blobStorageSpace));
```
andyHa, line 138:
```suggestion
            file.attach(BlobStorageSpace.class, parentDirectory.getStorageSpace());
```

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer2/L3Uplink.java
idlira: Won't work because in this path: `/work/existing-folder/foo/bar`
when we `bar`, parentDirectory will be null since its parent is a PlaceHolder. This is why we always inherit from the parent. The first placeholder gets it from a real Directory, all others down the chain shall get it.

PR #1342, opened 2022-01-26
Makes blobImageHardRefField a true "field" for Tycho.

andyHa:


src/main/java/sirius/biz/mongo/SortField.java
```diff
@@ -60,6 +60,10 @@ public SortField(MongoEntity owner) {
      * @return the normalized text in lower-case
      */
     public static String normalizeText(String text) {
+        if (Strings.isEmpty(text)) {
+            return null;
```
jakobvogel, line 64: Can this null value lead to problems further down the road?

jakobvogel (approved):

src/main/java/sirius/biz/mongo/SortField.java
meg1502, line 64:
```suggestion
            return text;
```

i would suggest to just return the input if its null/empty and not change an empty string to null

meg1502 (changes requested):

src/main/java/sirius/biz/mongo/SortField.java
andyHa: The underlying sort field would treat both the same so I commonly like normalizers to convert both to null to have a more predictable output. However, it can be changed here to make the PR less impactful.

andyHa: The method crashed before, so it wasn't a valid input anyway, now it is and as a defined result.

mkeckmkeck (approved):

sabieber (approved):

meg1502 (approved):

cschierle (approved):

ymo-sci (approved):

PR #1343, opened 2022-01-26
ErrorContext: permits to only log messages under the current context

idlira:
Exposes methods to only log an exception as either error or warning (haven't seen a case for info yet, can add as the need arise).

The current logic handles errors thrown nicely logging them under the current context, but this is not sufficient.
Many times we need to log a warning, respecting the current context but carry on with processing.
This way we expose 2 methods which can be used to simply log the HandledException given.
This way we also profit from the hints attached to the message, especially the ones used to limit the number of incidences to be logged.

Fixes: OX-8105

### DIFF
--- a/src/main/java/sirius/biz/process/ErrorContext.java
+++ b/src/main/java/sirius/biz/process/ErrorContext.java
@@ -9,6 +9,7 @@
 package sirius.biz.process;
 
 import sirius.biz.process.logs.ProcessLog;
+import sirius.biz.process.logs.ProcessLogType;
 import sirius.kernel.async.CallContext;
 import sirius.kernel.async.SubContext;
 import sirius.kernel.async.TaskContext;
@@ -165,9 +166,9 @@ public class ErrorContext implements SubContext {
     private void logException(HandledException exception, ProcessLogType processLogType) {
         TaskContext taskContext = TaskContext.get();
         if (taskContext.getAdapter() instanceof ProcessContext processContext) {
-            processContext.log(ProcessLog.error()
-                                         .withHandledException(exception)
-                                         .withMessage(enhanceMessage(exception.getMessage())));
+            processContext.log(new ProcessLog().withType(processLogType)
+                                               .withHandledException(exception)
+                                               .withMessage(enhanceMessage(exception.getMessage())));
         } else {
             taskContext.log(enhanceMessage(exception.getMessage()));
         }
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/process/ErrorContext.java
```diff
@@ -258,10 +283,11 @@ public void handleInContext(UnaryOperator<String> failureDescription, String lab
         try {
             return Optional.ofNullable(producer.create());
         } catch (HandledException exception) {
-            logException(exception);
+            logException(exception, ProcessLogType.ERROR);
```
jakobvogel, line 286: Use new `logError()`?
```suggestion
            logError(exception);
```

src/main/java/sirius/biz/process/ErrorContext.java
```diff
@@ -258,10 +283,11 @@ public void handleInContext(UnaryOperator<String> failureDescription, String lab
...
         } catch (Exception exception) {
             String message = exception.getMessage() + " (" + exception.getClass().getName() + ")";
-            logException(Exceptions.handle().to(Log.BACKGROUND).error(exception).withDirectMessage(message).handle());
+            logException(Exceptions.handle().to(Log.BACKGROUND).error(exception).withDirectMessage(message).handle(),
+                         ProcessLogType.ERROR);
```
jakobvogel, line 290: Use new `logError()`?
```suggestion
            logError(Exceptions.handle().to(Log.BACKGROUND).error(exception).withDirectMessage(message).handle());
```

jakobvogel (approved):

src/main/java/sirius/biz/process/ErrorContext.java
idlira: logException is correct, since we receive a HandledException for logging. Error and Warning is a concept for the ProcessLog, so I changed the API methods to logExceptionAsError and logExceptionAsWarning since it better represents what they do.

src/main/java/sirius/biz/process/ErrorContext.java
```diff
@@ -138,12 +139,36 @@ public String getContextAsString() {
...
+     * of entries logged for a specific type.
+     *
+     * @param exception the exception to log
+     */
+    public void logExceptionAsError(HandledException exception) {
+        logException(exception, ProcessLogType.ERROR);
+    }
+
+    /**
+     * Logs an {@link ProcessLogType#WARNING warning} including the current context.
```
ymo-sci, line 155:
```suggestion
     * Logs a {@link ProcessLogType#WARNING warning} including the current context.
```

ymo-sci (approved):

mkeckmkeck (approved):

sabieber (approved):

andyHa (approved):

PR #1345, opened 2022-01-26
ImportFileJob: new Parent Directory(ies) parameter

idlira:
Permits the user to define a parent directory (or more if split by `/`) as the destination to import auxiliary files.

Example:
Archive Contents:
```
file1
dir1/file2
foo/bar/file3
```

Parent Directory Parameter: `/foo/bar`

Destination -> ${root} is the main storage root where files are loaded, like /work, /attachments etc.:
```
${root}/foo/bar/file1
${root}/foo/bar/dir1/file2
${root}/foo/bar/file3
```

Fixes: OX-7899


src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -287,9 +300,17 @@ private String getTargetPath(ExtractedFile extractedFile) {
...
-            return targetPath;
         }
+
+        // Drops the parent directory from the path if we match it with the beginning:
+        // parent -> /foo/bar
+        // path obtained from archive -> /foo/bar/zoo/file
+        // result -> /zoo/file
+        if (targetPath.toLowerCase().startsWith(auxFilesParentDirectory.toLowerCase())) {
+            return targetPath.substring(auxFilesParentDirectory.length());
+        }
```
jakobvogel, line 311: Not sure whether I get the purpose of these lines ... if files in the archive already contain the "prefix", it will be dropped, causing such files to be moved up (relatively)?

If this is the case, the unpacked file structure will be different from the packed one? Wouldn't that be too much magic?

jakobvogel (approved):

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
idlira: Is a _desired magic_ which tries to cover a common user failure: the user wants its files to be unpacked in a parent folder, but the files inside the archive are already structured under the folder. By not doing it and simply exploding the archive in the target would yield `${root}/foo/bar/foo/bar/file3` in the example of the PR description.

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -58,22 +60,30 @@
      */
     public static final Parameter<AuxiliaryFileMode> AUX_FILE_MODE_PARAMETER = new EnumParameter<>("auxFileMode",
                                                                                                    "$FileImportJobFactory.auxFileMode",
-                                                                                                   AuxiliaryFileMode.class)
-            .withDefault(AuxiliaryFileMode.UPDATE_ON_CHANGE)
-            .markRequired()
-            .build();
+                                                                                                   AuxiliaryFileMode.class).withDefault(
+            AuxiliaryFileMode.UPDATE_ON_CHANGE).markRequired().build();
```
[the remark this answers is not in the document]
idlira: I'm afraid so. The declaration of the variable already takes 100 chars.
`public static final Parameter<AuxiliaryFileMode> AUX_FILE_MODE_PARAMETER = new EnumParameter<>(`
I guess only if increase the column where lines are wrapped would improve this...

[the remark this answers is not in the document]
sabieber: I dont really think that looks like our codestyle settings. Could you please run a `mvn compile` and reopen the project? Also ensure the codestyle settings are set to project default

src/main/resources/biz_en.properties
```diff
@@ -285,6 +285,8 @@ FileImportJob.requiredFileNotFound = The file '${filename}' cannot be found in t
 FileImportJobFactory.auxFileFlattenDirectoriesParameter = Ignore the folder structure of the additional files
 FileImportJobFactory.auxFileFlattenDirectoriesParameter.help = Specifies whether the folder structure in the archive should be ignored. If set, all additional files are stored in the target directory without creating sub-folders.
 FileImportJobFactory.auxFileMode = Additional files
+FileImportJobFactory.auxFileParentDirectory = Basic path
```
sabieber, line 288: I think `Base path` would be a more appropriate wording

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
sabieber: You could add to the comment above the code that the removed base path is later added again

sabieber (changes requested):

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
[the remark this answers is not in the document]
sabieber: Toni applied the correct formatting setting, so nothing to do here for now

sabieber (approved):

src/main/resources/biz_de.properties
```diff
@@ -285,6 +285,8 @@ FileImportJob.requiredFileNotFound = Die Datei '${filename}' wurde im Archiv nic
 FileImportJobFactory.auxFileFlattenDirectoriesParameter = Ordnerstruktur der zusätzlichen Dateien ignorieren
 FileImportJobFactory.auxFileFlattenDirectoriesParameter.help = Gibt an, ob die Ordnerstruktur im Archiv ignoriert werden soll. Wenn gesetzt, werden alle zusätzliche Dateien in das Zielverzeichnis abgelegt ohne Unterordner anzulegen.
 FileImportJobFactory.auxFileMode = Zusätzliche Dateien
+FileImportJobFactory.auxFileParentDirectory = Basispfad
+FileImportJobFactory.auxFileParentDirectory.help = Definiert einen Basispfad (übergeordnetes Verzeichnis), der im Zielverzeichnis verwendet oder erstellt wird, um die zusätzliche Dateien hochzuladen. Verwenden Sie '/', um mehrere Verzeichnisse anzugeben.
```
jakobvogel, line 289: Not sure about "mehrere Verzeichnisse" ... sounds like multiple independent directories existing "side by side", while a hierarchy of nested directories is meant

idlira: Suggestions are welcome! 😇

jakobvogel: What about something like the following? Depends a little on the audience, of course.
```suggestion
FileImportJobFactory.auxFileParentDirectory.help = Definiert einen Ordner innerhalb des Zielverzeichnisses, in den die neuen Dateien hochgeladen werden. Verwenden Sie '/' für einen Pfad mit Unterverzeichnissen. Der Ordner wird nötigenfalls neu angelegt.
```

sabieber (approved):

andyHa (approved):

jmuscireum (approved):

ymo-sci (approved):

PR #1360, opened 2022-02-15
Does not delete source directory when transferring content.

fhaScireum:
This is done as the source directory should probably not (or even can't) be deleted and deleting the source directory at the source location can be accomplished via TransferMode#MOVE.

- Fixes: SE-11421

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/Transfer.java
+++ b/src/main/java/sirius/biz/storage/layer3/Transfer.java
@@ -331,6 +331,8 @@ public class Transfer {
                 destination.createAsDirectory();
             }
 
+            // Transfer directory and delete all files and directories including the source directory at the source
+            // location if the action is a move.
             transferDirectory(source, destination.findChild(source.name()), delete, delete);
         }
     }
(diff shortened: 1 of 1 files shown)


mkeckmkeck (approved):

andyHa (approved):

ymo-sci (approved):

src/main/java/sirius/biz/storage/layer3/Transfer.java
```diff
@@ -331,26 +331,29 @@ private void transfer(boolean delete) {
...
             destinationDirectory.createAsDirectory();
         }
 
         sourceDirectory.allChildren().stream().forEach(child -> {
             if (child.isFile()) {
-                transferFileTo(child, destinationDirectory.findChild(child.name()), delete);
+                transferFileTo(child, destinationDirectory.findChild(child.name()), deleteContent);
             } else {
-                transferDirectory(child, destinationDirectory.findChild(child.name()), delete);
+                transferDirectory(child, destinationDirectory.findChild(child.name()), deleteContent, deleteContent);
```
sabieber, line 350: Even if thats correct this looks like an error, so maybe explain why the same param is passed twice here :)

src/main/java/sirius/biz/storage/layer3/Transfer.java
```diff
@@ -331,26 +331,29 @@ private void transfer(boolean delete) {
                 destination.createAsDirectory();
             }
 
-            transferDirectory(source, destination.findChild(source.name()), delete);
+            transferDirectory(source, destination.findChild(source.name()), delete, delete);
```
sabieber, line 334: Even if thats correct this looks like an error, so maybe explain why the same param is passed twice here :)

sabieber (changes requested):

sabieber (approved):

PR #1365, opened 2022-02-18
Improves Query Tool Functionality and moves to Tycho

sabieber:
This migrates the templates of the query tool to Tycho and tries to increase the usefulness of the Query Tool even more by enriching the rendering of some special properties.

Feel free to suggest other properties that could be improved or use this PR as inspiration to enhance the rendering even more :)

# Improved Properties

## 1) Enum

It now also displays the actual `name()` of the enum value next to the translated `toString` representation as a badge.

image

## 2) Entity Ref

It now includes an icon that is clickable instead of making the whole value/ID clickable. This way it should be easier to copy the value.

image

## 3) Entity Ref List

It now renders each reference included in the list similar to the `Entity Ref` including an icon to jump to the referenced entity in the Query Tool.

image

## 4) Maps / Multi Language Strings

It now renders them as small sub-tables including Key and Value columns which should increase readability quite a bit.

image

## 5) Dates / Numbers / Booleans

It now renders all properties left-aligned, instead of rendering some of them right-aligned (which is quite common in table use-cases). This should increase the readability/scanability of the entity as for entities with long fields the values of such properties where far off the actual field name.

image

## 6) Nested Lists

This was the whole reason for the PR which got a bit out of hand :D
The query tool now renders the entries of nested lists as small sub-tables which can be shown and by doing so displays all properties of the nested entity.
Because this could quite increase the size of an entity shown by the Query Tool those sub-tables are collapsed by default and can be expanded by a button press.

image

image

Currently not yet mergeable as its waits for https://github.com/scireum/sirius-db/pull/491 but can already be looked at. Feedback welcome :)


src/main/resources/default/templates/biz/model/query-results.html.pasta
```diff
@@ -4,61 +4,97 @@
...
+                                                                <i:else>
+                                                                    <i:if test="property.is(sirius.db.mixing.properties.EnumProperty.class)">
+                                                                        <i:local name="enumValue"
+                                                                                 value="@property.getValue(entity)"/>
+
+                                                                        @toUserString(enumValue)
+                                                                        <span class="border-sirius-gray tycho-tag d-inline-block">
+                                                                    @enumValue.as(java.lang.Enum.class).name()
+                                                                </span>
+
```
andyHa, line 78:
```suggestion
                                                                      <i:local name="enumValue" value="@property.getValue(entity)"/>
                                                                      @toUserString(enumValue)
                                                                      <span class="pl-2 border-sirius-gray tycho-tag d-inline-block">
                                                                         @enumValue.as(java.lang.Enum.class).name()
                                                                      </span>

```

andyHa: This tries to fix the code layout, nothing else...

meg1502 (approved):

src/main/resources/default/templates/biz/model/query-results.html.pasta
andyHa: Just a thought - instead of always using gray labels, we could use an array of colors and the ordinal() of the enum sonstant and index into this array - this would always render the first enum green, the second blue etc etc.

Maybe we have such an array already somewhere (??)  - could also be added in another PR...

src/main/resources/default/templates/biz/model/query-results.html.pasta
```diff
@@ -4,61 +4,97 @@
...
+                                                                        <i:else>
+                                                                            @toUserString(property.getValue(entity))
+                                                                        </i:else>
+                                                                    </i:if>
+                                                                </i:else>
+                                                            </i:if>
+                                                        </i:else>
+                                                    </i:if>
```
andyHa, line 86: 😱 :-P

src/main/resources/default/templates/biz/model/query-ref-list-property.html.pasta
```diff
@@ -0,0 +1,18 @@
...
+<i:if test="references.size() > 0">
+    <ol class="mb-0 pl-3">
+        <i:for type="String" var="reference" items="references">
+            <li>
+                <u>@reference</u>
+                <a href="/system/query?class=@property.getReferencedDescriptor().getName()&query=id:@reference"
+                   target="_blank">
+                    <i class="fa fa-external-link-alt"></i>
+                </a>
+            </li>
```
andyHa, line 15:
```suggestion
            <li>
                <a href="/system/query?class=@property.getReferencedDescriptor().getName()&query=id:@reference"
                   target="_blank">
                    <i class="fa fa-external-link-alt"></i>
                    @reference
                </a>
            </li>
```

src/main/resources/default/templates/biz/model/query-results.html.pasta
```diff
@@ -4,61 +4,97 @@
...
-                                 || property.is(sirius.db.mixing.properties.LongProperty.class)
-                                 || property.is(sirius.db.mixing.properties.BooleanProperty.class)">
-                                <div class="align-right">@toUserString(property.getValue(entity))</div>
-                                <i:else>
+    <t:emptyCheck data="@entities">
+        <t:datacards size="x-large">
+            <i:for type="sirius.db.mixing.BaseEntity" var="entity" items="entities">
+                <t:datacard>
+                    <table class="table table-bordered table-hover text-small">
+                        <thead class="table-dark">
```
andyHa, line 21: Although this looks nice, this is "inconsistent" to other tables - should we strive for consistency here? We could also keep this here and the come to a general decision in a follow-up step...

cschierle: consistency sounds good to me 👍

cschierle (approved):

mkeckmkeck (approved):

src/main/resources/default/templates/biz/model/query-results.html.pasta
sabieber: Didn't know a better way to do this in Tagliatelle as the switch block cant have that complex expressions, right?

So the only other way I see would be to introduce a method somewhere in the code that outputs the correct Query Tool property template for each property, which is then passed into a single invoke call. But this seems kinda overkill at the moment.

src/main/resources/default/templates/biz/model/query-ref-list-property.html.pasta
sabieber: The underlined text was a conscious design decision to highlight that these are not normal strings/IDs but references to other entities. But I can change it if wanted :)

src/main/resources/default/templates/biz/model/query-results.html.pasta
sabieber: I dont know if an internal debugging tool has to be consistent to the UI the users see or if usability should be more important in that case. I am open for other suggestions but if removed completely it is very hard so see when a new entity starts during scrolling as its just white, white, white :)

src/main/resources/default/templates/biz/model/query-results.html.pasta
```diff
@@ -4,61 +4,100 @@
...
-                                 || property.is(sirius.db.mixing.properties.IntegerProperty.class)
-                                 || property.is(sirius.db.mixing.properties.LongProperty.class)
-                                 || property.is(sirius.db.mixing.properties.BooleanProperty.class)">
-                                <div class="align-right">@toUserString(property.getValue(entity))</div>
-                                <i:else>
+    <t:emptyCheck data="@entities">
+        <t:datacards size="x-large">
+            <i:for type="sirius.db.mixing.BaseEntity" var="entity" items="entities">
+                <t:datacard title="@entity.getIdAsString()" subTitle="@entity.getClass().getName()">
+                    <table class="table table-hover">
```
sabieber, line 20: In my opinion the `text-small` improved the usability of the Query Tool quite a bit as more properties / information is shown at once.

jakobvogel (approved):

PR #1367, opened 2022-02-21
Improves /system/scripting to directly report into a Process(Context).

andyHa:
One now can start and use a process via:

```
part(Tenants.class).runAsAdminProcess('My Process', | ps | {
    log('Hello From the Other Side');
});
```

### DIFF
--- a/src/main/java/sirius/biz/process/output/TableOutput.java
+++ b/src/main/java/sirius/biz/process/output/TableOutput.java
@@ -139,7 +139,8 @@ public class TableOutput {
 
         /**
          * Adds a plain value to a row.
-         * @param data the plain value to add.
+         *
+         * @param data the plain value to add
          * @return the row builder itself for fluent method calls
          */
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/biz/process/output/TableOutput.java
```diff
@@ -136,6 +137,15 @@ public RowBuilder withCell(Cell cell) {
             return this;
         }
 
+        /**
+         * Adds a plain value to a row.
+         * @param data the plain value to add.
+         * @return the row builder itself for fluent method calls
+         */
```
jakobvogel, line 144: The following would probably be closer to the standards used elsewhere in this file.

```suggestion
        /**
         * Adds a plain value to a row.
         *
         * @param data the plain value to add
         * @return the row builder itself for fluent method calls
         */
```

jakobvogel (approved):

fhaScireum (approved):

sabieber (approved):

PR #1370, opened 2022-02-23
No longer logs conversion failures to storage log

meg1502:
As they are already logged in tenant standby processes and the ConversionEvent


sabieber (approved):

jakobvogel (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1616,9 +1616,8 @@ private void invokeConversionPipelineAsync(B blob, V variant) {
             eventRecorder.record(new BlobConversionEvent().withConversionProcess(conversionProcess)
                                                           .withConversionError(conversionException));
 
-            throw Exceptions.handle()
```
andyHa, line 1619: shoudln't the ConversionEngine make this decision? Like throw a HandledException to only abort, but NPEs stil should end up in the syslog?

sabieber (approved):

andyHa (approved):

PR #1372, opened 2022-03-03
Upgrades clickhouse to version 0.2.6.

cschierle:
Note: The default clickhouse user is now called "default" instead of being empty.

Fixes: SIRI-539

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>ru.yandex.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.1.40</version>
+            <version>0.2.6</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


sabieber (approved):

idlira (approved):

pom.xml
```diff
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>ru.yandex.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.1.40</version>
```
idlira, line 62: Round question: does it make sense this dependency explicitly declared here? Why not just inherit from sirius-db?

andyHa: sirius-db only declares as a test dependency, as sirius-db has no clickhouse specific code. biz does, hence the dependency here...

PR #1374, opened 2022-03-09
Provides a bunch of misc minor improvments.

andyHa:


src/main/java/sirius/biz/ide/Scripting.java
```diff
@@ -73,6 +73,14 @@
      */
     public static final String ALL_NODES = "*";
 
+    /**
+     * Imposes an upper limit on the message length for transcript messages.
+     * <p>
+     * As these are distributed across the cluster via redis, it would be fatal if a huge message would be sent
```
jakobvogel, line 79: _Will und would machen das If kaputt_ 😉

```suggestion
     * As these are distributed across the cluster via redis, it would be fatal if a huge message were to be sent
```

jakobvogel (approved):
Little remark.

idlira (approved):

PR #1375, opened 2022-03-10
Add parameter "showAddon" with default value "true".

ReadMails-ReallyFast:
Fixes: SIRI-542


sabieber (approved):

idlira (approved):

src/main/resources/default/taglib/t/lookupValue.html.pasta
```diff
@@ -6,6 +6,7 @@
 <i:arg name="help" type="String" default="@i18n(helpKey)"/>
 <i:arg name="optional" type="boolean" default="false"/>
 <i:arg name="readonly" type="boolean" default="false"/>
+<i:arg name="showAddon" type="boolean" default="true"/>
```
andyHa, line 9: maybe call this showSelectionButton or the like - addon is a broad term used by the single select itself, but here we know what it is used for

PR #1377, opened 2022-03-11
Provides a NO-OP shell for our built-in SSH server.

andyHa:
This is required by some SFTP clients, which first start a SSH
session and then switch over to SFTP.


marvinwoop (approved):

marvinwoop:
Wird nach Rücksprache mit aha gemerged.

PR #1378, opened 2022-03-11
Ensures system processes cannot accidentally be started in another tenant

cschierle:
Prevents exposing a system process started while logged in as another user or tenant.

**Reworked** with the provided feedback to achieve a framework-independent solution.
Also tried to improve naming and JavaDoc.

Fixes: SIRI-547

The previous approach violated the framework boundaries:
> - Exposes and uses some existing logic in `TenantController` to check, if a user is currently spying another tenant.
> - Adds the admin-only marker as a method in `BasicJobFactory`, so it can be overridden in subclasses.
> 
> Other projects using this might define different rulesets for when a process is considered as admin-only. So I thought performing the check via a method provides enough flexibility while still having a suitable default behavior.
 
Feel free to discuss possible other solutions. Comments are very much welcome :)

### DIFF
--- a/src/main/java/sirius/biz/jobs/BasicJobFactory.java
+++ b/src/main/java/sirius/biz/jobs/BasicJobFactory.java
@@ -12,6 +12,9 @@ import sirius.biz.jobs.infos.JobInfo;
 import sirius.biz.jobs.infos.JobInfoCollector;
 import sirius.biz.jobs.params.Parameter;
 import sirius.biz.jobs.presets.JobPresets;
+import sirius.biz.tenants.Tenant;
+import sirius.biz.tenants.TenantController;
+import sirius.biz.tenants.TenantUserManager;
 import sirius.kernel.async.TaskContext;
 import sirius.kernel.commons.Monoflop;
 import sirius.kernel.commons.Strings;
@@ -61,6 +64,18 @@ public abstract class BasicJobFactory implements JobFactory {
      * rendered nicely.
      */
     private static final String PARAM_UPDATE_ONLY = "updateOnly";
+
+    /**
+     * Contains the name of the key used to pass along the {@link Tenant admin tenant} ID,
+     * when a spying user starts an {@link #shouldExecuteInAdminTenant() admin-only} job.
+     */
+    public static final String ADMIN_TENANT_ID_KEY = "adminTenantId";
+
+    /**
+     * Contains the name of the key used to pass along the {@link Tenant admin tenant} name,
+     * when a spying user starts an {@link #shouldExecuteInAdminTenant() admin-only} job.
+     */
+    public static final String ADMIN_TENANT_NAME_KEY = "adminTenantName";
 
     @Part
     @Nullable
@@ -210,6 +225,11 @@ public abstract class BasicJobFactory implements JobFactory {
         });
 
         if (submit.get()) {
+            if (TenantController.isCurrentlySpying(request) && shouldExecuteInAdminTenant()) {
+                Tenant<?> adminTenant = TenantController.determineCurrentTenant(request);
+                context.put(ADMIN_TENANT_ID_KEY, adminTenant.getIdAsString());
+                context.put(ADMIN_TENANT_NAME_KEY, adminTenant.getTenantData().getName());
+            }
             executeInteractive(request, context);
             return;
         }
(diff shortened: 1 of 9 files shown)


cschierle:
Maybe the naming could be improved.

The process would still be executed in the context of the tenant the user is currently spying. Only the process entity is assigned to the admin tenant, so it is not visible to the other tenant.

andyHa:
This approach violates the decoupling of the individual frameworks. The idea is, that jobs, processes and user management make little to no assumptions of implementation of others. Therefore this needs to be split up:

Therefore the processes framework should remain untouched. Also, we don't need to modify interactive jobs, as they don't leave any traces. Therefore, we can poll most of the logik (see inline comments) into **BatchProcessJobFactory**.

Now the hard part is to untangle the user management. For this, we need to introduce a new class "SpyUser" which contains a single field "UserInfo rootUser". This needs to be attached to the UserInfo in TenantUserManager#becomeSpyUser and TenantUserManager#createUserWithTenant. In BatchProcessJobFactory#startWithContext we can then check for "shouldExecuteAsSystemTenant" and also user.tryAs(SpyUser.class). If this is filled, we can create the process as root user.

src/main/java/sirius/biz/jobs/BasicJobFactory.java
```diff
@@ -271,6 +291,19 @@ protected final void checkPermissions() {
...
+    /**
+     * Indicates if this job should only be executed in an admin tenant.
+     * <p>
+     * This helps prevent executing jobs in the wrong tenant, when the user that starts the job
+     * {@link TenantController#isCurrentlySpying(WebContext) is currently spying} another tenant.
+     *
+     * @return <tt>true</tt> if the system administrator permission is required, <tt>false</tt> otherwise. Override this
+     * and make it return <tt>true</tt> whenever it should be ensured that this job is only executed in an admin tenant
+     */
+    protected boolean shouldExecuteInAdminTenant() {
```
andyHa, line 303: move to BatchProcessJobFactory

src/main/java/sirius/biz/jobs/BasicJobFactory.java
```diff
@@ -210,6 +225,11 @@ public void startInteractively(WebContext request) {
         });
 
         if (submit.get()) {
+            if (TenantController.isCurrentlySpying(request) && shouldExecuteInAdminTenant()) {
```
andyHa, line 228: interative jobs don't matter as they don't create processes for other tenants...

src/main/java/sirius/biz/process/Processes.java
```diff
@@ -156,8 +157,9 @@ public String createProcess(@Nullable String type,
         process.setIcon(icon);
         process.setUserId(user.getUserId());
         process.setUserName(user.getUserName());
-        process.setTenantId(user.getTenantId());
```
andyHa, line 159: handle in BatchProcessJobFactory

src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -342,32 +342,38 @@ public void selectTenants(WebContext webContext) {
                   .template("/templates/biz/tenants/select-tenant.html.pasta", tenants, isCurrentlySpying(webContext));
     }
 
-    private boolean isCurrentlySpying(WebContext webContext) {
```
andyHa, line 345: encapsulate in SpyUser as attachment to the UserInfo...

andyHa (changes requested):

sabieber (approved):

src/main/java/sirius/biz/jobs/batch/BatchProcessJobFactory.java
```diff
@@ -115,18 +118,29 @@ protected String executeInBackground(Map<String, String> context) {
...
 
         return processId;
     }
 
+    private UserInfo getCurrentOrRootUser() {
+        if (isSystemProcess()) {
+            return UserContext.getCurrentUser()
+                              .tryAs(SpyUser.class)
+                              .map(SpyUser::getRootUser)
+                              .orElse(UserContext.getCurrentUser());
```
andyHa, line 139:
```suggestion
                              .orElseGet(UserContext::getCurrentUser);
```

src/main/java/sirius/biz/web/SpyUser.java
```diff
@@ -0,0 +1,27 @@
...
+
+package sirius.biz.web;
+
+import sirius.kernel.di.transformers.Composable;
+import sirius.web.security.UserInfo;
+
+/**
+ * Holds a reference to the {@link UserInfo root user}, if currently spying another user or tenant.
+ */
+public class SpyUser extends Composable {
```
andyHa, line 17:
```suggestion
public class SpyUser {
```

andyHa (commented):
Minor improvements as comments...

sabieber (approved):

PR #1383, opened 2022-03-28
Allow setting the next value of a sequence in console

meg1502:


src/main/java/sirius/biz/sequences/SequencesCommand.java
```diff
@@ -27,6 +28,42 @@
 
     @Override
     public void execute(Output output, String... params) throws Exception {
+        if (params.length == 0) {
+            listSequences(output);
+            return;
+        }
+        if (params.length == 3) {
```
andyHa, line 35: If params != 3 -> output usage
otherwise set value

always list

src/main/java/sirius/biz/sequences/SequencesCommand.java
```diff
@@ -27,6 +28,42 @@
...
+    }
+
+    private void outputUsage(Output output) {
+        output.apply("Usage:");
+        output.apply("'sequences' to list all known sequences");
+        output.apply("'sequences set NAME NEXTVALUE' to set the next value for a sequence");
+        output.apply("'sequences force NAME NEXTVALUE' to force the next value for a sequence !DANGEROUS! as it can lead to duplicate ids");
+    }
+
+    private void setSequence(Output output, String... params) {
```
andyHa, line 49: specifyNextSequence(String sequence name, boolean force, int nextValue);

call with:
specifyNextSequence(params[1], "force".equalsIgnoreCase(values[0]) Long.parseLong(params[2]));

andyHa (changes requested):

andyHa (approved):

mkeckmkeck (approved):

PR #1392, opened 2022-04-14
Fixes a StringIndexOutOfBoundsException exception

idlira:
Calling `http://.../dasd/` leads to the expected 404 (Page not found), but `http://.../dasd` (without trailing slash) causes a StringIndexOutOfBoundsException exception.

Fixes: SIRI-541

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -1804,8 +1804,7 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
 
         String conversionUrl = "http://"
                                + randomHost
-                               + BlobDispatcher.URI_PREFIX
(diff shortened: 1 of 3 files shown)


meg1502 (approved):
Maybe we could even change the constant, as every time its used '/ is appended anyway.


if we do this, we could also change `URI_PREFIX_LENGTH` from a own constant to a simple call to `URI_PREFIX`.length()

andyHa:
Please update the constant so that we don't always construct a new StringBuilder+String on each invocation.
Also please keep the constant length around, which also saves an evaluation each time (afaik, javac doesn't perform a constant propagation in this case...)...

andyHa (changes requested):

meg1502 (approved):

sabieber (approved):

meg1502 (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #1396, opened 2022-04-21
Adds ability to log the entity creation in DelegateJournalData

meg1502:


src/main/java/sirius/biz/protocol/DelegateJournalData.java
```diff
@@ -43,6 +43,9 @@ public class DelegateJournalData extends Composite {
     @Transient
     private volatile boolean silent;
 
+    @Transient
+    private boolean logCreation;
```
sabieber, line 47: Does this also have to be volatile like the silent flag above?

meg1502: 🤷‍♂️


i'll check

PR #1414, opened 2022-06-02
Supports creating uplinks to FTPS servers

sabieber:
To connect to an FTPS server successfully not the FTPClient has to be used but its subclass FTPSClient. To do this we extend the FTPUplink and FTPUplinkConnectorConfig and setup the new client internally.

Fixes: SIRI-584

### DIFF
--- a/src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPSUplinkConnectorConfig.java
+++ b/src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPSUplinkConnectorConfig.java
@@ -10,53 +10,42 @@ package sirius.biz.storage.layer3.uplink.ftp;
 
 import org.apache.commons.net.ftp.FTP;
 import org.apache.commons.net.ftp.FTPClient;
-import sirius.biz.storage.layer3.uplink.util.UplinkConnectorConfig;
+import org.apache.commons.net.ftp.FTPSClient;
 import sirius.biz.storage.util.StorageUtils;
 import sirius.kernel.commons.Value;
 import sirius.kernel.health.Exceptions;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.util.function.Function;
 
 /**
- * Keeps the configuration used to build an FTP connector using the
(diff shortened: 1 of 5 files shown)


idlira (approved):

andyHa (approved):

src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPSUplinkConnectorConfig.java
```diff
@@ -0,0 +1,62 @@
...
+            client.execPROT("P");
+            client.enterLocalPassiveMode();
+
+            return client;
+        } catch (IOException exception) {
+            throw Exceptions.handle()
+                            .to(StorageUtils.LOG)
+                            .error(exception)
+                            .withSystemErrorMessage(
+                                    "Layer 3/FTP: An error occurred while connecting the uplink %s: %s (%s)",
```
mkeckmkeck, line 57:
```suggestion
                                    "Layer 3/FTPS: An error occurred while connecting the uplink %s: %s (%s)",
```

mkeckmkeck (approved):

src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPSUplinkConnectorConfig.java
sabieber: Don't think thats a good idea. If we want that we have to overwrite a lot of methods of FTPUplink and FTPUplinkConnectorConfig to output /FTPS everywhere.

Or we have to extract the logged layer into a method.

meg1502 (approved):

src/main/java/sirius/biz/storage/layer3/uplink/ftp/FTPSUplinkConnectorConfig.java
sabieber: Discussed with @andyHa - we keep it as FTP (that its actually FTPS can be seen in the stacktrace)

PR #1431, opened 2022-07-08
File downloader bugfixing

qw3ry:
How to migrate:
- usages of `determineEffectiveFilename` should add `true` as second parameter to keep the previous behavior
- Overrides of `determineEffectiveFilename` should call the super method when `includeDate` is false to keep the previous behavior

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/file/FileExportJob.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/FileExportJob.java
@@ -95,7 +95,7 @@ public abstract class FileExportJob extends BatchJob {
                 return fileDestination.createOutputStream();
             }
 
-            return process.addFile(determineFilenameWithoutExtension() + "." + determineFileExtension());
+            return process.addFile(determineEffectiveFilename("", false));
         } catch (IOException e) {
             throw process.handle(e);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/jobs/batch/file/FileExportJob.java
```diff
@@ -178,8 +176,8 @@ protected String determineEffectiveFilename(String suffix) {
     /**
      * Determines the base name to use for the file.
      * <p>
-     * This will be expanded by the date and also by additional suffixes to generate a unique name. Also the file
-     * extension as supplied by {@link #determineFileExtension()} will be appended.
+     * The file extension as supplied by {@link #determineFileExtension()} will be appended.
+     * If an unique name is required, a date and potentially additional suffixes are appended.
```
ymo-sci, line 180:
```suggestion
     * If a unique name is required, a date and potentially additional suffixes are appended.
```

src/main/java/sirius/biz/jobs/batch/file/FileExportJob.java
```diff
@@ -154,18 +154,16 @@ private boolean shouldUseProvidedOutputDirectory() {
...
-               + "-"
-               + today.getYear()
-               + "-"
-               + today.getMonthValue()
-               + "-"
-               + today.getDayOfMonth()
-               + suffix
-               + "."
-               + determineFileExtension();
+    protected String determineEffectiveFilename(String suffix, boolean includeDate) {
```
mkeckmkeck, line 157: Javadoc for includeDate

src/main/java/sirius/biz/jobs/batch/file/FileExportJob.java
```diff
@@ -151,21 +151,20 @@ private boolean shouldUseProvidedOutputDirectory() {
     /**
      * Determines the effective file name to use.
      *
-     * @param suffix an optional suffix to append to the name to generate a unique file name
+     * @param suffix      an optional suffix to append to the name to generate a unique file name
+     * @param includeDate whether the current date should be appended to the filename
      * @return a full file name including the file extension
      */
-    protected String determineEffectiveFilename(String suffix) {
```
jakobvogel, line 157: Keep and deprecate old signature, forwarding to the new one in compatible fashion?

PR #1434, opened 2022-07-14
Increases the max upload size a little

mkeckmkeck:
- currently we have not uploadable files
- new limit should be 10 * 1024 * 1024 *10000 bytes = 104,8576 gb

Fixes: SE-11869

### DIFF
--- a/src/main/java/sirius/biz/storage/s3/ObjectStore.java
+++ b/src/main/java/sirius/biz/storage/s3/ObjectStore.java
@@ -87,7 +87,7 @@ public class ObjectStore {
      * {@link #INITIAL_LOCAL_AGGREGATION_BUFFER_SIZE}) before a chunk is uploaded to S3. Note that a single multipart
      * upload can at most consist of 10.000 parts - therefore this size limits the maximal total object size.
      */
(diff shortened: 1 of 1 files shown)


andyHa (approved):

jakobvogel (approved):

src/main/java/sirius/biz/storage/s3/ObjectStore.java
```diff
@@ -87,7 +87,7 @@ public class ObjectStore {
      * {@link #INITIAL_LOCAL_AGGREGATION_BUFFER_SIZE}) before a chunk is uploaded to S3. Note that a single multipart
      * upload can at most consist of 10.000 parts - therefore this size limits the maximal total object size.
      */
-    private static final int MAXIMAL_LOCAL_AGGREGATION_BUFFER_SIZE = 8 * 1024 * 1024;
+    private static final int MAXIMAL_LOCAL_AGGREGATION_BUFFER_SIZE = 10 * 1024 * 1024;
```
sabieber, line 90: Maybe add the calculation (result) to the JavaDoc so one does not have to calculate themselves when seeing this value in the future

sabieber (approved):

fhaScireum (approved):

PR #1437, opened 2022-07-16
Auto-Creation of System Tenant with Correct ID

jakobvogel:
Attention, this is a **breaking change:** Systems can no longer rely on `wasCreated()` or `isNew()` to determine the need for extended setup routines, for instance in `@AfterSave` routines. Implement a `TenantAutoSetupExtender` to perform setup tasks explicitly.

### DIFF
--- a/src/main/java/sirius/biz/tenants/jdbc/SQLTenantAutoSetup.java
+++ b/src/main/java/sirius/biz/tenants/jdbc/SQLTenantAutoSetup.java
@@ -9,12 +9,16 @@
 package sirius.biz.tenants.jdbc;
 
 import sirius.biz.tenants.BaseTenantAutoSetup;
+import sirius.biz.tenants.TenantData;
 import sirius.db.jdbc.OMA;
+import sirius.db.mixing.Mixing;
 import sirius.kernel.AutoSetup;
 import sirius.kernel.AutoSetupRule;
+import sirius.kernel.commons.Context;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Register;
+import sirius.kernel.health.Exceptions;
 
 /**
  * Creates an initial tenant and user if none are available.
@@ -29,6 +33,9 @@ public class SQLTenantAutoSetup extends BaseTenantAutoSetup {
     @Part
     private OMA oma;
 
+    @Part
+    private Mixing mixing;
+
     @Override
     public void setup() throws Exception {
         if (oma.select(SQLTenant.class).exists()) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/tenants/mongo/MongoTenantAutoSetup.java
```diff
@@ -45,24 +45,27 @@ public void setup() throws Exception {
...
-        tenant = mango.find(MongoTenant.class, systemTenantId)
-                      .orElseThrow(() -> new IllegalStateException(
-                              "Failed to resolve the system tenant after changing its ID!"));
+        // We can not modify the ID of a new tenant entity directly. Thus, we create an empty tenant with the proper ID
+        // using low-level techniques first, just to be filled properly in an instant.
+        mongo.insert().set(MongoEntity.ID, systemTenantId).into(MongoTenant.class);
+
+        // We load the newly created empty tenant via high-level techniques, making use of default values set in the
+        // entity
+        MongoTenant tenant = mango.find(MongoTenant.class, systemTenantId)
```
jakobvogel, line 62: Does this work w.l.o.g.? Can loading an almost empty tenant structure lead to a crash?

meg1502 (approved):

sabieber (approved):

src/main/java/sirius/biz/tenants/mongo/MongoTenantAutoSetup.java
andyHa: Properties should generally handle "null" gracefully but sticking to their defaut value...

src/main/java/sirius/biz/tenants/mongo/MongoTenantAutoSetup.java
```diff
@@ -45,24 +45,27 @@ public void setup() throws Exception {
         }
 
         AutoSetup.LOG.INFO("Creating system tenant....");
```
andyHa, line 47: should we adjust the logic in SQLTenantAutoSetup to match the following?

jakobvogel: done

src/main/java/sirius/biz/tenants/jdbc/SQLTenantAutoSetup.java
```diff
@@ -40,24 +47,35 @@ public void setup() throws Exception {
...
-                   .where(SQLTenant.ID, tenant.getId())
-                   .executeUpdate();
-            } catch (Exception e) {
-                AutoSetup.LOG.WARN("Failed to update ID of system tenant from %s to %s: %s",
-                                   tenant.getId(),
-                                   tenants.getSystemTenantId(),
-                                   e.getMessage());
-            }
+            systemTenantId = Long.parseLong(tenants.getSystemTenantId());
         }
```
jakobvogel, line 56: The new, mongo-like way to determine the ID is different from the old implementation (unless OMA assigns ID incrementally and starts at 1 anyway?) Is this what we want to have?

andyHa (approved):

mkeckmkeck (approved):

PR #1440, opened 2022-07-20
Adds an empty default value to the name argument

mko-sci:
This change allows the link tag to be used without a name and solely an icon.

Fixes: OX-8823

### DIFF
--- a/src/main/resources/default/taglib/k/link.html.pasta
+++ b/src/main/resources/default/taglib/k/link.html.pasta
@@ -1,5 +1,5 @@
 <i:arg type="String" name="link" description="Defines a url or link."/>
-<i:arg type="String" name="name" description="Defines a speaking name or description."/>
+<i:arg type="String" name="name" default="" description="Defines a speaking name or description."/>
 <i:arg type="String" name="icon" default="fas fa-external-link-alt" description="Defines the icon left to the link name."/>
 
 <i:pragma name="description">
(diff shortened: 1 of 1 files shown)


src/main/resources/default/taglib/k/link.html.pasta
```diff
@@ -1,5 +1,5 @@
 <i:arg type="String" name="link" description="Defines a url or link."/>
-<i:arg type="String" name="name" description="Defines a speaking name or description."/>
+<i:arg type="String" name="name" default="" description="Defines a speaking name or description."/>
```
andyHa, line 2: should we then also skip the &nbsp below? or, we could add s.th. like `<i:if test="isFilled(name)"><span class="ml-2">@name</span></i:if>` - maybe ml-1 - need t ocheck this. I think we do this elsewhere to get consistent spacing between text and icons.

mko-sci: thats a good call. I am going to add this

mko-sci: ml-1 should be sufficient

jakobvogel (approved):

src/main/resources/default/taglib/k/link.html.pasta
mko-sci: done

cschierle (approved):

PR #1442, opened 2022-07-22
Improves the versatility of the link tag

mko-sci:
Fixes: OX-8823

### DIFF
--- a/src/main/resources/default/taglib/k/link.html.pasta
+++ b/src/main/resources/default/taglib/k/link.html.pasta
@@ -2,7 +2,7 @@
 <i:arg type="String" name="link" description="Defines a url or link."/>
 <i:arg type="String" name="name" default="" description="Defines a speaking name or description."/>
 <i:arg type="String" name="icon" default="fas fa-external-link-alt" description="Defines the icon left to the link name."/>
-<i:arg type="String" name="target" default="_blank" description="Defines the target where to display the linked url or link."/>
+<i:arg type="String" name="target" default="_self" description="Defines the target where to display the linked url or link."/>
 
 <i:pragma name="description">
     Renders a link uniformly with a speaking name or description.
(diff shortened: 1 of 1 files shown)


idlira (changes requested):
This is a breaking change, as it modifies the default behavior for links to open in a new tab. IF it's OK, the PR should be marking as breaking.
Otherwise we need the inverted behavior: one must tell that the link should open in a new tab. Lets see what others / Andi say...

src/main/resources/default/taglib/k/link.html.pasta
```diff
@@ -1,11 +1,13 @@
+<i:arg type="String" name="id" default="" description="Defines the id."/>
 <i:arg type="String" name="link" description="Defines a url or link."/>
 <i:arg type="String" name="name" default="" description="Defines a speaking name or description."/>
 <i:arg type="String" name="icon" default="fas fa-external-link-alt" description="Defines the icon left to the link name."/>
+<i:arg type="String" name="target" default="_blank" description="Defines the target where to display the linked url or link."/>
```
jakobvogel, line 5: `_self` seems to be a more obvious default ... or empty, and then `isFilled`

sabieber: Would also prefer a non breaking version of this to not have to check all places where the link tag is used.

mko-sci: okay then I am going to put "_self" as default behavior

mko-sci: done

mko-sci:
> This is a breaking change, as it modifies the default behavior for links to open in a new tab. IF it's OK, the PR should be marking as breaking. Otherwise we need the inverted behavior: one must tell that the link should open in a new tab. Lets see what others / Andi say...

Change the default behavior to _self. Therefore this change should not be breaking anymore.

idlira (approved):

sabieber (approved):

andyHa (approved):

jakobvogel (approved):

PR #1444, opened 2022-08-01
SIRI-595: Tycho Jobs

qw3ry:
# Changes

- Update to Tycho for `/jobs` and `/job/:1`
- Job parameter Javascript magic:
  - a parameter can now be hidden/cleared/updated based on the value of other parameters
  - you can add validation messages to the parameters
- Deprecated `getAutocompleteUri`. Instead, an `Autocompleter` should be implemented and registered by name. You can probably move the code from the autocomplete controller to the autocompleter without many changes. However, you only need to provide the entities, not full `AutocompleteHelper.Completion`s.

## Breakings

- `DirectoryParameter`, `FileParameter` and `FileOrDirectoryParameter` have been removed. Use the new `FileParameter` instead (different package than the old `FileParameter`). You should specify at the parameter wheter files, directories or both are allowed.
- Job Icons and Job Category Icons need to be migrated to Fontawesome 5
  - Replace "fa-icon-name" with "fa fa-icon-name"
  - If the icon has been renamed in Fontawesome 5, [check the new name](https://fontawesome.com/search?s=solid%2Cbrands)
- `Parameter`s and `ParameterBuilder`s `getEffectiveTemplate(context)` was removed. They are no longer needed, because the dispatching between empty template and `getTemplateName()` is happening client-side now
-  Custom Parameter's templates have to be migrated to tycho, because the wondergem page-script is not supplied anymore
-  The deprecated `VirtualObjectParameter` has been removed

grafik
grafik

## Example use case for the javascript magic
grafik


mkeckmkeck (changes requested):

src/main/java/sirius/biz/jobs/BasicJobFactory.java
```diff
@@ -296,4 +299,30 @@ public Map<String, String> buildAndVerifyContext(Function<String, Value> paramet
 
         return context;
     }
+
+    @Override
+    public JSON computeRequiredParameterUpdates(WebContext ctx) {
```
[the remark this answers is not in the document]
qw3ry: unfortunately, no, because if I return a JSON String, the string will be quoted/escaped when written to the actual json response. I _could_ pass the StructuredOutput as parameter, BUT that would make it harder to include the json string in `job.html.pasta`

src/main/java/sirius/biz/jobs/params/EntityParameter.java
```diff
@@ -127,10 +157,27 @@ public Tuple<String, String> renderCurrentValue(Map<String, String> context) {
...
+        }
         return entity.toString();
     }
 
+    @Override
+    public Optional<?> updateValue(Map<String, String> ctx) {
+        return updater.apply(ctx).map(value -> {
+            Map<String, String> map = new HashMap<>();
+            map.put("value", value.getIdAsString());
+            map.put("text", createLabel(value));
```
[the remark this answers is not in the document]
qw3ry: This is the naming that is specified by token-autocomplete. We just pass the json object there to update the value

src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -124,6 +124,19 @@
                     } else {
                         _elem.setAttribute("hidden", true);
                     }
+                    const _message = sirius.querySelector('.param-' + name + '-validation');
+                    if (_message) {
+                        _message.parentNode.removeChild(_message);
+                    }
+                    if (typeof param.validation !== 'undefined') {
+                        // pretty hacky to split css classes, but it works, and unless we change the Message type there is no good way around it
+                        const alertType = param.validation.class.split("-")[1];
```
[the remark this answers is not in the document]
qw3ry: You mean we move the hack to the backend?
I think this would be an improvement 👍

src/main/java/sirius/biz/jobs/JobFactory.java
```diff
@@ -177,4 +178,12 @@ Map<String, String> buildAndVerifyContext(Function<String, Value> parameterProvi
      * @return the name of the job category
      */
     String getCategory();
+
+    /**
+     * Compute a json containing the required update operations for the javascript frontend.
```
sabieber, line 183:
```suggestion
     * Computes a JSON containing the required update operations for the JavaScript frontend.
```

src/main/java/sirius/biz/jobs/JobFactory.java
```diff
@@ -177,4 +178,12 @@ Map<String, String> buildAndVerifyContext(Function<String, Value> parameterProvi
...
+     *
+     * @param ctx the web context containing the values of all the parameters
+     * @return a json that can be handled by the javascript
```
sabieber, line 186:
```suggestion
     * @return a JSON that can be handled by the JavaScript
```

src/main/java/sirius/biz/jobs/JobsController.java
```diff
@@ -84,4 +103,22 @@ public void infos(WebContext ctx, String jobType) {
...
+     * @param ctx               the web context
+     * @param out               the output to write the JSON response to
+     * @param autocompleterName the name of the autocompleter
+     */
+    @Routed("/jobs/parameter-autocomplete/:1")
+    @InternalService
+    @SuppressWarnings("unchecked")
+    @Explain("Because Autocompleter#suggest does not use the template parameter in its signature,"
+             + " it really does not matter.")
+    public void autocomplete(WebContext ctx, JSONStructuredOutput out, String autocompleterName) {
```
sabieber, line 119: Optional: Rename to webContext

src/main/java/sirius/biz/jobs/params/Autocompleter.java
```diff
@@ -0,0 +1,63 @@
...
+import sirius.kernel.di.std.AutoRegister;
+import sirius.kernel.di.std.Named;
+import sirius.web.controller.AutocompleteHelper;
+import sirius.web.http.WebContext;
+
+import java.util.Collection;
+import java.util.function.Consumer;
+
+/**
+ * An abstract class that can be implemented to supply {@link AutocompleteHelper.Completion Completions} for entity-based token-autocomplete fields.
```
sabieber, line 21: Optional: Please begin JavaDoc with a verb describing what the interface does. The current JavaDoc is even false as it mentions an abstract class, but the documented code provides an interface instead.

Something like this would be my suggestion: `Allows to supply completions for entity-based token-autocomplete fields.`

src/main/java/sirius/biz/jobs/params/Autocompleter.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Computes the possible completions for the query and the web context.
+     * <p>
+     * Implementations may return the items in an ordered collection. This order is retained.
+     *
+     * @param query   the query to search for
+     * @param context the web context with possibly additional context to the search
+     * @return a potentially ordered collection of completions
+     */
+    Collection<T> getCompletions(String query, WebContext context);
```
sabieber, line 37: Optional: Rename to `webContext`. `context` is better than `ctx` but as we have quite a view different context classes in sirius in my opinion it helps quite a lot to name them as such.

src/main/java/sirius/biz/jobs/params/FileParameter.java
```diff
@@ -23,38 +24,93 @@
...
+    public boolean isFilesOnly() {
+        return allowFile && !allowDirectory;
+    }
+
+    /**
+     * Whether files are allowed as parameter.
+     *
+     * @return true if files are allowed
+     */
+    public boolean allowFile() {
```
sabieber, line 92: Optional: Rename to `allowsFile`. The wording `allowFile` implies it is actually allowing files and not simply checking what the current value is.

src/main/java/sirius/biz/jobs/params/FileParameter.java
```diff
@@ -23,38 +24,93 @@
...
+        return allowFile;
+    }
+
+    /**
+     * Whether directories are allowed as parameter.
+     *
+     * @return true if directories are allowed
+     */
+    public boolean allowDirectory() {
```
sabieber, line 101: Optional: Rename to `allowsDirectory`. The wording `allowDirectory` implies it is actually allowing files and not simply checking what the current value is.

src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -109,6 +109,35 @@
...
+
+            function handleParameterUpdates(json) {
+                for (const name in json.params) {
+                    const _elem = sirius.querySelector('.param-' + name);
+                    let param = json.params[name];
+                    if (param.visible) {
+                        _elem.removeAttribute("hidden");
+                    } else {
+                        _elem.setAttribute("hidden", true);
+                    }
```
sabieber, line 129: Optional: Sorry if this sounds snarky but what is the benefit of saving 3 characters here instead of just writing `_element` which is more readable to the human eye?

src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -109,6 +109,35 @@
...
+                }
+            }
+
+            document.querySelectorAll('.edit-form input[name]').forEach(function (input) {
```
sabieber, line 152: Optional: `_input` instead of `input` as it is a DOM element
Also check the code below.

src/main/resources/default/templates/biz/jobs/job.html.pasta
```diff
@@ -124,6 +124,12 @@
                     } else {
                         _elem.setAttribute("hidden", true);
                     }
+                    if (param.clear) {
+                        _elem.dispatchEvent(new CustomEvent("clear"));
+                    }
+                    if (typeof param.updatedValue !== 'undefined') {
+                        _elem.dispatchEvent(new CustomEvent("updated-value", {detail: param.updatedValue}));
+                    }
```
sabieber, line 135: There is now a sirius method that allows to trigger custom events while supporting all browsers. it is called `sirius.dispatchEvent('event')'`. Maybe we could expand it to optionally also take a details object.

src/main/resources/default/templates/biz/jobs/params/lookuptable.html.pasta
```diff
@@ -9,4 +9,26 @@
...
+        });
+    </i:block>
+
+    <i:if test="isFilled(suggestionUri)">
+        <i:block name="autocompleteOptions">
+            {
+                suggestionsUriBuilder: function (query) {
+                    const context = sirius.getFormData(sirius.querySelector(".edit-form"));
+                    return "@suggestionUri" + "?query="
+                            + query + "&" + new URLSearchParams(context).toString();
```
sabieber: Why not add the query to the search params and append the toString directly after the questionmark? seems a bit less "hacky"

src/main/java/sirius/biz/codelists/LookupTableParameter.java
```diff
@@ -67,6 +69,16 @@ protected String checkAndTransformValue(Value input) {
                                                         .handle());
     }
 
+    @Override
+    public Optional<?> updateValue(Map<String, String> ctx) {
```
sabieber: Optional: Choose more descriptive name

src/main/java/sirius/biz/jobs/JobConfigData.java
```diff
@@ -155,7 +155,7 @@ public Value fetchParameter(String key) {
     public void loadFromContext(WebContext ctx) {
         // Check all parameters here to notify the user about config short comings...
         ValueHolder<HandledException> errorHolder = new ValueHolder<>(null);
-        Map<String, String> data = getJobFactory().buildAndVerifyContext(ctx::get, true, ex -> {
+        Map<String, String> data = getJobFactory().buildAndVerifyContext(ctx::get, true, (p, ex) -> {
```
sabieber, line 158: Optional: At least write it as `param` if `parameter` is to long

sabieber (changes requested):
General observations:
- Some results of querySelector are not null checked (like `_input` for example) is it save that the element is always present?
- I actually don't like the old mason grid for the job categories that got ported. Was this discussed? Is this the scope of another ticket? If I remember correctly Andy mentioned that this will be improved when we port / overhaul it.
- Nice job with the parameter logic 👍 Will be very useful.
- Shouldn't the field also be highlighted when a validation warning occurs? (Like this: https://getbootstrap.com/docs/4.0/components/forms/#server-side)

src/main/resources/default/templates/biz/jobs/job.html.pasta
qw3ry: It's actually 6*3 charactes 😲 changed it anyway :P

sabieber (approved):

sabieber:
Danke für die Überarbeitungen 👌

qw3ry:
> Some results of querySelector are not null checked (like `_input` for example) is it save that the element is always present?
In theory yes (because the templates contain those fields), however, if the used templates change at some point, this might indeed cause NPEs (which will be an actual programming error, then)

> I actually don't like the old mason grid for the job categories that got ported. Was this discussed? Is this the scope of another ticket? If I remember correctly Andy mentioned that this will be improved when we port / overhaul it.
Actually I changed the grid from an actual grid to the mason grid because I felt that in our products, some categories hold many jobs, while others dont, which leads to a messy layout. However, this was not discussed at all, but I am open to discussion, so please feel free to schedule a meeting.

> Shouldn't the field also be highlighted when a validation warning occurs? (Like this: https://getbootstrap.com/docs/4.0/components/forms/#server-side)
I agree this would be nice; however, this does not work so well with how the code is structured currently. To avoid an even bigger PR (and probably a harder migration) I would suggest adding this at some point in the future

mkeckmkeck (approved):

sabieber (approved):

cschierle (approved):

src/main/java/sirius/biz/jobs/params/ParameterBuilder.java
```diff
@@ -100,24 +100,24 @@ public P withDescription(String description) {
...
-     * @param test the predicate
+     * @param shouldHide the predicate
      * @return the parameter itself for fluent method calls
      * @see #isVisible(Map) for the usage
      */
-    public P hideWhen(Predicate<Map<String, String>> test) {
-        shouldHide = test;
+    public P hideWhen(Predicate<Map<String, String>> shouldHide) {
+        this.shouldHide = shouldHide;
         return self();
```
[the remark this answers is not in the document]
mkeckmkeck: It's a kind of type magic with generics. Have a look at the implementation of self for a hint.

fhaScireum (approved):

meg1502 (approved):

PR #1456, opened 2022-08-29
Adds enable/disable auto refresh button to process view

mko-sci:
View:
Bildschirmfoto 2022-08-29 um 16 48 41
View on Hover:
Bildschirmfoto 2022-08-29 um 16 48 47

The "button" (which are actually two) enables/disables the auto refresh of the process view. This could be handy if someone would like to copy text. The advantage of using two a-tag typed "buttons" is that the selection will be added to the url and can be shared that way.

Fixes: [SIRI-608](https://scireum.myjetbrains.com/youtrack/issue/SIRI-608)


src/main/resources/default/templates/biz/process/process.html.pasta
```diff
@@ -148,7 +149,9 @@
         <script type="text/javascript">
             sirius.ready(function () {
                 setTimeout(function () {
-                    window.location.reload();
+                    if (___isAutoRefreshEnabled) {
+                        window.location.reload();
+                    }
```
fhaScireum, line 154: This solution is static, meaning if 'isAutoRefreshEnabled' is false, it will stay false (without a page reload) and the time out block spins around doing nothing.

I think we could have an even better user experience and simpler code by simply storing the blockage of th refresh in a simple JS boolean field and control the behaviour of the refresh with that.

Also, if we write the setting into the URL parameters (e.g. ".../...?refresh=false") we could propagate the setting when filtering the messages or switching the page in order to not reload even though we are still practically from a user point of view (even if there are page reloads) on the same page.

mko-sci: > This solution is static, meaning if 'isAutoRefreshEnabled' is false, it will stay false (without a page reload) and the time out block spins around doing nothing.
- In that case the `setTimout` block will be triggered exactly **once** without doing anything, but I agree. In that case it might be better to move the `if` statement a few lines up, in order to prevent it triggering even once.
- Note: The complete <script> tag is only added if `<i:if test="process.shouldAutorefresh()">`
- Note: Take a look at how the reload worked before. I basically only added one `if` statement.

> I think we could have an even better user experience [...]
- How does your suggestion affect the user experience? Currently, after pressing "disable auto refresh" the page stops auto refreshing. Therefore it does exactly what a user would expect.

> [...] and simpler code by simply storing the blockage of th refresh in a simple JS boolean field and control the behaviour of the refresh with that.
- I thougt about solving it with pure JS at first too, but @sabieber did not like the solution, or at least suggested that I solve it as I did here.

> Also, if we write the setting into the URL parameters (e.g. ".../...?refresh=false") we could propagate the setting when filtering the messages or switching the page in order to not reload even though we are still practically from a user point of view (even if there are page reloads) on the same page.
- Could you please explain this further, I did not get your point. (Gern auch in deutsch). :)

src/main/resources/default/templates/biz/process/process-logs.html.pasta
```diff
@@ -1,13 +1,27 @@
 <i:arg type="sirius.biz.process.Process" name="process"/>
 <i:arg type="Page" name="logs"/>
+<i:arg type="boolean" name="isAutoRefreshEnabled"/>
```
sabieber, line 3:
```suggestion
<i:arg type="boolean" name="autoRefresh"/>
```

src/main/resources/default/templates/biz/process/process-logs.html.pasta
```diff
@@ -1,13 +1,27 @@
...
 
 <i:invoke template="/templates/biz/process/process.html.pasta"
           process="process"
-          page="logs">
+          page="logs"
+          isAutoRefreshEnabled="isAutoRefreshEnabled">
```
sabieber, line 8:
```suggestion
          autoRefresh="autoRefresh">
```

src/main/resources/default/templates/biz/process/process-logs.html.pasta
```diff
@@ -1,13 +1,27 @@
...
     <i:block name="sidebar">
         <t:filterbox page="logs" baseUrl="@apply('/ps/%s', process.getId())"/>
     </i:block>
     <i:block name="header-actions">
+        <i:if test="process.shouldAutorefresh()">
+            <a id="enableAutoRefresh" class="btn btn-link d-none" href="@apply('/ps/%s?isAutoRefreshEnabled=true', process.getId())">@i18n("Process.enableAutoRefresh")</a>
+            <a id="disableAutoRefresh" class="btn btn-link d-none" href="@apply('/ps/%s?isAutoRefreshEnabled=false', process.getId())">@i18n("Process.disableAutoRefresh")</a>
```
sabieber, line 15: Check via i:if directly which button should be rendered instead of JS?

src/main/resources/default/templates/biz/process/process.html.pasta
```diff
@@ -1,5 +1,6 @@
 <i:arg type="sirius.biz.process.Process" name="process"/>
 <i:arg type="String" name="page"/>
+<i:arg type="boolean" name="isAutoRefreshEnabled"/>
```
sabieber, line 3:
```suggestion
<i:arg type="boolean" name="autoRefresh"/>
```

src/main/resources/default/templates/biz/process/process.html.pasta
```diff
@@ -148,7 +149,9 @@
         <script type="text/javascript">
             sirius.ready(function () {
                 setTimeout(function () {
-                    window.location.reload();
+                    if (___isAutoRefreshEnabled) {
```
sabieber, line 152:
```suggestion
                    if (___autoRefresh) {
```

fhaScireum (approved):

src/main/resources/default/templates/biz/process/process.html.pasta
sabieber: @fhaScireum I suggested the parameter solution as it would stick around even when the user actively reloads the page (which he will after deactivating the auto refresh). If there are arguments against that we can change it of course.

src/main/resources/default/templates/biz/process/process-logs.html.pasta
mko-sci: Yeah, actualy that might be a good idea. I am going to check it out. =)

mko-sci: Done

src/main/resources/default/templates/biz/process/process.html.pasta
```diff
@@ -148,11 +148,11 @@
     <i:if test="process.shouldAutorefresh()">
```
sabieber, line 148: Ginge es nicht hier direkt den Check auf den autoRefresh Param mit rein zu nehmen? Also `autoRefresh && process.shouldAutorefresh()`

mko-sci: doch wird gemacht

PR #1463, opened 2022-09-01
 Provides support for time-series charts within the Data-Explorer.

andyHa:


src/main/java/sirius/biz/analytics/charts/explorer/EventCountingTimeseries.java
```diff
@@ -0,0 +1,34 @@
...
+
+import sirius.biz.analytics.events.Event;
+
+/**
+ * Provides a simple {@link TimeseriesComputer} which just counts the occurrences of an {@link Event} in the given
+ * time slice.
+ *
+ * @param <O> the type of entities being referenced by the chart
+ */
+public class EventCountingTimeseries<O> extends EventTimeseries<O> {
```
sabieber, line 19: Wouldn't be `TimeSeries` be the correct camel-casing?

src/main/java/sirius/biz/analytics/charts/explorer/EventTimeseries.java
```diff
@@ -0,0 +1,162 @@
...
+
+        return this;
+    }
+
+    @Override
+    public void compute(O object, Timeseries timeseries) throws Exception {
+        if (fieldsToSelect == null) {
+            throw new IllegalStateException("Missing call to withSelect");
+        }
+        String query = "SELECT " + fieldsToSelect + ", " + "MONTH(eventDate) as month, " + "YEAR(eventDate) as year";
```
sabieber, line 120: Isn't it potentially dangerous to build such queries by hand?

sabieber (changes requested):

src/main/java/sirius/biz/analytics/charts/explorer/EventTimeseries.java
andyHa: It sort of is, hence the warning in the JavaDoc of the class. The other option would be to accept a complete SQL statement, however the group by and the where part would be the same for all queries...

sabieber: Already missed that sentence in the JavaDoc during review 😅
So I'm skeptical that future Implementers of that class remember this important security notice. But I don't have a better suggestion right away...

idlira: Idea: make fieldsToSelect a list, which is filled with calls to addSelectField().
This method in turn parses field names passed with a pattern, and so avoiding injections.
This way the SQL generation would be somewhat safer.

andyHa: these field names are not names but things like `countIf(foo > bar) AS foobar)` which are very hard to parse and verify as correct. The approach is to eliminate the helpers and always provide and execute a whole query - an example will be in the follow-up PR.

sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/analytics/charts/explorer/Timeseries.java
```diff
@@ -0,0 +1,214 @@
...
+     */
+    public Stream<LocalDate> startDates() {
+        return ranges.stream().map(Tuple::getFirst);
+    }
+
+    /**
+     * Enforces {@link Granularity#MONTH} by creating a new {@link Timeseries} if necessary.
+     * <p>
+     * This can be used to answer to "daily" queries if only monthly queries are present. This will compact the
+     * ranges, so that only a reduces number of ranges is present.
```
cschierle:
```suggestion
     * ranges, so that only a reduced number of ranges is present.

**Review by cschierle — APPROVED** [review]

File: src/main/java/sirius/biz/analytics/charts/explorer/TimeSeries.java [hunk 4]
```diff
@@ -0,0 +1,214 @@
...
+     */
+    public Stream<LocalDate> startDates() {
+        return ranges.stream().map(Tuple::getFirst);
+    }
+
+    /**
+     * Enforces {@link Granularity#MONTH} by creating a new {@link TimeSeries} if necessary.
+     * <p>
+     * This can be used to answer to "daily" queries if only monthly queries are present. This will compact the
+     * ranges, so that only a reduces number of ranges is present.
```
cschierle, line 173:
```suggestion
     * ranges, so that only a reduced number of ranges is present.
```

PR #1467, opened 2022-09-02
SIRI-615: Properly limits the height of Doughnut Charts.

fhaScireum:
Previously, the set height of 180 was already ignored and instead, the chart was defined by the columns width.
This however now far exceeds the displayed height of the website.

Example:
Bildschirmfoto 2022-09-02 um 11 28 10


sabieber (approved):

andyHa:
Fix uniformly for all charts?

andyHa (changes requested):
Shouldn't we aply the same logic for all charts?

sabieber (approved):

qw3ry (approved):

ymo-sci (approved):

PR #1471, opened 2022-09-05
Fixes image width being wider than its parent container

mko-sci:
Fixes: OX-8830

### DIFF
--- a/src/main/resources/default/taglib/k/image.html.pasta
+++ b/src/main/resources/default/taglib/k/image.html.pasta
@@ -7,5 +7,5 @@
 </i:pragma>
 
 <a href="@hrefUrl">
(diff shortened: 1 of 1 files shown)


idlira (approved):

sabieber (approved):

fhaScireum (approved):

ymo-sci (approved):

jakobvogel (approved):

meg1502 (approved):
do we maybe also need mh-100 for images which have a bigger height than width?

fhaScireum:
> do we maybe also need mh-100 for images which have a bigger height than width?

I merged the PR but this might still be a good question :)

PR #1480, opened 2022-09-13
improve searching scheduled jobs typos, autoformat, toList

ymo-sci:
- fixes: SIRI-627


src/main/java/sirius/biz/jobs/BasicJobFactory.java
```diff
@@ -117,9 +116,7 @@ protected void collectJobInfos(JobInfoCollector collector) {
 
     @Override
     public List<String> getRequiredPermissions() {
-        return Arrays.stream(getClass().getAnnotationsByType(Permission.class))
-                     .map(Permission::value)
-                     .collect(Collectors.toList());
+        return Arrays.stream(getClass().getAnnotationsByType(Permission.class)).map(Permission::value).toList();
```
cschierle, line 119: This seems like a tiny change, but it might be dangerous and breaking. In contrast to `Collectors.toList()`, `Stream.toList()` returns an unmodifiable list.

Aren't there usages, where something is added to the required permissions are where they are overridden?

Same thought goes to the other `toList()` changes of this PR. Please make sure that the return value is not modified elsewhere.

cschierle (changes requested):

src/main/java/sirius/biz/jobs/BasicJobFactory.java
qw3ry: Please double-check, I remember problems with `toList()`-refactoring when working on tycho jobs in august. Might have been another place, though.

qw3ry (approved):
If the`.toList()`-refactorings are fine

src/main/java/sirius/biz/jobs/BasicJobFactory.java
mkeckmkeck: Change to explicit mutable List implementation, or break it and fix all usages and document as immutable

andyHa (approved):

sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/jobs/BasicJobFactory.java
andyHa: We should now be good to go - after the changes are merged....

sabieber (approved):

cschierle (approved):

mkeckmkeck (approved):

PR #1483, opened 2022-09-15
Properly computes the effective computation time of a proces.

andyHa:
We currently add the queue time for jobs. Also the metric was completele wrong if an export within the process was run after the actual task (as start / completed was modified).

Drive-by - changes the styling of the auto-refresh to be in line with other Tycho pages:
grafik
grafik


fhaScireum (approved):

src/main/java/sirius/biz/process/Process.java
```diff
@@ -219,6 +219,16 @@ public class Process extends SearchableEntity {
     @NullAllowed
     private LocalDateTime completed;
 
+    /**
+     * Contains the estimated computation time performed in this process.
+     * <p>
+     * Note that this is now an exact value, as we might skip very short computations (esp. for standby processes)
```
sabieber, line 225:
```suggestion
     * Note that this is not an exact value, as we might skip very short computations (esp. for standby processes)
```

src/main/java/sirius/biz/process/Process.java
```diff
@@ -219,6 +219,16 @@ public class Process extends SearchableEntity {
     @NullAllowed
     private LocalDateTime completed;
 
+    /**
+     * Contains the estimated computation time performed in this process.
```
sabieber, line 223:
```suggestion
     * Contains the estimated computation time performed in this process in seconds.
```
😜

sabieber (approved):

cschierle (approved):

mko-sci (approved):

PR #1497, opened 2022-09-24
Properly document and extends services provided by biz.

andyHa:
We now properly document the service to start jobs and to check the state of processes.

We also provide an entity query API for testing purposes.

Depends on:
* https://github.com/scireum/sirius-web/pull/1106

### DIFF
--- a/src/main/java/sirius/biz/model/QueryController.java
+++ b/src/main/java/sirius/biz/model/QueryController.java
@@ -21,6 +21,7 @@ import sirius.db.mixing.query.Query;
 import sirius.db.mixing.query.constraints.Constraint;
 import sirius.db.mongo.MongoEntity;
 import sirius.db.mongo.MongoQuery;
+import sirius.kernel.Sirius;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Watch;
@@ -31,6 +32,8 @@ import sirius.web.controller.Routed;
 import sirius.web.http.WebContext;
 import sirius.web.security.Permission;
 import sirius.web.security.UserContext;
+import sirius.web.services.InternalService;
+import sirius.web.services.JSONStructuredOutput;
 
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


src/main/java/sirius/biz/model/QueryController.java
```diff
@@ -165,6 +168,47 @@ private List<Property> determineVisibleProperties(EntityDescriptor descriptor) {
...
+     * Just like the UI, this expects <tt>type</tt> to contain a Mixing descriptor name as produced by
+     * {@link sirius.db.mixing.Mixing#getNameForType(Class)} and <tt>query</tt> to contain a proper query
+     * e.g. "id:X". Additionally <tt>limit</tt> can specify how many entities are to be returned.
+     * <p>
+     * Note that this isn't a public API as it isn't available on production systems.
+     *
+     * @param webContext the request to handle
+     * @param output     the output to write the retrieved entities to
+     */
+    @Permission(TenantUserManager.PERMISSION_SYSTEM_ADMINISTRATOR)
```
sabieber, line 185: This would mean we would have to place username and API Token of an account with system admin authority in the systemtest repo. Isn't this kind of a security hazard? It's just the test and staging systems but maybe there is a better solution?

sabieber (approved):

src/main/java/sirius/biz/model/QueryController.java
idlira: Yeah, perhaps this API could rely on a config-key with a white-list of IPs or Networks masks.

idlira (approved):

src/main/java/sirius/biz/model/QueryController.java
andyHa: drone can provide (and securely store) "secrets" (which we also use for docker tasks etc.")....

andyHa: maybe jenkins can do the same (?)

sabieber: Maybe we can check that with Jonas

idlira: Its noted in the ticket where we will implement such queries in our API tests.

PR #1501, opened 2022-09-30
Permits to load groups of CSV files in stages

idlira:
This is _cloning_ the existing approach used by the XMLImportJob.

By default, `collectImportFiles()` functions as before and provides a list of files to collect.

If desired though, several stages can be defined, each loading a different set of files. For example: a header file is processed first and based on its contents, a different set will be processed later.

Fixes: [OX-9070](https://scireum.myjetbrains.com/youtrack/issue/OX-9070)

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
@@ -167,7 +167,7 @@ public abstract class DictionaryBasedArchiveImportJob extends ArchiveImportJob {
 
             for (ImportFile importFile : importFiles) {
                 if (!TaskContext.get().isActive()) {
-                    break;
+                    return;
                 }
                 Optional<ExtractedFile> extractedFile = fetchEntry(importFile.filename);
                 if (extractedFile.isPresent()) {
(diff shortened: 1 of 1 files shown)


mko-sci (approved):

src/main/java/sirius/biz/jobs/batch/file/DictionaryBasedArchiveImportJob.java
```diff
@@ -143,24 +144,40 @@ protected DictionaryBasedArchiveImportJob(ProcessContext process) {
...
-        for (ImportFile importFile : importFiles) {
-            if (!TaskContext.get().isActive()) {
-                break;
+        for (Consumer<Consumer<ImportFile>> fileCollector : fetchStages()) {
+            List<ImportFile> importFiles = new ArrayList<>();
+            fileCollector.accept(importFiles::add);
+
+            for (ImportFile importFile : importFiles) {
+                if (!TaskContext.get().isActive()) {
+                    break;
```
andyHa, line 170: extract method and return?

andyHa: ...or just return

mkeckmkeck (approved):

fhaScireum (approved):

qw3ry (approved):

PR #1506, opened 2022-10-06
Allows the query api everywhere but prod

idlira:
Fixes: [SIRI-631](https://scireum.myjetbrains.com/youtrack/issue/SIRI-631)

### DIFF
--- a/src/main/java/sirius/biz/model/QueryController.java
+++ b/src/main/java/sirius/biz/model/QueryController.java
@@ -186,8 +186,8 @@ public class QueryController extends BizController {
     @Routed("/system/query/api")
     @InternalService
     public void queryApi(WebContext webContext, JSONStructuredOutput output) {
-        if (!Sirius.isDev() && !Sirius.isTest()) {
-            throw new IllegalStateException("The query API is only available in development or test systems.");
+        if (Sirius.isProd()) {
+            throw new IllegalStateException("The query API is not available in productive systems.");
         }
 
         EntityDescriptor descriptor = mixing.findDescriptor(webContext.get("type").asString())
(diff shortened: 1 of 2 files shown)


sabieber (approved):

andyHa (approved):

src/main/java/sirius/biz/model/QueryController.java
```diff
@@ -186,8 +190,11 @@ private List<Property> determineVisibleProperties(EntityDescriptor descriptor) {
     @Routed("/system/query/api")
     @InternalService
     public void queryApi(WebContext webContext, JSONStructuredOutput output) {
-        if (!Sirius.isDev() && !Sirius.isTest()) {
-            throw new IllegalStateException("The query API is only available in development or test systems.");
+        if (!queryApiEnabled) {
+            throw new IllegalStateException("The query API is not enabled in this system.");
+        }
+        if (Sirius.isProd()) {
+            throw new IllegalStateException("The query API is not available in productive systems.");
```
fhaScireum, line 197:
```suggestion
        if (!queryApiEnabled) {
            throw new IllegalStateException("The query API is not enabled on this system.");
        }
        if (Sirius.isProd()) {
            throw new IllegalStateException("The query API is not available on productive systems.");
```
Sounds better (and maybe is only correct?) this way.

sabieber (approved):

sabieber (approved):

src/main/java/sirius/biz/model/QueryController.java
idlira: The wise Internet says:
_Both “in the system” and “on the system” are correct. However, they have different meanings. “In the system” typically means that something is part of or related to the system. On the other hand, “on the system” typically means that something is being done with or using the system._

fhaScireum: Fascinating, so basically, "in the system" is even more correct here...

PR #1519, opened 2022-11-08
Handles an exception replicating 0-byte objects

idlira:
It is possible to upload 0-byte files to S3, but when using a multipart upload, an AmazonS3Exception will be thrown attempting to complete the upload if no parts are provided.

We detect this condition and ignore the error, as backing up a file with no content is pointless.

Fixes: [SIRI-651](https://scireum.myjetbrains.com/youtrack/issue/SIRI-651)

### DIFF
--- a/src/main/java/sirius/biz/storage/s3/ObjectStore.java
+++ b/src/main/java/sirius/biz/storage/s3/ObjectStore.java
@@ -47,6 +47,7 @@ import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Watch;
 import sirius.kernel.di.std.Part;
+import sirius.kernel.health.ExceptionHint;
 import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
 
@@ -73,6 +74,8 @@ public class ObjectStore {
      * Contains the name of the thread pool, which is used for all up- and downloads.
      */
     public static final String EXECUTOR_S3 = "s3";
+
+    public static final ExceptionHint EMPTY_INPUT_STREAM = new ExceptionHint("empty-input-stream");
 
     /**
      * When performing a multipart upload in {@link #upload(BucketName, String, InputStream)} we keep
@@ -714,6 +717,8 @@ public class ObjectStore {
         ensureBucketExists(bucket);
         InitiateMultipartUploadResult multipartUpload =
                 getClient().initiateMultipartUpload(new InitiateMultipartUploadRequest(bucket.getName(), objectId));
+
+        boolean emptyInputStream = false;
         try (Operation operation = new Operation(() -> Strings.apply("S3: Multipart upload of object % to %s",
                                                                      objectId,
                                                                      bucket), Duration.ofHours(4))) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/biz/storage/s3/ObjectStore.java
```diff
@@ -714,6 +717,8 @@ public void upload(BucketName bucket, String objectId, InputStream inputStream)
         ensureBucketExists(bucket);
         InitiateMultipartUploadResult multipartUpload =
                 getClient().initiateMultipartUpload(new InitiateMultipartUploadRequest(bucket.getName(), objectId));
+
+        boolean emptyInputStream = false;
```
mko-sci, line 721: I would rename the variable to: isInputStreamEmpty

mko-sci: Or better: describe from which the input stream originates or what the bool is actualy there to check for

mko-sci (approved):

src/main/java/sirius/biz/storage/s3/ObjectStore.java
```diff
@@ -726,6 +731,8 @@ public void upload(BucketName bucket, String objectId, InputStream inputStream)
                                       eTags.size());
             }
 
+            emptyInputStream = eTags.isEmpty();
+
             getClient().completeMultipartUpload(new CompleteMultipartUploadRequest(bucket.getName(),
```
andyHa, line 736: could we just abort the empty upload here? has we know for sure that an empty MP upload is pointless anyway? could be use a conventional upload to ensure that the empty object exists?

andyHa (changes requested):

src/main/java/sirius/biz/storage/s3/ObjectStore.java
andyHa: maybe we could just call

```
getClient().abortMultipartUpload(new AbortMultipartUploadRequest(bucket.getName(),
                                                                             objectId,
                                                                             multipartUpload.getUploadId()));
                                                                             ```
                                                                             
                                                                             followed by s.th. like
```
getClient().putObject(bucket.getName(),
                                  objectId,"");
```

src/main/java/sirius/biz/storage/s3/ObjectStore.java
sabieber: Why should the name of the boolean start with `is`? I find this a good rule of thumb for a boolean getter method name, but in the actual variable/field name this is unnecessary noise

src/main/java/sirius/biz/storage/s3/ObjectStore.java
idlira: In the replication event is indeed pointless. But a "normal" upload is the real question: shall we always assume no data as OK? If so, we should than load a 0-byte file. Aborting the transaction means the object id does not exist in the bucket.
Inside ObjectStore, we cannot know that the space is being used for replication (one can even have the idea to chain replications)... 
This is why I decided to keep it non-breaking, preserving the previous behavior and only catch the case when executing a replication.

PR #1525, opened 2022-11-18
Defines a source in import transactions

idlira:
An optional `source` can be passed to the ImportTransactionHelper, which is used as an additional filter upon deleting non-marked entities when using `SyncMode = SYNC`.

Providing no source keeps the previous behavior, and all entities not matching the `txnId` will be deleted.

The standard RelationalEntityImportJob / RelationalEntityImportJobFactory have been extended so a source can be optionally provided, either as a job parameter or as a fixed value.

Fixes: [OX-9276](https://scireum.myjetbrains.com/youtrack/issue/OX-9276)


src/main/java/sirius/biz/importer/txn/ImportTransactionHelper.java
```diff
@@ -72,8 +72,10 @@ public ImportTransactionHelper start() {
      * @return the helper itself for fluent method calls
      */
     public ImportTransactionHelper start(String source) {
-        this.transactionId = System.currentTimeMillis();
-        this.source = source;
+        start();
+        if (Strings.isFilled(source)) {
+            this.source = source.trim();
```
jakobvogel, line 77: What if the `source` is empty after trimming? Do we consider both, `null` and the empty string further down the line?

mkeckmkeck (approved):

jakobvogel (approved):

src/main/java/sirius/biz/importer/txn/ImportTransactionHelper.java
idlira: Yes, we do. The value is only used if Strings.isFilled delivers something.

sabieber (approved):

cschierle (approved):

andyHa (changes requested):

PR #1548, opened 2022-12-08
Documents error codes

cschierle:
Uses defaults from https://github.com/scireum/sirius-web/pull/1139

Fixes: [OX-9330](https://scireum.myjetbrains.com/youtrack/issue/OX-9330)


src/main/java/sirius/biz/cluster/ClusterController.java
```diff
@@ -260,6 +260,9 @@ public void bleed(WebContext webContext, String setting, String node) {
     @ApiResponse(responseCode = "200",
             description = "Successful response",
             content = @Content(mediaType = "text/plain", examples = @ExampleObject("OK")))
+    @ApiResponse(responseCode = "403",
```
idlira, line 263: Can't we use the defaults here as well?

PR #1549, opened 2022-12-09
Conversion: Allows to register additional handlers for conversion failures

cschierle:
If the conversion fails for a variant, the error and the involved blob and variant are now forwarded to error handlers. These can for example be used to perform additional handling of conversion failures in entities referring to the blob or variant.

Fixes: [SIRI-658](https://scireum.myjetbrains.com/youtrack/issue/SIRI-658)


andyHa:
Now that we have several different converters for one blob (videos), I wonder, if a global flag is the way to go. We could simply reduce the change to only propagate the "cannot convert" error to a listener (as currently implemented in the BlobChangeLoop). This can then notify upstream entities (like Attachments) - which can be reset once the blob changes. Storing the flag within the framework isn't actually necessarry at all...

src/main/java/sirius/biz/storage/layer2/FailedVariantConversionListener.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Propagates the conversion error to registered {@link FailedVariantConversionHandler handlers} for further processing.
+     *
+     * @param error     the error causing the conversion failure
+     * @param blobKey   the blob for which the variant should have been created
+     * @param variantId the failed variant
+     */
+    public void propagateError(Throwable error, String blobKey, String variantId) {
+        errorHandlers.forEach(handler -> handler.handle(error, blobKey, variantId));
+    }
```
andyHa, line 38: whoy not move this where it is needed? sounds like an overkill to extract this into its own class...

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

PR #1552, opened 2022-12-12
Adds the capability to create variants of a blob by a newly introduced method

mko-sci:
Fixes: [OX-9323](https://scireum.myjetbrains.com/youtrack/issue/OX-9323/Video-Modernisierung-Optimierung-des-Verarbeitungsjobs-Processing)

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -1411,6 +1411,16 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
     }
 
     /**
+     * Ensures that the desired variant exists.
+     *
+     * @param blob        the blob for which the variant should exist
+     * @param variantName the name of the desired variant
+     */
+    public void ensureVariantExists(B blob, String variantName) {
+        findOrCreateVariant(blob, variantName, false);
+    }
+
+    /**
      * Tries to either find or create the requested variant for the given blob.
      * <p>
      * Note that there this is a recursive optimistic locking algorithm at work where
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -20,6 +20,7 @@ import sirius.biz.storage.layer2.variants.ConversionEngine;
 import sirius.biz.storage.layer2.variants.ConversionProcess;
 import sirius.biz.storage.util.StorageUtils;
 import sirius.db.KeyGenerator;
+import sirius.kernel.async.Future;
 import sirius.kernel.async.Tasks;
 import sirius.kernel.cache.Cache;
 import sirius.kernel.cache.CacheManager;
@@ -1401,7 +1402,7 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
             throw new IllegalArgumentException(Strings.apply("Unknown variant type: %s", variantName));
         }
 
-        V variant = findOrCreateVariant(blob, variantName, nonblocking);
+        V variant = findOrCreateVariantAsync(blob, variantName, nonblocking);
 
         if (variant == null || Strings.isEmpty(variant.getPhysicalObjectKey())) {
             return null;
@@ -1417,15 +1418,15 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
      * @param variantName the name of the desired variant
      */
     public void ensureVariantExists(B blob, String variantName) {
-        findOrCreateVariant(blob, variantName, false);
+        findOrCreateVariant(blob, variantName, false, false);
     }
 
     /**
-     * Tries to either find or create the requested variant for the given blob.
+     * Tries to either find or create the requested variant asynchronously for the given blob.
      * <p>
      * Note that there this is a recursive optimistic locking algorithm at work where
-     * {@link #attemptToFindOrCreateVariant(Blob, String, boolean, int)} and
-     * {@link #awaitConversionResultAndRetryToFindVariant(Blob, String, int)} build the "loop".
+     * {@link #attemptToFindOrCreateVariant(Blob, String, boolean, int, boolean)} and
+     * {@link #awaitConversionResultAndRetryToFindVariant(Blob, String, int, boolean)} build the "loop".
      *
      * @param blob        the blob for which the variant is to be resolved
      * @param variantName the variant of the blob to find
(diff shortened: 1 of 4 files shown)


andyHa (approved):

jakobvogel (approved):

mkeckmkeck (approved):

sabieber (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1410,6 +1410,16 @@ protected String lookupPhysicalKey(String blobKey, @Nonnull String variantName,
         }
     }
 
+    /**
+     * Ensures that the desired variant exists.
+     *
+     * @param blob        the blob for which the variant should exist
+     * @param variantName the name of the desired variant
+     */
+    public void ensureVariantExists(B blob, String variantName) {
```
idlira, line 1419: Why do we need the whole method around, instead of just making findOrCreateVariant public, which even delivers the variant created or obtained?

idlira (changes requested):

src/main/java/sirius/biz/storage/layer2/Blob.java
```diff
@@ -118,6 +118,14 @@ public interface Blob {
      */
     Optional<FileHandle> download();
 
+    /**
+     * Ensures that the desired variant exists.
+     *
+     * @param variantName the name of the desired variant
+     */
+    void ensureVariantExists(String variantName);
```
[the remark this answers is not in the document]
mko-sci: The idea of the name was that this method ensures that the varaint exists, and that is all what the caller should know about the method.

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
mko-sci: I was pair programming with @andyHa and he suggested that we build the method like this. The idea was to shield as much as possible from a caller and only provide a simple mechanism to trigger the variant creating if necessary.

src/main/java/sirius/biz/storage/layer2/Blob.java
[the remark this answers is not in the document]
mko-sci: It might be an good idea to return the variant.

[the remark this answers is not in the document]
idlira: Than is not shielding so much. Leave like this, and we add a new method if needed.
Have no cristal ball, but something tells me we will need it 🧙‍♂️

idlira (approved):

src/main/java/sirius/biz/storage/layer2/Blob.java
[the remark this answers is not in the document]
mko-sci: I agree, I have the same feeling. We probably going to need it for the creation of the poster (preview) images... :D

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1615,14 +1645,16 @@ private V awaitConversionResultAndRetryToFindVariant(B blob, String variantName,
...
     /**
-     * Spawns a thread which will actually invoke the appropriate conversion pipeline.
+     * Spawns a thread which will actually invoke the appropriate conversion pipeline asynchronously.
      *
      * @param blob    the blob for which the variant is to be created
      * @param variant the variant to generate
+     * @return a future holding the conversion process
      */
-    private void invokeConversionPipelineAsync(B blob, V variant) {
+    private Future invokeConversionPipelineAsync(B blob, V variant) {
```
idlira, line 1654: Thinking loud: aren't we making it way too complex? The VideoProcessingJob could call the next 2 lines directly, and handle the Future on its own, leaving stuff like max duration at job level. Even logging, since you than have access to the Future's onFailureCallback and the like.

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1410,6 +1411,16 @@ protected String lookupPhysicalKey(String blobKey, @Nonnull String variantName,
...
+    /**
+     * Ensures that the desired variant exists.
+     *
+     * @param blob        the blob for which the variant should exist
+     * @param variantName the name of the desired variant
+     */
+    public void ensureVariantExists(B blob, String variantName) {
+        findOrCreateVariant(blob, variantName, false);
+    }
+
```
andyHa, line 1423: as we no longer need this, shouldn't we kill it?

andyHa (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
mko-sci: yep we could kill `ensureVariantExists`

sabieber (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1698,6 +1716,68 @@ private boolean tryCreateVariantAsync(B blob, String variantName) {
...
+            if (detectAndRemoveDuplicateVariant(variant, blob, variantName)) {
+                // An optimistic lock error occurred (another thread or node attempted the same). So we backup,
+                // wait a short and random amount of time and retry...
+                Wait.randomMillis(0, 150);
+                // the variant does already exist
+                return tryCreateVariant(blob, variantName, retries - 1);
+            } else {
+                return invokeConversionPipelineAsync(blob, variant);
+            }
```
fhaScireum, line 1755: Sollte der Kommentar "the variant does already exist" nicht im else-Zweig stehen?

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1443,6 +1444,28 @@ protected V findOrCreateVariant(B blob, String variantName, boolean nonblocking)
...
+     * @param variantName the variant of the blob to find
+     * @return the variant for the given blob with the given name or null if no such variant exists
+     */
+    private V tryFetchVariant(B blob, String variantName) {
+        V variant = findAnyVariant(blob, variantName);
+        if (variant != null
+            && !variant.isQueuedForConversion()
+            && retryLimitReached(variant)
+            && Strings.isEmpty(variant.getPhysicalObjectKey())) {
+            // The conversion has failed - signal that the to client. We use a handled exception here, as the problem
```
fhaScireum, line 1460:
```suggestion
            // The conversion has failed - signal that to the client. We use a handled exception here, as the problem
```

sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
mko-sci: nope. `detectAndRemoveDuplicateVariant` - Detects if the variant is still unique. Note that this is kind of a negative approach - if the method returns <tt>true</tt> a collision was detected and the given variant was removed. The <tt>false</tt> case is the expected one which verified that the variant is unique and good to go.

mko-sci: I improved the comment

fhaScireum (approved):

PR #1556, opened 2022-12-16
Extracts common interface for blob change handlers

cschierle:
Needs changes from #1549

Fixes: [OX-9285](https://scireum.myjetbrains.com/youtrack/issue/OX-9285)


src/main/java/sirius/biz/storage/layer2/BlobConvertibilityChangedHandler.java
```diff
@@ -0,0 +1,27 @@
...
+import sirius.kernel.di.std.Priorized;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Defines handlers to process {@link Blob blobs} after their {@link Blob#isInconvertible() convertibility status} changed.
+ *
+ * @see ProcessBlobChangesLoop
+ */
+public interface BlobConvertibilityChangedHandler extends Priorized {
```
andyHa, line 20: extract a super-interface for all BlobXXXChangedHandlers?

src/main/java/sirius/biz/storage/layer2/BlobConvertibilityChangedHandler.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package sirius.biz.storage.layer2;
+
+import sirius.kernel.di.std.Priorized;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Defines handlers to process {@link Blob blobs} after their {@link Blob#isInconvertible() convertibility status} changed.
```
andyHa, line 16: add a remark that handlers must be registered....

src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlob.java
```diff
@@ -208,6 +208,12 @@ public class SQLBlob extends SQLEntity implements Blob, OptimisticCreate {
     public static final Mapping INCONVERTIBLE = Mapping.named("inconvertible");
     private boolean inconvertible;
 
+    /**
+     * Stores if the convertibility of the blob changed, e.g. after updating the physical file.
+     */
```
andyHa, line 213:
```suggestion
     * Stores if the convertibility of the blob changed, e.g. after updating the physical file.
     * <p>
     * We store an extra flag here, which is internally toggled and then propagated via the
     * <tt>ProcessBlobChangesLoop</tt>
     * 
     */
```

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlob.java
```diff
@@ -239,13 +245,18 @@ protected void beforeSave() {
             createdOrRenamed = true;
         }
 
+        if (isNew() || isChanged(CONVERTIBILITY_CHANGED)) {
```
sabieber, line 248: Will this not result in the field being set back to true when it is set to false via `mango.update`?

src/main/java/sirius/biz/storage/layer2/mongo/MongoBlob.java
```diff
@@ -246,13 +252,18 @@ protected void beforeSave() {
             createdOrRenamed = true;
         }
 
+        if (isNew() || isChanged(CONVERTIBILITY_CHANGED)) {
```
sabieber, line 255: s.a.

cschierle:
Replaced by https://github.com/scireum/sirius-biz/pull/1562

PR #1567, opened 2023-01-23
SE-12352: Tenant Permission for Log Limit Job Parameter

fhaScireum:
Requires a tenant permission for showing the limit log message parameter.

If the tenant does not have the permission, the parameter isn't shown and thus the default behavior of limiting messages is applied.

Fixes: SE-12352


idlira (approved):

src/main/resources/component-biz.conf
```diff
@@ -1357,8 +1358,7 @@ security {
     subScopes = ${?security.subScopes} ["ui", "api", "vfs", "empty"]
 
     # Defines all tenant permissions (aka features). This also can be redefined by the application.
-    # Init to an empty array, in case it is not redefined somewhere else
-    tenantPermissions = ${?security.tenantPermissions} []
+    tenantPermissions = ${?security.tenantPermissions} ["permission-bypass-process-log-limits"]
```
andyHa, line 1361: shouldn't we name this tenant-permission-XX or the like to not confuse this with a user permisson? or a feature- or whatever we use elsewhere...

fhaScireum: Do we want to have an intermediate permission that gives you the actual permission that is being checked or should the existing permission simply be called "tenant-permission-bypass-process-log-limits"?

We do not have an existing naming scheme for tenant permissions yet in the sirius framework. S2 and OX have completely different naming schemes.

sabieber (approved):

src/main/resources/component-biz.conf
```diff
@@ -1320,23 +1320,24 @@ security {
...
+        permission-system-journal            : "Required to view the system journal"
+        permission-select-tenant             : "Required to switch to another tenant"
+        permission-select-user-account       : "Required to switch to another user"
+        permission-execute-jobs              : "Required to execute jobs"
+        permission-manage-scheduler          : "Required to plan and edit scheduled jobs"
+        permission-system-scripting          : "Required to execute administrator level scripts on all nodes of the cluster"
+        permission-view-files                : "Required to view files stored in the system"
+        permission-view-processes            : "Required to view processes"
+        permission-view-audit-log            : "Required to view the audit log"
+        permission-bypass-process-log-limits : "Required to bypass log limits of processes via a job parameter"
```
cschierle, line 1340: Also rename here?
```suggestion
        feature-bypass-process-log-limits : "Required to bypass log limits of processes via a job parameter"
```

cschierle (approved):

ymo-sci (approved):

andyHa (approved):

PR #1574, opened 2023-01-28
Provides various improvements for import jobs.

andyHa:
- Handling and management of side-tasks has been improved
- Handling of the ErrorContext has been made thread-safe
- Some overly complex facilities have been deprecated
- A generic errorhandling for ImportHandlers (esp. tryFind) has been added


andyHa:
As a migration guide:
- Mainly, ErrorContext should be obtained properly
- Check existing imports and import handlers, if the newly created fail/warn methods can be used
- Check that only the non-deprecated hints are used within import jobs

idlira (approved):

PR #1584, opened 2023-02-02
Changes the way global "batches" are executed.

andyHa:
For global metric tasks, we used abstract classes like MongoEntity or SQLEntity and created a "fake" batch which was empty otherwise. Additionally, in some cases - namely 1) in case of an exception and 2) in case of no entities present at all - no batch was generated at all.

However, having multiple levels we actually rely on at least one batch being present, so that we can add the LAST_BATCH marker and thus scheduler the next
level. So, for global runs, we now no longer use an completely empty batch description, but one with a special flag being set. Additionally, in the cases named above, we emit a completely empty batch - which is used to forward the LAST_BATCH marker but is then safely discarded.

### DIFF
--- a/src/main/java/sirius/biz/analytics/scheduler/SQLAnalyticalTaskScheduler.java
+++ b/src/main/java/sirius/biz/analytics/scheduler/SQLAnalyticalTaskScheduler.java
@@ -84,7 +84,7 @@ public abstract class SQLAnalyticalTaskScheduler extends BaseAnalyticalTaskSched
     @Override
     public void executeBatch(JSONObject batchDescription, LocalDate date, int level) {
         if (batchDescription.containsKey(BaseAnalyticalTaskScheduler.CONTEXT_MARKER_GLOBAL_ENTITY)) {
-            // A global run was detected, execute with "null" and MongoEntity as type...
+            // A global run was detected, execute with "null" and SQLEntity as type...
             executeEntity(null, SQLEntity.class, date, level);
         } else if (batchDescription.containsKey(BaseEntityBatchEmitter.TYPE)) {
             // Note that we check for the presence of the TYPE, as (in case of some schedulers) we emit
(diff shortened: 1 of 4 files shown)


jakobvogel (approved):

sabieber (approved):

src/main/java/sirius/biz/analytics/scheduler/SQLAnalyticalTaskScheduler.java
```diff
@@ -80,10 +83,16 @@ protected boolean isMatchingEntityType(AnalyticalTask<?> task) {
 
     @Override
     public void executeBatch(JSONObject batchDescription, LocalDate date, int level) {
-        if (!batchDescription.containsKey(BaseEntityBatchEmitter.TYPE)) {
+        if (batchDescription.containsKey(BaseAnalyticalTaskScheduler.CONTEXT_MARKER_GLOBAL_ENTITY)) {
+            // A global run was detected, execute with "null" and MongoEntity as type...
```
ymo-sci, line 87:
```suggestion
            // A global run was detected, execute with "null" and SQLEntity as type...
```

src/main/java/sirius/biz/analytics/scheduler/SQLAnalyticalTaskScheduler.java
```diff
@@ -39,9 +39,12 @@ protected Class<?> getMinimalTargetType() {
     @Override
     protected void scheduleBatches(Class<? extends SQLEntity> type, Consumer<JSONObject> batchConsumer) {
         try {
-            if (Modifier.isAbstract(type.getModifiers()) || mixing.findDescriptor(type).isEmpty()) {
-                batchConsumer.accept(new JSONObject());
-            } else {
+            if (Modifier.isAbstract(type.getModifiers())) {
+                // We use MongoEntity.class as place-holder to run global metric computers.
```
ymo-sci, line 43:
```suggestion
                // We use SQLEntity.class as place-holder to run global metric computers.
```

ymo-sci (approved):

idlira (approved):

PR #1592, opened 2023-02-06
Outputs the counts in query tool in a more readable form

sabieber:
image

### DIFF
--- a/src/main/java/sirius/biz/model/QueryController.java
+++ b/src/main/java/sirius/biz/model/QueryController.java
@@ -23,6 +23,7 @@ import sirius.db.mixing.query.constraints.Constraint;
 import sirius.db.mongo.MongoEntity;
 import sirius.db.mongo.MongoQuery;
 import sirius.kernel.Sirius;
+import sirius.kernel.commons.Amount;
 import sirius.kernel.commons.NumberFormat;
 import sirius.kernel.commons.Strings;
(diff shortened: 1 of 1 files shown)


qw3ry (approved):

fhaScireum (approved):

ymo-sci (approved):

src/main/java/sirius/biz/model/QueryController.java
```diff
@@ -124,8 +126,8 @@ public void query(WebContext webContext) {
 
                 UserContext.message(Message.info()
                                            .withTextMessage(Strings.apply("Showing %s of %s results - Query took %sms",
-                                                                          result.size(),
-                                                                          numberOfEntities,
+                                                                          Amount.of(result.size()).toRoundedString(),
+                                                                          Amount.of(numberOfEntities).toRoundedString(),
```
andyHa, line 130: NLS.toUserString could be used as well (?)

andyHa (approved):

src/main/java/sirius/biz/model/QueryController.java
sabieber: That was my first try and didn't work

sabieber: Or do you mean wrapping it in an Optional and then calling toUserString on that?

idlira (approved):

PR #1595, opened 2023-02-06
Displays the canceled date when the job has not been completed properly (and therefore completed has not been set)

mko-sci:
Fixes: [OX-9511](https://scireum.myjetbrains.com/youtrack/issue/OX-9511)


mko-sci:
> 

Good question. It looks like it has been done intentionally. I was about to add the completion date on cancel, but then I discovered that the canceledDate is set there. Which is why I assumend that this is the way here...

andyHa:
As we seem to somehow rely on this logic, I'd keep it for now. If we ever deciede to fill the date, the rendered result would be the same in this case...

PR #1598, opened 2023-02-07
Adds all performance flags of a tenant as virtual roles of their users.

andyHa:
Each flag is made available as role "performance-flag-FLAG_NAME". This greatly simplifies checking if a certain flag is set.


src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -152,6 +153,21 @@
...
+    /**
+     * Contains the prefix added to the account number which is added as artificial role.
+     * <p>
+     * Therefore, each user of a tenant has the following additional role (if filled):
+     * {@code ROLE_PREFIX_ACCOUNT_NUMBER + tenant.getTenantData().getAccountNumber()}
+     */
+    public static final String ROLE_PREFIX_ACCOUNT_NUMBER = "tenant-";
+
+    /**
+     * Contains the prefix added to all performance flag when they're made visible as role.
```
sabieber, line 165:
```suggestion
     * Contains the prefix added to all performance flags when they're made visible as roles.
```

sabieber (approved):

mko-sci (approved):

PR #1601, opened 2023-02-08
Exposes delegateUrl method and constant.

cschierle:
This allows to request conversions from any node without waiting for a Response, as is needed in deliverAsync(). By invoking the delegate URL e.g. via an Outcall, the conversion is properly delegated to conversion-entitled nodes.

Fixes: [OX-9491](https://scireum.myjetbrains.com/youtrack/issue/OX-9491)


cschierle:
found another solution that does not require exposure

PR #1603, opened 2023-02-10
Returns null for empty StringParameter

idlira:
This basically restores the previous behavior. If no default was returned, and the parameter is empty (either not in context or provided blank) we want to return a null string. Reason for that is that the output is often used in DB queries where null is expected instead of an empty string.

Fixes: [SIRI-713](https://scireum.myjetbrains.com/youtrack/issue/SIRI-713)

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/StringParameter.java
+++ b/src/main/java/sirius/biz/jobs/params/StringParameter.java
@@ -48,6 +48,12 @@ public class StringParameter extends TextParameter<String, StringParameter> {
         if (input.getRawString() == null) {
             return defaultValue;
         }
+
+        // Always return null when no default value was used and the input is empty or null.
+        if (input.isEmptyString()) {
+            return null;
+        }
+
         return input.getString();
     }
 


fhaScireum (approved):

sabieber (approved):

andyHa (approved):

src/main/java/sirius/biz/jobs/params/StringParameter.java
```diff
@@ -48,6 +48,12 @@ protected String checkAndTransformValue(Value input) {
         if (input.getRawString() == null) {
             return defaultValue;
         }
+
+        // Always return null when no default value was used and the input is empty or null.
+        if (input.isEmptyString()) {
```
jakobvogel, line 53: In the comment, you say: "if the input is null" ... Can `input` itself be `null` here hence causing a NPE, or did you mean the payload of it?

jakobvogel (approved):

src/main/java/sirius/biz/jobs/params/StringParameter.java
idlira: Bad choice of words 😔 
input as _user input_, and not input as the actual variable (which is a Value object and in this context never null)

PR #1610, opened 2023-02-14
BREAKING: adds a description to the file and auxFileMode parameter

idlira:
The FileImportJob's `file` parameter has now a default description. That can be overwritten in the FileImportJobFactory or by jobs calling `FileImportJobFactory.createFileParameter()` directly. For the later, we mark it as breaking so custom jobs eventually borrowing this parameter's creation shall pass the description.

A default description has also been added to the `auxFileMode` parameter.

Fixes: [SIRI-717](https://scireum.myjetbrains.com/youtrack/issue/SIRI-717)


src/main/resources/biz_de.properties
```diff
@@ -337,6 +338,7 @@ FileImportJobFactory.auxFileMode = Zusätzliche Dateien
 FileImportJobFactory.auxFileParentDirectory = Basispfad
 FileImportJobFactory.auxFileParentDirectory.help = Definiert einen Ordner innerhalb des Zielverzeichnisses, in den die neuen Dateien hochgeladen werden. Verwenden Sie '/' für einen Pfad mit Unterverzeichnissen. Der Ordner wird nötigenfalls angelegt.
 FileImportJobFactory.file = Datei
+FileImportJobFactory.file.help = Geben Sie hier eine Datei oder eine Archivdatei (z.B. zip) zur Verarbeitung an. Sie können eine existierende Datei aus dem Work-Verzeichnis wählen oder einfach eine Datei per Drag&Drop auf den Job ziehen.
```
sabieber, line 341:
```suggestion
FileImportJobFactory.file.help = Geben Sie hier eine Datei oder eine Archivdatei (z.B. ZIP) zur Verarbeitung an. Sie können eine existierende Datei aus dem Work-Verzeichnis wählen oder einfach eine Datei per Drag&Drop auf den Job ziehen.
```

sabieber (approved):

sabieber (approved):

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
```diff
@@ -183,10 +182,15 @@ private static Boolean hasNoArchiveFile(Map<String, String> params) {
      * Helps to create a custom file parameter with an appropriate file extension filter.
      *
      * @param acceptedFileExtensions the file extensions to accept
+     * @param description            the description to use in the parameter
      * @return the parameter used to select the import file
      */
-    public static Parameter<VirtualFile> createFileParameter(@Nullable List<String> acceptedFileExtensions) {
+    public static Parameter<VirtualFile> createFileParameter(@Nullable List<String> acceptedFileExtensions,
+                                                             String description) {
```
mkeckmkeck, line 189: Maybe switch the nullable annotation here? Or completely remove it? My expectation is that there is no need for collection parameters to be null at API level. Internal robustness remains of course :-)

mkeckmkeck (approved):

andyHa (approved):

src/main/java/sirius/biz/jobs/batch/file/FileImportJob.java
idlira: I opt for the first. When collecting, makes definitely no sense. But the creation method is also used in job instances, where its goal is just to obtain the parameter's value, and in there we need no description.

PR #1611, opened 2023-02-15
Tycho: Code Style for process menu icon

cschierle:


src/main/resources/default/extensions/tycho-page-menu/menu.html.pasta
```diff
@@ -27,15 +27,13 @@
...
-                <t:menuImageItem class="d-none d-xxl-inline"
-                                 icon="/images/icons/ps.png"
-                                 activeIcon="/images/icons/ps-notif.png"
-                                 active="part(sirius.biz.process.Processes.class).hasActiveProcesses()"
-                                 link="/ps"
-                                 permission="permission-view-processes"/>
-            </i:if>
-
+            <t:menuImageItem framework="biz.processes"
+                             class="d-xxl-inline"
```
jakobvogel, line 31: Why is `d-none` and `d-xxl-inline` conflicting? We use something similar at other places as well, and it is supposed to mean `display: none` by default apart from large screens where it becomes `display: inline`.

cschierle: Thank you! You are correct 👍
The styles are not the issue, but the different behavior between tycho and wondergem pages. Since `d-xxl-inline` is a tycho style, it cannot be applied in the same way for wondergem. I got a bit confused by that.

I guess, this is just a small Code Style PR now. And I'll look into a solution to unify the behavior separately

jakobvogel (approved):

PR #1613, opened 2023-02-16
Misc optimizations for Blob / BlobDispatcher and URLBuilder

andyHa:
Please revise the individual commints for elaborate change descriptions.

This main tipping point for this refactoring was: [SIRI-725](https://scireum.myjetbrains.com/youtrack/issue/SIRI-725)

Breaking changes:

Some methods and return types in URLBuilder have been removed/renamed:
- eternallyValid() is replaced by reusable()
- delayedResolved() is replaced by delayedResolve()
- The url in URLBuilder.UrlResult is now non-optional as its value is dictated by the UrlType already

For Blob:
- deliver() was removed as all traffic is sent through the BlobDispatcher


jakobvogel (approved):

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
```diff
@@ -384,19 +388,37 @@ private void virtualDelivery(WebContext request,
...
         Response response = request.respondWith();
+
         if (cacheable) {
-            response.cached();
+            // If a virtual request is marked as cacheable, we try to redirect to the proper physical blob key
+            // as this will remain in cache much longer (and the redirect itself will also be cached). The additional
+            // HTTP round-trip for the redirect shouldn't hurt too much, as it is most probably optimized away due to
+            // keep-alive. However, using a physical delivery with infinite cache settings will enable any downstream
+            // reverse-proxies to maximize their cache utilization...
+            if (attemptRedirectToPhysical(response,
```
idlira, line 403: Just a design topic: I don't like the approach of changing the response inside the method (makes reading it more complex). Would make the method return the url, and if non empty set the response/redirect hier.

src/main/java/sirius/biz/storage/layer2/Blob.java
```diff
@@ -134,7 +134,10 @@ public interface Blob {
      * Delivers the data of this blob into the given HTTP response.
      *
      * @param response the HTTP response to populate
+     * @deprecated use {@link URLBuilder} and let the {@link BlobDispatcher} perform the heavy lifting. If a response
+     * is to be fulfilled, redirect to the URL generated by the url builder...
      */
+    @Deprecated
```
idlira, line 140: Kill right now?

src/main/java/sirius/biz/storage/layer2/URLBuilder.java
```diff
@@ -243,21 +242,25 @@ public URLBuilder withBaseURL(String baseURL) {
      * Note that this implies that a {@link #reusable() reusable} URL is generated.
      *
      * @return the builder itself for fluent method calls
+     * @deprecated can be replaced by {@link #reusable()} - which should have been invoked in the first place and which
+     * now automatically generates eternally valid URLs.
      */
+    @Deprecated
```
idlira, line 276: Kill right now?

src/main/java/sirius/biz/storage/layer2/URLBuilder.java
```diff
@@ -269,6 +269,14 @@ public URLBuilder reusable() {
         return this;
     }
 
+    /**
+     * @deprecated use {@link #delayResolve()} which is the same method except for the typo in the name
+     */
+    @Deprecated
```
idlira, line 303: Kill right now?

idlira (changes requested):

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
```diff
@@ -411,4 +429,31 @@ private void virtualDelivery(WebContext request,
                              response,
                              request::markAsLongCall);
     }
+
+    private boolean attemptRedirectToPhysical(@Nullable Response response,
```
sabieber, line 433: The Response parameter is not actually nullable in the Code below

src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
```diff
@@ -481,6 +482,28 @@ public void deliver(Response response, String objectId, boolean largeFileExpecte
...
+    /**
+     * Determines if a special delivery for very large files should be attempted.
+     * <p>
+     * For very large files (multiple GBs), we use special URLs generated by the
+     * {@link sirius.biz.storage.layer2.URLBuilder} and handled by the {@link sirius.biz.storage.layer2.BlobDispatcher}.
+     * These contain the "/xxl/" marker so that downstream caches like <tt>Varnish</tt> don't even attempt to put the
+     * data in their cache. These files therefore are piped through which may lead to increasingly large internal
+     * buffers as an upstream storage like <b>S3</b> might send data faster than we pump through to slower clients
+     * like mobile users. We therefore use a blocking/stream based approach in a thread pool of limited size. However,
+     * in the presence of a <tt>Range</tt> header, the date to transfer is most probably way smaller and also the stream
```
sabieber, line 494:
```suggestion
     * in the presence of a <tt>Range</tt> header, the data to transfer is most probably way smaller and also the stream
```

src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
```diff
@@ -481,6 +482,28 @@ public void deliver(Response response, String objectId, boolean largeFileExpecte
...
+     * implementation cannot satisfy such requests. We therefore only use this approach is a large file is expected and
```
sabieber, line 495:
```suggestion
     * implementation cannot satisfy such requests. We therefore only use this approach if a large file is expected and
```

src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
```diff
@@ -481,6 +482,28 @@ public void deliver(Response response, String objectId, boolean largeFileExpecte
...
+     * no <tt>Range</tt> header is present.
+     *
+     * @param response          the response to handle
+     * @param largeFileExpected a flag which determines if a large file is expected at all
+     * @return <tt>true</tt> if the delivery should be handled in a separate thread pool and a non-blocking fashion,
+     * <tt>false</tt> otherwise
```
sabieber, line 501:
```suggestion
     * @return <tt>true</tt> if the delivery should be handled in a separate thread pool and a non-blocking fashion
```
In my opinion this its redundant to tell that a boolean parameter can also be false :)

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
sabieber: In general I don't like the pattern of altering a parameter inside a method, too. But in the case of the response here it is not really "altering" the response but completing it?
But I also agree that keeping all actually response calls in this method and not extract them away would simplify understanding what can be the response we send out.

src/main/java/sirius/biz/storage/layer2/URLBuilder.java
```diff
@@ -76,10 +76,10 @@ public class URLBuilder {
      * Contains the {@linkplain #buildUrlResult() generated URL} and an {@linkplain UrlType indicator} if it is a
      * virtual, physical, the fallback URL, or empty.
      *
-     * @param url     the URL generated by this url builder wrapped as an Optional
+     * @param url     the URL generated by this url builder
      * @param urlType the type of URL that was generated
      */
-    public record UrlResult(Optional<String> url, UrlType urlType) {
+    public record UrlResult(String url, UrlType urlType) {
```
sabieber, line 82: This is a breaking change? Please list it in the PR description

sabieber (approved):

jakobvogel (approved):

idlira (approved):

mko-sci (approved):

PR #1617, opened 2023-02-17
Adds an alternative fallback image for imageUrls which indicates that the variant conversion has finally failed and its logic to display it in such cases

mko-sci:
Bildschirm­foto 2023-02-17 um 22 59 39

Bildschirm­foto 2023-02-17 um 22 59 58

Fixes: [SIRI-704](https://scireum.myjetbrains.com/youtrack/issue/SIRI-704/Tycho-tblobImage-Failed-Img-einbauen-welches-signalisiert-das-kein-Vorschaubild-da-ist-weil-ein-Fehler-aufgetreten-ist-bzw-keins)


src/main/java/sirius/biz/storage/layer2/URLBuilder.java
```diff
@@ -405,6 +411,29 @@ public boolean isConversionExpected() {
         return isFilled() && !Strings.areEqual(variant, VARIANT_RAW) && Strings.isEmpty(determinePhysicalKey());
     }
 
+    /**
+     * Determines if the conversion of the given variant has finally failed and therefore no further conversion attempts are made.
+     *
+     * @return <tt>true</tt> if the variant is failed, otherwise <tt>false</tt>.
+     */
+    public boolean isConversionFailed() {
```
jakobvogel, line 419: Sounds grammatically strange. I would prefer `hasConversionFailed()`, but that is just an opinion. Also, `is…` is a getter for booleans, and this is no getter.

jakobvogel (approved):

src/main/java/sirius/biz/storage/layer2/URLBuilder.java
mko-sci: yep, you are right, right above this method is another method with the same naming patter, which is why I setteled on it...but I consider renaming it =)

jakobvogel: "Des hamm mer immer schon so gemacht" war noch nie ein gutes Argument 😉😅

mko-sci: Das ist richtig, wollte nur nix renamend und es irgendwie harmonisch haben

cschierle (commented):
Seems like we need many lookups for the simple benefit of the different fallback image. Not sure if the change is worth it like this.
Needs further discussion.

andyHa:
can't we either handle this via the existing cache or on the JS side which receives the response from the BlobDispatcher and could then swap/use the proper image. This seems to add an overhead to every URL being generated - which is wayyy to heavy..

PR #1624, opened 2023-02-24
CodeListLookupTable: Returns emtpy stream instead of throwing an exception.

cschierle:
This matches the behavior of other not yet implemented methods of this class.

Fixes: OX-9525


cschierle:
Found a different solution for the issue of the referenced Ticket, which does not require changes to the framework

PR #1633, opened 2023-03-01
Metrics: adds helper to sum up values

cschierle:


sabieber (approved):

idlira (approved):

mko-sci (approved):

andyHa:
We need to clarify if we really need this, as the is a very inefficient way of computing sums.

cschierle:
will solve the underlying issue by other means

PR #1635, opened 2023-03-02
adjust css for menu size on small widths

ymo-sci:
increased the fa search-icon size in the header on big widths and decreased the icon size of skip and the system-view on small widths. also removed destroying left padding and border on small width and move the breakpoint (see screenshot)
before:
Bildschirm­foto 2023-03-02 um 07 41 56

after:
Bildschirm­foto 2023-03-02 um 07 39 56

### DIFF
--- a/src/main/resources/default/assets/styles/menu.scss
+++ b/src/main/resources/default/assets/styles/menu.scss
@@ -16,14 +16,31 @@ li.right-menu-icon {
 
 }
 
-@media (max-width: 768px) {
+@media (max-width: 991px) {
   // Make fa icons the same size as our provided pngs in the collapsed (hamburger) menu...
-  .right-menu-icon svg{
-    width: 1.5rem !important;
+  li.right-menu-icon {
+    padding-left: 0;
+    border-right: none !important;
+
+    svg {
+      width: 1.5rem !important;
+    }
+  }
+  // Do not oversize the user icon in the collapsed (hamburger) menu...
+  li.right-large-menu-icon {
+    img {
+      height: 1.5rem;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/styles/menu.scss
```diff
@@ -16,14 +16,31 @@ li.right-menu-icon {
 
 }
 
-@media (max-width: 768px) {
+@media (max-width: 991px) {
```
andyHa, line 19: These numbers and the ones below look sort of strange (off by one)?

src/main/resources/default/assets/styles/menu.scss
```diff
@@ -16,14 +16,31 @@ li.right-menu-icon {
...
+  // Do not oversize the user icon in the collapsed (hamburger) menu...
+  li.right-large-menu-icon {
+    img {
+      height: 1.5rem;
+    }
   }
 }
 
-@media (min-width: 768px) {
+@media (min-width: 992px) {
```
andyHa, line 37: s.a.

andyHa (approved):
The PR is basically fine - as pointed out as inline comment some numbers look a bit random.

src/main/resources/default/assets/styles/menu.scss
ymo-sci: ...will fit, otherwise the borders are missing at exactly 991px 🙈

mkeckmkeck (approved):

fhaScireum (approved):

sabieber (approved):

idlira (approved):

cschierle (approved):

PR #1641, opened 2023-03-07
Adds the 'asInline' method which allows one to build an inline download url using the given filename

mko-sci:
Fixes: [OX-9698](https://scireum.myjetbrains.com/youtrack/issue/OX-9698/Modernisierung-des-DirectPdfService)


mkeckmkeck (approved):

idlira (commented):
At least refer in both asInline and asDownload, that they shall not be used together?

A solution with `.withFileName(path)` + `.markAsDownload()` would be cleaner, but breaking...

cschierle (approved):

PR #1642, opened 2023-03-08
Allows the user to enter a URL as source in blobImageHardRefField

bkositza:
Currently, the user can select an image from his directory and upload it. This commit also allows adding resources from a URL. Only the URL is saved, but not its content is uploaded.


src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,106 @@
...
+        container.classList.add("popover-content-max");
+        triggerElement.parentElement.appendChild(container);
+        return container;
+    }
+
+    function buildBody() {
+        const container = document.createElement("div");
+        container.classList.add("row");
+
+        // Input field
```
andyHa, line 35: line comment => extract method with proper name...

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,106 @@
...
+        inputLabel.innerHTML = i18n.label;
+
+        const inputField = document.createElement("input");
+        inputField.classList.add("form-control");
+        inputField.setAttribute("type", "text");
+
+        formGroup.append(inputLabel, inputField);
+        inputCol.appendChild(formGroup);
+
+        // Button OK
```
andyHa, line 52: s.a.

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,106 @@
...
+        buttonOk.classList.add("btn", "btn-block", "btn-primary", "btn-apply");
+        buttonOk.innerHTML = i18n.ok + ' <i class="fa fa-check"></i>';
+        buttonOk.addEventListener("click", function(event) {
+            event.preventDefault();
+            updateURL(inputField.value);
+        })
+
+        leftCol.appendChild(buttonOk);
+
+        // Button cancel
```
andyHa, line 66: s.a.

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,106 @@
...
+        formGroup.append(inputLabel, inputField);
+        inputCol.appendChild(formGroup);
+
+        // Button OK
+        const leftCol = document.createElement("div");
+        leftCol.classList.add("col", "col-12", "col-md-6");
+
+        const buttonOk = document.createElement("button");
+        buttonOk.classList.add("btn", "btn-block", "btn-primary", "btn-apply");
+        buttonOk.innerHTML = i18n.ok + ' <i class="fa fa-check"></i>';
```
andyHa, line 58: use FA 6 names... ;-)

bkositza:
Closed because a URL on a hardRefField does not work. First, a softRefField component must be created. There this feature is then better placed.

PR #1648, opened 2023-03-09
Fixes the new inline download not working properly due to missing filename in URL and missing BlobDispatcher handling

mko-sci:
Fixes: [SIRI-749](https://scireum.myjetbrains.com/youtrack/issue/SIRI-749/URLBuilder-Inline-Download-handling-ermoglichen)


idlira (approved):

src/main/java/sirius/biz/storage/layer2/URLBuilder.java
```diff
@@ -507,6 +505,14 @@ private UrlResult createPhysicalDeliveryUrlResult() {
         return new UrlResult(result.toString(), UrlType.PHYSICAL);
     }
 
+    private void appendFilename(StringBuilder sb, String physicalKey, String filename) {
+        sb.append(physicalKey);
+        sb.append("/");
+        appendAddonText(sb);
+        sb.append(Strings.urlEncode(Files.toSaneFileName(filename)
+                                         .orElse(physicalKey + fetchUrlEncodedFileExtension())));
+    }
```
sabieber, line 514:
```suggestion
    private void appendFilename(StringBuilder result, String physicalKey, String filename) {
        result.append(physicalKey);
        result.append("/");
        appendAddonText(result);
        result.append(Strings.urlEncode(Files.toSaneFileName(filename)
                                         .orElse(physicalKey + fetchUrlEncodedFileExtension())));
    }
```

sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
```diff
@@ -181,12 +181,26 @@ public DispatchDecision dispatch(WebContext request) throws Exception {
         Values uriParts = Values.of(uri.split("/"));
         String type = uriParts.at(0).asString();
         if (Strings.areEqual(type, PHYSICAL_DELIVERY) && uriParts.length() == 5) {
-            String filename = stripAdditionalText(uriParts.at(4).asString());
+            String physicalKey = stripAdditionalText(uriParts.at(4).asString());
```
bkositza, line 184: Would it perhaps be possible to eliminate the MagicNumbers? Wherever uriParts.at() is used a descriptive constant might be helpful.

bkositza (changes requested):

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
```diff
@@ -181,12 +181,26 @@ public DispatchDecision dispatch(WebContext request) throws Exception {
         Values uriParts = Values.of(uri.split("/"));
         String type = uriParts.at(0).asString();
         if (Strings.areEqual(type, PHYSICAL_DELIVERY) && uriParts.length() == 5) {
```
bkositza, line 183: What is the difference between a length of 5 or 6? Constants could also be helpful here.

bkositza (changes requested):

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
mko-sci: Also please read my comment above.
The main difference is the length of the URL or in otherwords the number of parts of which the URL consists of

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
bkositza: Would appreciate it if you create the constants as you described. Only the ones that are really used should probably be enough or?

bkositza (changes requested):
As answered in the comment, I would appreciate it if you create the constants as described by you.

src/main/java/sirius/biz/storage/layer2/BlobDispatcher.java
mko-sci: I agree

bkositza (approved):

PR #1649, opened 2023-03-09
Adds new tycho tag for uploading a softRef

bkositza:
Analogous to the Tycho tag for a hardRef there is now a Tycho tag for a softRef with the possibility to specify a URL as source in addition to a file.

Fixes: [SIRI-740](https://scireum.myjetbrains.com/youtrack/issue/SIRI-740)

image
image
image


src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,139 @@
...
+     * Updates the image in the upload box and the url in the form.
+     *
+     * @param url the new URL
+     */
+    const updateURL = function (url) {
+        if (url.startsWith('http://') || url.startsWith('https://')) {
+            let img = uploadContainer.querySelector('.img-preview img');
+            img.setAttribute('src', url);
+            img.onerror = function() {
+                img.setAttribute('src', '/assets/frontend/defaulticons/default-item.png');
```
jmuscireum, line 112: This is an asset from SellSite, maybe use `/assets/images/blob_image_failed.png` instead

src/main/resources/default/taglib/t/blobImageSoftRefField.html.pasta
```diff
@@ -0,0 +1,124 @@
...
+
+                });
+            });
+
+            if(___urlAllowed) {
+                initUrlUploadPopover(urlButton, outerDiv, resetButton, {
+                    label: '@i18n("StoredObject.URL")',
+                    ok: '@i18n("NLS.ok")',
+                    cancel: '@i18n("NLS.cancel")',
+                    errorMsg: '@i18n("StoredObject.specifyURL")'
```
jmuscireum, line 105: Create a new NLS key for a better error message

src/main/resources/default/taglib/t/blobImageSoftRefField.html.pasta
```diff
@@ -0,0 +1,124 @@
...
+                        img.src = json.downloadUrl;
+                        resetButton.classList.remove('d-none');
+                    });
+
+                });
+            });
+
+            if(___urlAllowed) {
+                initUrlUploadPopover(urlButton, outerDiv, resetButton, {
+                    label: '@i18n("StoredObject.URL")',
```
jmuscireum, line 102: StoredObject is legacy, maybe copy the StoredObject.* NLS keys

jmuscireum (changes requested):

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * Uses jQuery to init bootstraps popover.
+ *
+ * @param triggerElement The element (e.g. Button) that triggers the popover to enter an URL
+ * @param uploadContainer The container the triggerElement and other components belongs to
+ * @param resetButton The button that resets the upload field
+ * @param i18n Contains information about the label, the ok-Button and the cancel-Button of the popover
+ */
+function initUrlUploadPopover(triggerElement, uploadContainer, resetButton, i18n) {
+    $(triggerElement).popover({
```
idlira, line 14: I thought we want to get away from jquery

idlira (changes requested):
Dummy question: why do we need a specific template for BlobSoftRef? 
Perhaps it would haben been easier to understand if this was split in 2 commits: ones copies the hardref into softref, the next applies the deltas.

bkositza:
> Dummy question: why do we need a specific template for BlobSoftRef? 

There are several reasons for this. A softRef can also usefully hold a URL. Furthermore, a softRef will not cope with a temporary file, but the user should select from a pool of already existing files. For this we have implemented a VFS. This also makes the dropzone obsolete.

jmuscireum:
> Dummy question: why do we need a specific template for BlobSoftRef? Perhaps it would haben been easier to understand if this was split in 2 commits: ones copies the hardref into softref, the next applies the deltas.

The UI is completely different, blobImageHardRefField directly opens a native file selection dialog, while blobImageSoftRefField opens a VFS selection dialog and allows specifying a URL.

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
bkositza: This is also my current state of knowledge. Therefore I have tried as far as possible not to use jQuery. Unfortunately we need the popover functionality of Bootstrap here. Is there a vanilla-js way to call a bootstrap popover?

idlira:
> The UI is completely different, blobImageHardRefField directly opens a native file selection dialog, while blobImageSoftRefField opens a VFS selection dialog and allows specifying a URL.

Maybe than post some screenies of the UI?

I just failing to correlate the need to select a file from a URL vs soft or hard refs.
A BlobSoftRef is like a BlobHardRef, with the difference that the latter lives as long as the underlying entity lives (the blob will be deleted by the framework if the entity is deleted) as the soft ref can stay in storage.

Why can't I upload an URL in both field types?

bkositza:
> Maybe than post some screenies of the UI?
I added some screenshots

> Why can't I upload an URL in both field types?
As you correctly pointed out, a hardRef is a binding link between the entity and the blob. If I have an entity that is a hardRef (but not a softRef), I can assume that the blob exists. This is not the case with a softRef. Here, multiple entities can also reference a blob, which does not even necessarily have to exist.

So a URL can't work for a hardRef, because we don't know if the file behind the URL still exists or not.

idlira (approved):

mko-sci (approved):

src/main/resources/default/taglib/t/blobImageSoftRefField.html.pasta
```diff
@@ -0,0 +1,128 @@
...
+            <small class="error-block form-text">@UserContext.get().getFieldErrorMessage(name)</small>
+        </div>
+    </i:if>
+</div>
+
+<script type="text/javascript">
+    {
+        const outerDiv = document.querySelector('#___id');
+        sirius.ready(function () {
+            const urlButton = outerDiv.querySelector('[data-popover]');
```
ymo-sci, line 71: Code-Style: Don't we want to name all elements with a _ as a prefix?

jmuscireum (approved):

oolscireum (approved):

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
```diff
@@ -0,0 +1,158 @@
...
+    function buildErrorCol(body) {
+        body._errorCol = document.createElement('div');
+        body._errorCol.classList.add('col', 'col-12');
+    }
+
+    function buildBodyContainer(body) {
+        body._container = document.createElement('div');
+        body._container.classList.add('row');
+        body._container.append(body._inputCol, body._errorCol, body._leftCol, body._rightCol);
+    }
```
qw3ry, line 101: We should have access to mustache here, right? Should be a lot easier and more concise via mustache.

bkositza: At first, I also tried Mustache. However, the amount of event listeners and the variables that have to be set make it confusing again with Mustache and lead to a mixture of both worlds, which was somehow unattractive.

ymo-sci (approved):

sabieber (approved):

qw3ry (approved):

src/main/resources/default/assets/scripts/blob-upload-url-popover.js
andyHa: I count 2

PR #1654, opened 2023-03-20
tycho enhancements

ymo-sci:
Bildschirm­foto 2023-03-20 um 15 51 16

### DIFF
--- a/src/main/resources/default/templates/biz/storage/list.html.pasta
+++ b/src/main/resources/default/templates/biz/storage/list.html.pasta
@@ -46,22 +46,22 @@
                     <t:filterbox page="children" baseUrl="@apply('/fs?path=%s', dir.path())"/>
                 </i:block>
 
-                <div class="row">
-                    <t:datacard>
-                        <table class="table table-hover">
+                <div class="card mb-4">
+                    <div class="card-body">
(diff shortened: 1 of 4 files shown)


src/main/resources/default/templates/biz/storage/list.html.pasta
```diff
@@ -46,22 +46,22 @@
...
+                <div class="card mb-4">
+                    <div class="card-body">
+                        <table class="table table-hover" style="table-layout: fixed">
                             <thead>
                             <tr>
-                                <th>
+                                <th scope="col">
                                     @i18n("Model.name")<br>
-                                    <small class="text-muted">@i18n("Model.description")</small>
+                                    <small class="text-muted text-nowrap">@i18n("Model.description")</small>
```
fhaScireum, line 56: Does `text-nowrap` work or do we actually need `text-truncate` here?

fhaScireum (approved):

src/main/resources/default/templates/biz/storage/list.html.pasta
ymo-sci: text-nowrap is ok here, only to now break "Description"/"Beschreibung" for the last letter on really small screens

PR #1661, opened 2023-03-27
Adds classes for better look

ymo-sci:
Bildschirm­foto 2023-03-27 um 11 08 16

also see https://github.com/scireum/sirius-web/pull/1198
- fixes: SIRi-769

### DIFF
--- a/src/main/resources/default/templates/biz/jobs/scheduler/entry.html.pasta
+++ b/src/main/resources/default/templates/biz/jobs/scheduler/entry.html.pasta
@@ -14,10 +14,11 @@
     </i:block>
 
     <i:block name="page-header">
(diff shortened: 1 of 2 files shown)


jakobvogel (approved):

sabieber (approved):
**Feedback:** I don't really like the practice of putting multiple unrelated changes into one commit. Generally speaking when a commit description needs an "and", bullet point list or comma to list all changes these should be multiple separate commits.
Also, the way the commit message is phrased it is actually not explained why the if is introduced or rather what the purpose of the if is ... that an if is added is clearly visible in the diff already.

cschierle (approved):

fhaScireum (approved):

PR #1669, opened 2023-03-30
Fixes and improves blob cache invalidation

mkeckmkeck:
- in case of blob changes, we need to cleanup all relevant caches
- unfortunately, the blobByPathCache works with paths that are not available at some layer 2 methods, so we cannot determine the cache keys
- so we use removal by filename as a good compromise to remove all required, but not to much. null value entries get removed as well

Fixes: SIRI-776

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -269,13 +269,11 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
      * Caches the {@link Blob blobs} that belong to certain paths
      */
     protected static Cache<String, Blob> blobByPathCache = CacheManager.<Blob>createCoherentCache("storage-paths")
-                                                                       .addRemover(REMOVE_BY_FILENAME,
-                                                                                   (input, entry) -> entry.getValue()
-                                                                                                     == null
-                                                                                                     || Strings.areEqual(
-                                                                                           entry.getValue()
-                                                                                                .getFilename(),
-                                                                                           input));
+                                                                       .addValueBasedRemover(REMOVE_BY_FILENAME)
+                                                                       .removeIf((input, blob) -> blob == null
+                                                                                                  || Strings.areEqual(
+                                                                               blob.getFilename(),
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -263,10 +263,19 @@
...
             CacheManager.createCoherentCache("storage-physical-keys");
 
+    protected static final String REMOVE_BY_FILENAME = "filename_remover";
+
     /**
      * Caches the {@link Blob blobs} that belong to certain paths
      */
-    protected static Cache<String, Blob> blobByPathCache = CacheManager.createCoherentCache("storage-paths");
+    protected static Cache<String, Blob> blobByPathCache = CacheManager.<Blob>createCoherentCache("storage-paths")
+                                                                       .addRemover(REMOVE_BY_FILENAME,
```
sabieber, line 272: **Optional:** Could in theory use `addValueBasedRemover` which directly passes the `entry.getValue()`

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -975,7 +984,7 @@ public void rename(B blob, String newName) {
         updateBlobName(blob, newName);
 
         blobKeyToFilenameCache.remove(blob.getBlobKey());
-        blobByPathCache.clear();
+        blobByPathCache.removeAll(REMOVE_BY_FILENAME, blob.getFilename());
```
sabieber, line 987: **Question:** Shouldn't this be done before AND after the name changes so also the old entry is removed?

sabieber (changes requested):

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
mkeckmkeck: Of course that is required. Thanks for noticing that!

sabieber (approved):

idlira (approved):

fhaScireum (approved):

ymo-sci (approved):

PR #1679, opened 2023-04-14
Adds a parameter which defines for up to how many days a generated url is valid per space

mko-sci:
Non-breaking as the default stays the same (2 days), but this allows to configure a longer time-frame for generated blob URLs via the config of the storage space.

Fixes: [OX-9707](https://scireum.myjetbrains.com/youtrack/issue/OX-9707/Gultigkeit-von-OXOMI-Downloadlinks-ist-zu-kurz)

### DIFF
--- a/src/main/java/sirius/biz/storage/util/StorageUtils.java
+++ b/src/main/java/sirius/biz/storage/util/StorageUtils.java
@@ -43,6 +43,11 @@ import java.util.regex.Pattern;
 public class StorageUtils {
 
     /**
+     * Defines the default url validity time in days.
+     */
+    public static final int DEFAULT_URL_VALIDITY_DAYS = 2;
+
+    /**
      * Names the framework which must be enabled to activate the storage feature.
      */
     public static final String FRAMEWORK_STORAGE = "biz.storage";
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/biz/storage/util/StorageUtils.java
```diff
@@ -90,9 +90,10 @@ public Collection<Extension> getStorageSpaces(ConfigScope scope) {
      *
      * @param key  the key to verify
      * @param hash the hash to verify
+     * @param validityDays the number of days the hash should be valid into the past
      * @return <tt>true</tt> if the hash verifies the given object key, <tt>false</tt> otherwise
      */
-    public boolean verifyHash(String key, String hash) {
+    public boolean verifyHash(String key, String hash, int validityDays) {
```
idlira, line 101: By changing the signature of a public method, how is that not breaking?

idlira (changes requested):

src/main/java/sirius/biz/storage/util/StorageUtils.java
mko-sci: Simple, we overloaded the method, the old method is still there

idlira (approved):

mkeckmkeck (approved):

PR #1687, opened 2023-04-20
Provides a simple way for inserting help/user assistance into backed …

andyHa:
…pages.

Using either some settings or (most probably) the system config, a product can now inject academy videos or knowledge base articles in their own and also in sirius-provided pages.

Furthermore, a storage space can specify these infos (which are shown in /fs) as well.

Fixes: OX-9803


mko-sci (approved):

sabieber (approved):

src/main/resources/default/extensions/tycho-page-header/user-assistent.html.pasta
```diff
@@ -0,0 +1,19 @@
+<i:arg type="String" name="point"/>
```
jmuscireum, line 1: assist**a**nt

mkeckmkeck (approved):

src/main/resources/default/extensions/tycho-page-header/user-assistent.html.pasta
mko-sci: why? - can you please elaborate

PR #1691, opened 2023-04-21
Lokalise: Translations update

scireumskip:
This pull request was initiated by Lokalise (user Matthias Brunner) at 2023-04-21 08:44:09


idlira (approved):
Spanish translation is not quite right, but closing an eye as we don't do grammatic

sabieber (approved):

PR #1707, opened 2023-05-05
Small Improvements of Administrative Tools

sabieber:
### Query Tool: Tenant Select Link on References

image

### Mail Protocol: Rendered HTML Preview

image

### DIFF
--- a/src/main/resources/default/templates/biz/model/query-results.html.pasta
+++ b/src/main/resources/default/templates/biz/model/query-results.html.pasta
@@ -45,6 +45,12 @@
                                                target="_blank">
                                                 <i class="fa fa-external-link-alt"></i> Query
                                             </a>
+                                            <i:if test="sirius.biz.tenants.Tenant.class.isAssignableFrom(reference.getReferencedType())">
+                                                <a href="/tenants/select/@reference.getValue(entity)"
+                                                   target="_blank">
+                                                    <i class="fa fa-external-link-alt"></i> Select
+                                                </a>
+                                            </i:if>
                                             </span>
                                         </i:if>
                                         <i:else>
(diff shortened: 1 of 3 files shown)


cschierle (approved):

fhaScireum (approved):

jakobvogel (approved):

src/main/resources/default/templates/biz/model/query-results.html.pasta
```diff
@@ -45,6 +53,12 @@
                                                target="_blank">
                                                 <i class="fa fa-external-link-alt"></i> Query
                                             </a>
+                                            <i:if test="sirius.biz.tenants.Tenant.class.isAssignableFrom(reference.getReferencedType())">
+                                                <a href="/tenants/select/@reference.getValue(entity)"
+                                                   target="_blank">
+                                                    <i class="fa fa-external-link-alt"></i> Select
+                                                </a>
+                                            </i:if>
```
andyHa, line 61: couldn't we use `<t:smartValue type="tenant" payload="@process.getTenantId()" label="@process.getTenantName()"/>` like in the processes UI - the same might be eligible for user-accounts

sabieber: Good idea, can try that. Ref-Fields containing a User Account are rather rare but adding the mechanism for them too is a no-brainer.

Actually I thought about if it would be useful/nice to add them to the TraceData as references that are not cascaded with the currently stored mail address as a fallback when the user is no longer present. This would make tracing a bit easier while not having a real overhead.

ymo-sci (approved):

cschierle (approved):

fhaScireum (approved):

PR #1708, opened 2023-05-05
EventRecorder: Allows counting distinct event field values

cschierle:
Provides possibility to count distinct values of a given field in queried events.

This is useful to e.g. determine the number of distinct entities that triggered certain events.

Fixes: [OX-9910](https://scireum.myjetbrains.com/youtrack/issue/OX-9910)

### DIFF
--- a/src/main/java/sirius/biz/analytics/events/EventRecorder.java
+++ b/src/main/java/sirius/biz/analytics/events/EventRecorder.java
@@ -15,10 +15,12 @@ import sirius.db.jdbc.SmartQuery;
 import sirius.db.jdbc.batch.BatchContext;
 import sirius.db.jdbc.batch.InsertQuery;
 import sirius.db.jdbc.schema.Schema;
+import sirius.db.mixing.Mapping;
 import sirius.db.mixing.annotations.Realm;
 import sirius.kernel.Sirius;
 import sirius.kernel.Startable;
 import sirius.kernel.Stoppable;
+import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Register;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/analytics/events/EventRecorder.java
```diff
@@ -221,6 +223,38 @@ public <E extends Event> int countEventsInRange(Class<E> eventType,
...
+     * the API of {@link sirius.kernel.health.metrics.Metrics}.
+     * @throws SQLException in case of a database error
+     */
+    public <E extends Event> int countDistinctValuesInEvents(Class<E> eventType,
+                                                             Mapping mapping,
+                                                             @Nullable Consumer<SmartQuery<E>> queryTuner)
+            throws SQLException {
+        SmartQuery<E> query = oma.select(eventType)
+                                 .aggregationField(Strings.apply("COUNT(DISTINCT %s) as distinctCount",
+                                                                 mapping.getName()));
```
fhaScireum, line 246: Can't we use prepared statements instead of insecure `Strings#apply`?

sabieber (approved):

jakobvogel (approved):
`distinctCount` in Konstante, so zur Sicherheit?

src/main/java/sirius/biz/analytics/events/EventRecorder.java
idlira: In local tests we got some errors trying to bind arguments in different portions of an SQL statement. I works 100% with arguments, but not always everywhere. Surely a shot could be done here, but since the source is a Mapping object, it's practically impossible to inject a bad statement here.

idlira (approved):

fhaScireum (approved):

PR #1716, opened 2023-05-11
Lazy Initialisation of String Select Parameter

jakobvogel:
> [OX-9733](https://scireum.myjetbrains.com/youtrack/issue/OX-9733)

This is intended for cases where options need to be loaded via `@Part` helper instances. These references are typically not availailable at construction time of the parameter, as the job factories (that use the parameters) are created "too early" during dependency injection.


mko-sci (approved):

src/main/java/sirius/biz/jobs/params/SelectStringParameter.java
```diff
@@ -37,23 +42,61 @@ public SelectStringParameter(String name, String label) {
...
+     *
+     * @param entriesProvider the provider that returns list of entries
+     * @return the parameter itself for fluent method calls
+     */
+    public SelectStringParameter withEntriesProvider(Supplier<Map<String, String>> entriesProvider) {
+        if (!entries.isEmpty()) {
+            throw Exceptions.createHandled()
+                            .withSystemErrorMessage(
+                                    "An entry provider can not be set after entries have already been added.")
+                            .handle();
```
andyHa, line 66: s.a.

sabieber (approved):

fhaScireum (approved):

PR #1720, opened 2023-05-15
Storage Framework meets read-only files

idlira:
Introduces the concept of read-only files. Key aspects:

- VirtualFiles can return if they are read-only `VirtualFile.readOnly()` and permits to modify the flag with `VirtualFile.updateReadOnlyFlag(boolean)`
- `VirtualFile.forceDelete()` has been added, which attempts to delete files marked as read-only
- Blobs have a new flag at DB-level
- Upstreams not based on Blobs implements the new flag based on existing API methods
- A tag identifies read-only files displayed in the `/fs` route. For those, only with a subset of permitted actions are displayed. System-admins have the permission to forcedly unlock a file
- Modal file selections filters out read-only files
- Jobs have a mechanism to lock files during submission. It's the job's task to unlock the file when appropriate. This is implemented in all `FileImportJobFactory` / `FileImportJob` factories/jobs.

**BREAKING**: all FileImportJobFactory + FileImportJob derivates will lock/unlock its input file. If this action is not desired, override the `collectFilesToLock` to perform nothing.

image

Fixes: [SIRI-805](https://scireum.myjetbrains.com/youtrack/issue/SIRI-805)


src/main/java/sirius/biz/storage/layer2/mongo/MongoBlob.java
```diff
@@ -524,4 +530,14 @@ public LocalDateTime getLastTouched() {
...
     }
+
+    @Override
+    public boolean isReadOnly() {
+        return readOnly;
+    }
+
+    @Override
+    public void setReadOnly(boolean readOnly) {
+        getStorageSpace().setReadOnly(this, readOnly);
```
sabieber, line 541: This sounds like a simple setter but it actually triggers an update query via the storage space. Is this a good combination?
Other setters of this class are just that - simple setters.

src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlobStorageSpace.java
```diff
@@ -475,6 +475,24 @@ protected void updateBlobName(SQLBlob blob, String newName) {
...
     }
 
+    @Override
+    protected void updateBlobReadOnlyFlag(SQLBlob blob, boolean readOnly) {
+        try {
+            oma.updateStatement(SQLBlob.class)
+               .set(SQLBlob.READ_ONLY, readOnly)
+               .where(SQLBlob.ID, blob.getId())
+               .executeUpdate();
+        } catch (SQLException e) {
```
sabieber, line 485: **Optional:** Rename `e` to `exception`

src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlobStorageSpace.java
```diff
@@ -475,6 +475,24 @@ protected void updateBlobName(SQLBlob blob, String newName) {
...
+            Exceptions.handle()
+                      .to(StorageUtils.LOG)
+                      .error(e)
+                      .withSystemErrorMessage(
+                              "Layer 2/SQL: An error occurred, when marking the read-only flag of blob '%s': %s (%s)",
```
sabieber, line 490: **Optional:** Include wether it was set to true or false?

src/main/java/sirius/biz/storage/layer2/mongo/MongoBlobStorageSpace.java
```diff
@@ -404,6 +404,14 @@ protected void updateBlobName(MongoBlob blob, String newName) {
         mango.update(blob);
     }
 
+    @Override
+    protected void updateBlobReadOnlyFlag(MongoBlob blob, boolean readOnly) {
+        mongo.update()
+             .set(MongoBlob.READ_ONLY, readOnly)
+             .where(MongoBlob.ID, blob.getId())
+             .executeForOne(MongoBlob.class);
```
sabieber, line 412: **Optional:** Check if the update result has a count of one and if not log an error like in SQL?

src/main/java/sirius/biz/storage/layer3/VirtualFile.java
```diff
@@ -460,6 +480,23 @@ public boolean exists() {
...
     }
 
+    /**
+     * Determines if the file is read-only.
+     *
+     * @return <tt>true</tt> if the file is read-only, <tt>false</tt> otherwise
+     */
+    public boolean readOnly() {
+        try {
+            if (readOnlyFlagSupplier != null) {
```
sabieber, line 490: **Optional:** Two new methods in the file, one uses the if as a guard-case, one the other way around. maybe switch one of them to be more consistent?
Other methods (as far as I can see in the PR view) seem to follow the way of the method I commented on.

sabieber (approved):

src/main/java/sirius/biz/storage/layer2/mongo/MongoBlob.java
idlira: A few other actions are processed like this: update, rename, touch...
Either another method name (which sounds less like a setter), or 2 methods (lock / unlock)?

src/main/java/sirius/biz/storage/layer2/mongo/MongoBlobStorageSpace.java
idlira: In SQL we don't count updated rows, but have to catch for SQLExceptions.
Other Mongo Operations in this class also run in fire-and-forget mode (eg. markBlobAsDeleted). If we change this, than should revise all other calls.

sabieber (approved):

mko-sci (approved):

src/main/java/sirius/biz/storage/layer3/VirtualFileSystemController.java
```diff
@@ -39,6 +39,8 @@
 import java.util.List;
 import java.util.Optional;
 
+import static sirius.biz.tenants.TenantUserManager.PERMISSION_SYSTEM_ADMINISTRATOR;
```
mkeckmkeck, line 42: Don't use static import (but for tests)

mkeckmkeck (approved):

src/main/resources/default/templates/biz/storage/list.html.pasta
```diff
@@ -130,14 +130,16 @@
...
-                                                   items="quickActions">
-                                                <t:dropdownItem url="@quickAction.getUrl()"
-                                                                icon="@quickAction.getIcon()"
-                                                                label="@quickAction.getLabel()"/>
-                                            </i:for>
+                                            <i:if test="!child.readOnly()">
+                                                <i:local name="quickActions"
+                                                         value="@controller.resolveQuickActionsForFile(child)"/>
+                                                <i:for type="sirius.biz.tycho.QuickAction" var="quickAction"
+                                                       items="quickActions">
```
cschierle, line 137:
```suggestion
                                                <i:for type="sirius.biz.tycho.QuickAction"
                                                       var="quickAction"
                                                       items="quickActions">

File: src/main/java/sirius/biz/storage/layer2/jdbc/SQLBlobStorageSpace.java [hunk 8]
```diff
@@ -487,8 +487,9 @@ protected void updateBlobReadOnlyFlag(SQLBlob blob, boolean readOnly) {
                       .to(StorageUtils.LOG)
                       .error(exception)
                       .withSystemErrorMessage(
-                              "Layer 2/SQL: An error occurred, when marking the read-only flag of blob '%s': %s (%s)",
-                              blob.getBlobKey())
+                              "Layer 2/SQL: An error occurred, when marking the read-only flag of blob '%s' to %s: %s (%s)",
```
cschierle, line 490:
```suggestion
                              "Layer 2/SQL: An error occurred, when updating the read-only flag of blob '%s' to %s: %s (%s)",
```
or
```suggestion
                              "Layer 2/SQL: An error occurred, when setting the read-only flag of blob '%s' to %s: %s (%s)",
```

cschierle (approved):

andyHa (approved):

PR #1729, opened 2023-05-24
Changes the accepted file extension for XML-Exports to "zip"

oolscireum:
Fixes: SE-12786

### DIFF
--- a/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
+++ b/src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
@@ -34,7 +34,7 @@ public abstract class XMLExportJobFactory extends FileExportJobFactory {
 
     @Override
     protected void collectAcceptedFileExtensions(Consumer<String> fileExtensionConsumer) {
-        fileExtensionConsumer.accept("xml");
+        fileExtensionConsumer.accept("zip");
     }
 
     @Override


fhaScireum (approved):

mkeckmkeck (approved):

src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
```diff
@@ -34,7 +34,7 @@ protected void collectParameters(Consumer<Parameter<?>> parameterCollector) {
 
     @Override
     protected void collectAcceptedFileExtensions(Consumer<String> fileExtensionConsumer) {
-        fileExtensionConsumer.accept("xml");
+        fileExtensionConsumer.accept("zip");
```
andyHa, line 37: do we always enforce compression or should we accept both?

sabieber (approved):

mko-sci (approved):

src/main/java/sirius/biz/jobs/batch/file/XMLExportJobFactory.java
oolscireum: currently we do always enforce compression.
even when there is ".xml" in the targetpath.

sabieber: Currently we seem to do so, would be cool to support both

idlira: I like forcing it, and so avoiding GB-big files "poluting" our storage.

idlira (approved):

andyHa (approved):

PR #1744, opened 2023-06-13
Permits to sum events over the provided mapping

idlira:
Note that a Syntax error is thrown trying to use a bind-variable as in `SELECT SUM(${foo}) AS ...`.

But since the value added here comes from a Mapping's name, only a _programmer_ would be able to inject bad code here.

Fixes: [OX-10023](https://scireum.myjetbrains.com/youtrack/issue/OX-10023)

### DIFF
--- a/src/main/java/sirius/biz/analytics/events/EventRecorder.java
+++ b/src/main/java/sirius/biz/analytics/events/EventRecorder.java
@@ -80,7 +80,7 @@ public class EventRecorder implements Startable, Stoppable, MetricProvider {
     private static final int MAX_EVENTS_PER_PROCESS = 16 * 1024;
 
     private static final String AGGREGATION_COUNTER = "counter";
-    private static final String AGGREGATION_SUM = "sum";
+    private static final String AGGREGATION_SUM = "summation";
     private static final String AGGREGATION_DISTINCT_COUNT = "distinctCount";
 
     private LocalDateTime lastProcessed;
--- a/src/main/java/sirius/biz/analytics/events/EventRecorder.java
+++ b/src/main/java/sirius/biz/analytics/events/EventRecorder.java
@@ -80,6 +80,7 @@ public class EventRecorder implements Startable, Stoppable, MetricProvider {
     private static final int MAX_EVENTS_PER_PROCESS = 16 * 1024;
 
     private static final String AGGREGATION_COUNTER = "counter";
+    private static final String AGGREGATION_SUM = "sum";
     private static final String AGGREGATION_DISTINCT_COUNT = "distinctCount";
 
     private LocalDateTime lastProcessed;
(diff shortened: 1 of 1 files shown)


mko-sci (approved):

src/main/java/sirius/biz/analytics/events/EventRecorder.java
```diff
@@ -226,6 +227,69 @@ public <E extends Event> int countEventsInRange(Class<E> eventType,
...
+     * @throws SQLException in case of a database error
+     * @see #countEventsInRange(Class, LocalDateTime, LocalDateTime, Consumer)
+     */
+    public <E extends Event> int sumEvents(Class<E> eventType,
+                                           Mapping mappingToSum,
+                                           @Nullable Consumer<SmartQuery<E>> queryTuner) throws SQLException {
+        SmartQuery<E> query = oma.select(eventType)
+                                 .aggregationField(Strings.apply("sum(%s) AS %s",
+                                                                 mappingToSum.getName(),
+                                                                 AGGREGATION_SUM));
```
andyHa, line 251: isn't sum a reserved word?

idlira: Well, works as an alias - just like `count` does 😇

src/main/java/sirius/biz/analytics/events/EventRecorder.java
```diff
@@ -226,6 +227,69 @@ public <E extends Event> int countEventsInRange(Class<E> eventType,
         });
     }
 
+    /**
+     * Sums the number of events for the provided column which have occurred based on the given <tt>queryTuner</tt>.
```
cschierle, line 231: It does not sum the number of events, but the values in the given field of those events, right?

```suggestion
     * Sums the values for the provided column of events which have occurred based on the given <tt>queryTuner</tt>.
```

src/main/java/sirius/biz/analytics/events/EventRecorder.java
```diff
@@ -226,6 +227,69 @@ public <E extends Event> int countEventsInRange(Class<E> eventType,
...
+
+        return query.asSQLQuery()
+                    .markAsLongRunning()
+                    .first()
+                    .flatMap(row -> row.getValue(AGGREGATION_SUM).asOptionalInt())
+                    .orElse(0);
+    }
+
+    /**
+     * Sums the number of events for the provided column which have occurred based on the given <tt>queryTuner</tt> and time range.
```
cschierle, line 264:
```suggestion
     * Sums the values for the provided column of events which have occurred based on the given <tt>queryTuner</tt>.
```

cschierle (approved):

mkeckmkeck (approved):

PR #1764, opened 2023-06-30
Tycho: Displays save hint after changes to a blobHardRefField

cschierle:
Fixes: [OX-10131](https://scireum.myjetbrains.com/youtrack/issue/OX-10131)

Bildschirmfoto 2023-06-30 um 12 23 12

Also fixes a wrong URL in filters for user-accounts

### DIFF
--- a/src/main/resources/biz_de.properties
+++ b/src/main/resources/biz_de.properties
@@ -904,6 +904,7 @@ StoredObject.errInvalidExtension = Ungültiger Dateityp
 StoredObject.errTooLarge = Datei zu groß
 StoredObject.errTooSmall = Datei zu klein
 StoredObject.resetBtn = Zurücksetzen
+StoredObject.saveHint = Die Änderungen werden erst nach dem Speichern wirksam.
 StoredObject.specifyURL = URL angeben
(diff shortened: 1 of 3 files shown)


jakobvogel (approved):
Is there a general rule to declare argument names before types? (If yes, I could watch out for similar cases in my own PRs)

mkeckmkeck (approved):

idlira (approved):

andyHa:
> Is there a general rule to declare argument names before types? (If yes, I could watch out for similar cases in my own PRs)

Actually it’s the other way round - we commonly try to keep it “Java style” type -> name - but that’s not enforced anywhere

sabieber:
> Actually it’s the other way round - we commonly try to keep it “Java style” type -> name - but that’s not enforced anywhere

Funnily I prefer it the other way round and follow that in all the templates I create.
The reasoning is simple: Most of the time the name of an argument is a more important information than its type and declaration should follow importance of information. Thats the same reason why languages like Kotlin switch them around. Or why I like to keep boilerplate (getters and setters) at the very bottom of the class below the actual logic.

sabieber (approved):

mko-sci (approved):

ymo-sci (approved):

PR #1772, opened 2023-07-06
Provides a framework to check customization flags.

andyHa:
This has been ported from OXOMI were we used it to accommodate several migrations / modernization projects.

grafik

grafik

grafik

### DIFF
--- a/src/main/java/sirius/biz/tenants/flags/CustomizationFlags.java
+++ b/src/main/java/sirius/biz/tenants/flags/CustomizationFlags.java
@@ -8,6 +8,7 @@
 
 package sirius.biz.tenants.flags;
 
+import sirius.biz.tenants.Tenant;
 import sirius.biz.tenants.Tenants;
 import sirius.biz.web.TenantAware;
 import sirius.kernel.async.CallContext;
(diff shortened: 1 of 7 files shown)


src/main/java/sirius/biz/tenants/flags/CustomizationFlags.java
```diff
@@ -0,0 +1,137 @@
...
+
+        return readFromUserAgent(flagName).or(() -> readFromUserSettings(flagName))
+                                          .or(() -> readFromScopeSettings(flagName))
+                                          .or(() -> readFromScopeTenantSettings(flagName))
+                                          .or(() -> readFromSystemTenantSettings(flagName))
+                                          .orElseGet(() -> Tuple.create(defaultValue, "default"));
+    }
+
+    private Optional<Tuple<Boolean, String>> readFromUserAgent(String flagName) {
+        WebContext webContext = CallContext.getCurrent().get(WebContext.class);
```
sabieber, line 86: **Question:** Can this be null or fail when checking a flag in a background process or job?

sabieber (commented):
The only thing that is not possible with this solution that was possible with the flag logic in OX is, that we can't set the flag for the system tenant independently of the global state. Right?

src/main/java/sirius/biz/tenants/flags/CustomizationFlags.java
andyHa: Nope - this would yiel and empty WebContext which is not valid (see below)

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

fhaScireum (approved):

PR #1798, opened 2023-07-25
Migrates some tests to jUnit5 + Kotlin

sabieber:


mko-sci (approved):

src/test/kotlin/sirius/biz/analytics/events/EventRecorderTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        recorder.record(TestEvent1())
+        recorder.record(TestEvent1())
+        recorder.record(TestEvent1())
+
+        recorder.record(TestEvent2())
+        recorder.record(TestEvent2())
+        recorder.record(TestEvent2())
+        recorder.record(TestEvent2())
+
+        recorder.javaClass.getDeclaredField("bufferedEvents").apply {
```
jakobvogel, line 78: Komische Frage vom Kotlin-Noob – Warum ist das im Vergleich zu Groovy komplizierter?

src/test/kotlin/sirius/biz/cluster/RedisUserMessageCacheTest.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import sirius.kernel.SiriusExtension
+import sirius.kernel.di.std.Part
+import sirius.web.controller.Message
+import sirius.web.controller.MessageLevel
+import sirius.web.http.DistributedUserMessageCache
+import java.util.*
```
jakobvogel, line 20: Wollen wir den Wildcard in Kotlin?

src/test/kotlin/sirius/biz/elastic/SearchableEntityTest.kt
```diff
@@ -0,0 +1,85 @@
...
+package sirius.biz.elastic
+
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
+import sirius.kernel.SiriusExtension
+import java.util.*
```
jakobvogel, line 18: Siehe oben

jakobvogel (approved):

src/test/kotlin/sirius/biz/analytics/events/EventRecorderTest.kt
sabieber: Because Kotlin, like Java, handles visibility properly and `bufferedEvents` is a private field. With Groovy you can access everything and nothing (even non-existent methods, ...), with Java and Kotlin only what an external class in the same package could access.

An alternative solution would be declare the field as `protected`.

src/test/kotlin/sirius/biz/cluster/RedisUserMessageCacheTest.kt
sabieber: I think we have to set up the Kotlin style/formatting rules correctly in sirius-parent.
There are official coding conventions that are available and should be applied.

As other imports are not converted to wildcard imports my guess is that only `java.util` is allows to be imported as wildcard.

sabieber: Would also suggest to enable the "wildcard imports" inspection in sirius parent :)

sabieber: Here is a link to the style guide: https://kotlinlang.org/docs/coding-conventions.html

sabieber: The main problem that arises with star imports in Java (name clashes) is actually not present in kotlin because there is this small feature:
`import org.test.Message as TestMessage`

src/test/kotlin/sirius/biz/analytics/events/EventRecorderTest.kt
jakobvogel: Ah💡 ... Thanks for the clarification!

jakobvogel (approved):

mkeckmkeck (approved):

idlira (approved):

PR #1804, opened 2023-07-31
Adds a method which counts the number of events which have occurred in the given calendar month based on the given queryTuner

mko-sci:
Fixes: OX-10078


fhaScireum (commented):
`#countEventsInLastMonth` could now call the new method.

jakobvogel (approved):

fhaScireum (approved):

PR #1810, opened 2023-08-02
Jupiter: Usability for Products without the Tenant Framework

jakobvogel:
Enables products that do not use the _Tenants_ framework to use Jupiter by providing the system tenant differently.

### DIFF
--- a/src/main/java/sirius/biz/analytics/explorer/EntityChartObjectResolver.java
+++ b/src/main/java/sirius/biz/analytics/explorer/EntityChartObjectResolver.java
@@ -14,6 +14,7 @@ import sirius.db.mixing.BaseEntity;
 import sirius.db.mixing.Mixing;
 import sirius.kernel.di.std.Part;
 
+import javax.annotation.Nullable;
 import java.util.Optional;
 
 /**
@@ -30,6 +31,7 @@ public abstract class EntityChartObjectResolver<E extends BaseEntity<?>> impleme
     protected Mixing mixing;
 
     @Part
+    @Nullable
     protected Tenants<?, ?, ?> tenants;
 
     @Override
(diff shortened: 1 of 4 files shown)


sabieber (approved):
Couldn't MEMOIO register its own `Tenants` class that just provide the ID an name?

jakobvogel:
> Couldn't MEMOIO register its own `Tenants` class that just provide the ID an name?

The `Tenants` class is rather large, and implementing it may inadvertently enable other functionality. For now, I would like to keep a clear separation, and maintain the hope that one day, we will be able to migrate MEMOIO to use the tenants framework as well. 🤞

sabieber:
> The `Tenants` class is rather large, and implementing it may inadvertently enable other functionality. For now, I would like to keep a clear separation, and maintain the hope that one day, we will be able to migrate MEMOIO to use the tenants framework as well. 🤞

Yeah but implementing workarounds in various places throughout the framework until that day comes doesn't sound desirable as well 😶

jakobvogel:
> Yeah but implementing workarounds in various places throughout the framework until that day comes doesn't sound desirable as well 😶

Correct, neither are desirable solutions. But comparing effort, possible side effects, resources, etc., the proposed way of this PR still seems to be the more justifiable one to me — or we decide not to use _Jupiter_ for MEMOIO, which would lead to incomplete support of the knowledge base, and with it, of the API documentation.

sabieber:
No I certainly don't want to hinder usage of Jupiter/DataVault in MIO :)
Also thought about other workaround options like overwriting the Sync but all I came up with wasn't easily implementable.

mkeckmkeck (approved):

sabieber:
Sadly the `framework` setting in the `@Register` annotation only allows for a single framework to be mentioned, otherwise we could set jupiter + tenants there and MIO could inherit the class, overwrite the two methods and register it itself.

mko-sci (approved):

PR #1816, opened 2023-08-07
Optimizes user UI

juliawuerthele:
- saving user routes to user instead of overview
- user image centered
- long mail addresses are shortened and don't result in a line break anymore 
    -> made changes in web necessary (datacard.html.pasta), the layout of other datacards isn't affected by this as far as I 
         can see but if anyone has a better solution which doesn't involve web tell me:)
    -> see https://github.com/scireum/sirius-web/pull/1270

fixes: OX-10216

Bildschirmfoto 2023-08-07 um 13 45 06

### DIFF
--- a/src/main/resources/default/templates/biz/tenants/user-accounts-list.html.pasta
+++ b/src/main/resources/default/templates/biz/tenants/user-accounts-list.html.pasta
@@ -23,8 +23,10 @@
                     </div>
                 </i:block>
                 <i:block name="title">
+                    <div class="overflow-hidden text-ellipsis text-nowrap">
                     <t:langFlag lang="@account.getUserAccountData().getLanguage().getValue()"/>
                     <span>@account.getUserAccountData()</span>
+                    </div>
                 </i:block>
                 <i:block name="actions">
                     <i:if test="enableCardActions">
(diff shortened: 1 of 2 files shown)


mko-sci (approved):

sabieber (approved):

src/main/resources/default/templates/biz/tenants/user-accounts-list.html.pasta
```diff
@@ -12,7 +12,7 @@
                     headerLeftAlignmentClass="align-items-stretch text-break break-word"
                     link="@apply('%s/%s', cardBaseUrl , account.getIdAsString())">
                 <i:block name="header">
-                    <div class="pl-2 pr-1 align-self-center">
+                    <div class="pl-3 pr-1 align-self-center">
```
idlira, line 15: Why are `pl` and `pr` different?

jakobvogel (approved):

src/main/resources/default/templates/biz/tenants/user-accounts-list.html.pasta
sabieber: Because the card body (the right section here) itself has a padding of `1.25rem` that has to be factored in.

idlira (approved):

mkeckmkeck (approved):

fhaScireum (approved):

PR #1818, opened 2023-08-09
Validates mail addresses and phone numbers via new shared `PropertyValidators`

sabieber:
## **Behaviour Change**

* Phone numbers and mail address of `ContactData` is always validated, not only when calling the validation methods. This is coordinated with S2. 

---

Defines shared validators for mail addresses and phone numbers. Motivation for this is that the same logic can easily be shared/reused across multiple entities that contain such fields and behaviour of those fields is more consistent.

Fixes: [SIRI-854](https://scireum.myjetbrains.com/youtrack/issue/SIRI-854)


fhaScireum (approved):

src/main/java/sirius/biz/tenants/PhoneNumberValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                         .withNLSKey("ContactData.invalidPhone")
+                         .set("field", property.getLabel())
+                         .set("value", value)
+                         .handle();
+    }
+
+    @Nonnull
+    @Override
+    public String getName() {
+        return "email-address-validator";
```
jakobvogel, line 66:
```suggestion
        return "phone-number-validator";
```

sabieber: Thanks 👍🏻

jakobvogel (approved):

idlira (approved):

mkeckmkeck (approved):

PR #1824, opened 2023-08-22
Introduces the Blob.duplicate() method

idlira:
Used to clone a blob. This method delivers a BlobDuplicator class which collects parameters for the copy, such as variants to clone with or a new target storage space.

The new blob created is temporary and must be handled after creation (for example, attaching it to an actual BlobHardRef), otherwise it will be deleted from the system after a while.

Under the hood, it can handle copies from/to different stores.

In case of S3, a server-side copy can be performed if possible, falling back to regular download/upload when a straight copy cannot be performed.

Fixes: [SIRI-851](https://scireum.myjetbrains.com/youtrack/issue/SIRI-851)

### DIFF
--- a/src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
@@ -69,6 +69,9 @@ public abstract class ObjectStorageSpace {
     @Part
     private static Tasks tasks;
 
+    @Part
+    protected static ObjectStorage objectStorage;
+
     private static final Counter UPLOADS = new Counter();
     private static final Counter DOWNLOADS = new Counter();
     private static final Counter STREAMS = new Counter();
@@ -245,6 +248,28 @@ public abstract class ObjectStorageSpace {
      */
     protected abstract void storePhysicalObject(String objectKey, File file, ByteBlockTransformer transformer)
             throws IOException;
+
+    /**
+     * Copies the given object to the given target object key in the current storage space.
+     *
+     * @param sourceObjectKey    the source object key to copy from
+     * @param targetObjectKey    the target object key to copy to
+     * @param targetStorageSpace the target storage space to copy to
+     * @throws IOException in case of an IO error
+     */
+    public abstract void copyPhysicalObject(String sourceObjectKey, String targetObjectKey, String targetStorageSpace)
+            throws IOException;
+
+    /**
+     * Copies the given object to the given target object key in the current storage space.
+     *
+     * @param sourceObjectKey the source object key to copy from
+     * @param targetObjectKey the target object key to copy to
+     * @throws IOException in case of an IO error
+     */
+    public void copyPhysicalObject(String sourceObjectKey, String targetObjectKey) throws IOException {
+        copyPhysicalObject(sourceObjectKey, targetObjectKey, getName());
+    }
 
     /**
      * Stores the given data for the given object key.
--- a/src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
@@ -255,20 +255,23 @@ public abstract class ObjectStorageSpace {
      * @param sourceObjectKey    the source object key to copy from
      * @param targetObjectKey    the target object key to copy to
      * @param targetStorageSpace the target storage space to copy to
-     * @throws IOException in case of an IO error
      */
-    public abstract void copyPhysicalObject(String sourceObjectKey, String targetObjectKey, String targetStorageSpace)
-            throws IOException;
+    public void duplicatePhysicalObject(String sourceObjectKey, String targetObjectKey, String targetStorageSpace) {
+        download(sourceObjectKey).ifPresent(fileHandle -> {
+            try (fileHandle) {
+                objectStorage.getSpace(targetStorageSpace).upload(targetObjectKey, fileHandle.getFile());
+            }
+        });
+    }
 
     /**
      * Copies the given object to the given target object key in the current storage space.
      *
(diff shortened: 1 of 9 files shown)


src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
```diff
@@ -0,0 +1,111 @@
...
+     * @return a new duplicator for the given blob
+     */
+    public static BlobDuplicator create(Blob blob) {
+        return new BlobDuplicator(blob);
+    }
+
+    private BlobDuplicator(Blob blob) {
+        this.blob = blob;
+        this.storageSpace = blob.getSpaceName();
+    }
```
jakobvogel, line 47: Why not just a public constructor?

src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Adds a variant name to be duplicated among with the blob.
+     *
+     * @param variantName the variant name to duplicate
+     * @return the duplicator itself for fluent method calls
+     */
+    public BlobDuplicator withVariant(String variantName) {
+        this.variants.add(variantName);
```
jakobvogel, line 56: What if the variant does not exist?

jakobvogel (approved):

src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
idlira: It will silently not be copied.
I choose this way because one could define variants in DB but not necessarily be attached to a real "converter".
At the same time, a real variant name might not exist for the blob being copied.

src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
idlira: Just following the pattern used in several other classes.

fhaScireum (approved):

src/main/java/sirius/biz/storage/layer1/S3ObjectStorageSpace.java
```diff
@@ -237,4 +242,60 @@ public void iterateObjects(Predicate<ObjectMetadata> objectHandler) throws IOExc
...
+                        store.getClient().getObjectMetadata(bucketName().getName(), sourceObjectKey).getContentLength();
+                replicationManager.notifyAboutUpdate(s3ObjectStorageSpace, targetObjectKey, size);
+            } else {
+                // ... but the source and target buckets resides in different systems. We have to download and upload.
+                downloadAndUploadFile(sourceObjectKey, targetObjectKey, targetSpace);
+            }
+        } else {
+            // We are copying from S3 to a different storage system (eg. FS). We have to download and upload.
+            downloadAndUploadFile(sourceObjectKey, targetObjectKey, targetSpace);
+        }
```
andyHa, line 270: can we combine those with an early return or the like?

src/main/java/sirius/biz/storage/layer1/ObjectStorageSpace.java
```diff
@@ -246,6 +249,28 @@ public void upload(String objectId, File file) {
...
 
+    /**
+     * Copies the given object to the given target object key in the current storage space.
+     *
+     * @param sourceObjectKey    the source object key to copy from
+     * @param targetObjectKey    the target object key to copy to
+     * @param targetStorageSpace the target storage space to copy to
+     * @throws IOException in case of an IO error
+     */
+    public abstract void copyPhysicalObject(String sourceObjectKey, String targetObjectKey, String targetStorageSpace)
```
andyHa, line 260: do we really need a targetStorageSpace - I don't think we can optimize copying accross multiple spaces - but each impl would have to take care of it.

or we handle the "not same case" in the base class. we could provide a "hand rolled copy" and only optimize the s3 case and otherwise always call super.copy... 

Also, maybe we call this "duplicate" to better show what it does(?)

src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
andyHa: I think this is mostly an anti-pattern left over from times where the diamond operator wasn't supported and generics had to be repeated for new calls

src/main/java/sirius/biz/storage/s3/ObjectStore.java
```diff
@@ -210,6 +210,15 @@ public AmazonS3 getClient() {
...
 
+    /**
+     * Returns the name of the store.
+     *
+     * @return the store's name
+     */
+    public String getName() {
+        return name;
+    }
+
```
andyHa, line 221: or we properly implement equals for ObjectStor and check this above(?)

andyHa (changes requested):

src/main/java/sirius/biz/storage/layer2/BlobDuplicator.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Returns the temporary blob as a replica of the original one.
+     *
+     * @return the cloned blob
+     */
+    @SuppressWarnings("unchecked")
+    public Blob execute() {
+        Blob newBlob = blobStorage.getSpace(storageSpace).createTemporaryBlob(blob.getTenantId());
+        BasicBlobStorageSpace space = (BasicBlobStorageSpace) newBlob.getStorageSpace();
```
andyHa, line 92: we should never downcast to an implementation (?)

idlira: k.A wie ich ohne schaffen kann.
Das Problem ist, dass der Blob interface keinen parameter hat...

andyHa (approved):

PR #1827, opened 2023-08-23
BREAKING: KBA - Remodels the fallback language mechanism for the knowledgebase system in order to support multiple fallback languages

mko-sci:
Fixes: SIRI-865

### DIFF
--- a/src/main/java/sirius/biz/tycho/kb/KnowledgeBase.java
+++ b/src/main/java/sirius/biz/tycho/kb/KnowledgeBase.java
@@ -113,34 +113,6 @@ public class KnowledgeBase {
     }
 
     /**
-     * Determines the fallback or default lang of the knowledge base.
-     *
-     * @return the default language as two-letter ISO code
-     * @deprecated call {@link #getFallbackLanguage()} instead.
-     */
-    @Deprecated
-    public final String getFallbackLang() {
-        return getFallbackLanguage();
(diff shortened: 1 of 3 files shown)


idlira (approved):
Probably someone would prefer .stream() versus regular loops :-P

andyHa (approved):

src/main/java/sirius/biz/tycho/kb/KnowledgeBase.java
```diff
@@ -125,7 +129,15 @@ public final String getFallbackLang() {
      * @return the default language as two-letter ISO code
      */
     public String getFallbackLanguage() {
-        return fallbackLanguage;
+        return !fallbackLanguages.isEmpty() ? null : fallbackLanguages.get(0);
+    }
```
andyHa, line 133: Check if we can kill this method and the getFallbackLang above

mko-sci: we can kill it since they are not used among scireum-core and ox =)

mko-sci:
> Probably someone would prefer .stream() versus regular loops :-P

But its way easier to maintain and understand since there is already a lot of stuff going on inside the loop

fhaScireum (approved):

idlira (approved):

PR #1834, opened 2023-09-08
SIRI-861: Minor Blob Upload Fix

fhaScireum:
Fixes updating the preview for URLs by utilizing a callback function (implemented in a non breaking way).

### DIFF
--- a/src/main/resources/default/assets/scripts/blob-upload-url-popover.js
+++ b/src/main/resources/default/assets/scripts/blob-upload-url-popover.js
@@ -9,7 +9,7 @@
  * @param _fileNameContainer The container that shows the filename
  * @param i18n Contains information about the label, the ok-Button and the cancel-Button of the popover
  */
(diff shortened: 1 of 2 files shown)


sabieber (approved):

idlira (approved):

mko-sci (approved):

jakobvogel (approved):
Hmm, if one relied on the the image source being set… wouldn't this PR be breaking? A "migration guide" may be nice in this case, showing how to put that all into the new callback function … or is the feature so young that nobody has been using it yet?

fhaScireum:
> Hmm, if one relied on the the image source being set… wouldn't this PR be breaking? A "migration guide" may be nice in this case, showing how to put that all into the new callback function … or is the feature so young that nobody has been using it yet?

The feature is very young and only used for the t:blobImageSoftRefField tag right now.

mkeckmkeck (approved):

bkositza (approved):

PR #1843, opened 2023-09-21
Fixes an issue where paperclip files have not been deleted from the S3 storage, although they were older than the retention date

mko-sci:
This was caused by a limit inside the 'mark as deleted' loop. The limit was set to 256 entities. Also, the loop was only executed once per day. This was problematic since we create over 1000 paperclips a day which resulted in a big yam, and it looked like we are not deleting files past their retention date.

Removing the limit should be safe since we only switch a flag/update each entity. The actual deletion is performed in another task which has a limit. The limit here is not a problem since the actual deletion task is executed "constantly" in a loop with other tasks.

Fixes: [SIRI-785](https://scireum.myjetbrains.com/youtrack/issue/SIRI-785)


jakobvogel (approved):

sabieber (approved):

mkeckmkeck (commented):
Shouldn't we change this at sirius.biz.storage.layer2.jdbc.SQLBlobStorageSpace as well?

mkeckmkeck (approved):

PR #1845, opened 2023-09-22
SE-13063: Fixes not being able to save configurations more than once.

fhaScireum:
This error only manifests itself in Chrome browser (possibly other chromium browsers), as the javascript is not executed due to `event.stopPropagation();`. Firefox for instance was never affected. This issue was introduced by https://github.com/scireum/sirius-web/pull/1286

Fixes: [SE-13063](https://scireum.myjetbrains.com/youtrack/issue/SE-13063)

### DIFF
--- a/src/main/resources/default/templates/biz/tenants/tenant-config.html.pasta
+++ b/src/main/resources/default/templates/biz/tenants/tenant-config.html.pasta
@@ -3,7 +3,7 @@
 
 <i:invoke template="/templates/biz/tenants/tenant.html.pasta" tenant="tenant" controller="controller" page="config">
     <t:codeEditor id="editor" mode="json">@tenant.getTenantData().getConfigString()</t:codeEditor>
-    <t:formBar backButton="false"/>
+    <t:formBar backButton="false" singleClick="false"/>
 
     <script type="text/javascript">
         sirius.ready(function () {
--- a/src/main/resources/default/templates/biz/tenants/user-account-config.html.pasta
+++ b/src/main/resources/default/templates/biz/tenants/user-account-config.html.pasta
@@ -2,7 +2,7 @@
 
 <i:invoke template="/templates/biz/tenants/user-account.html.pasta" account="account" page="config">
     <t:codeEditor id="editor" mode="json">@account.getUserAccountData().getPermissions().getConfigString()</t:codeEditor>
-    <t:formBar backButton="false"/>
+    <t:formBar backButton="false" singleClick="false"/>
 
     <script type="text/javascript">
         sirius.ready(function () {


mko-sci:
What is meant by more than once?
Once per page visit?
Once (total since application deployment)?

mko-sci (approved):

jmuscireum (approved):

mkeckmkeck (approved):

oolscireum (approved):

src/main/resources/default/templates/biz/tenants/tenant-config.html.pasta
```diff
@@ -3,7 +3,7 @@
 
 <i:invoke template="/templates/biz/tenants/tenant.html.pasta" tenant="tenant" controller="controller" page="config">
     <t:codeEditor id="editor" mode="json">@tenant.getTenantData().getConfigString()</t:codeEditor>
-    <t:formBar backButton="false"/>
+    <t:formBar backButton="false" singleClick="false"/>
```
andyHa, line 6: I think we should rather limit the single-click handling to pages where a form actually exists...:
https://github.com/scireum/sirius-web/blob/4836f26736699091181787d5e4591c25cf9e36b1/src/main/resources/default/assets/tycho/scripts/form.js.pasta#L2C36-L19C8

andyHa: We also should perform a null-check here:
https://github.com/scireum/sirius-web/blob/4836f26736699091181787d5e4591c25cf9e36b1/src/main/resources/default/assets/tycho/scripts/form.js.pasta#L17C1-L17C1

andyHa (changes requested):

src/main/resources/default/templates/biz/tenants/tenant-config.html.pasta
fhaScireum: > We also should perform a null-check here: https://github.com/scireum/sirius-web/blob/4836f26736699091181787d5e4591c25cf9e36b1/src/main/resources/default/assets/tycho/scripts/form.js.pasta#L17C1-L17C1

We already have a null check here: https://github.com/scireum/sirius-web/blob/4836f26736699091181787d5e4591c25cf9e36b1/src/main/resources/default/assets/common/core.js.pasta#L375



> I think we should rather limit the single-click handling to pages where a form actually exists...: https://github.com/scireum/sirius-web/blob/4836f26736699091181787d5e4591c25cf9e36b1/src/main/resources/default/assets/tycho/scripts/form.js.pasta#L2C36-L19C8

I feel like there could be instances where data i submitted "indirectly" (outside any form like it is the case here) where we want to only be able to submit the data once. I think it's cleaner when singleClick == false always means that the submit action is only triggered once.

andyHa: If there is no form, there is no page reload and having single click == true will always screw the behaviour. On the other hand, if there is a page-reload, why not use a simple form anyway?

TL;DR - no form -> no single-click enforcing seems always right...

fhaScireum:
Closed via alternative solution in https://github.com/scireum/sirius-web/pull/1290

PR #1896, opened 2023-11-30
Support MultiSelectParameter in job scheduler entries

jmuscireum:
### Description
This changes how list types are stored in JobConfigData. Instead of pipe separated strings, the entries are now stored as true JSON arrays.

This provides two benefits:
1. StringParameters with pipes in the value are supported again (these were wrongfully handled as list values)
2. Scheduler entries now also support MultiSelectParameters without having to duplicate a lot of logic

This is breaking because job presets for jobs using MultiSelectParameters are not compatible anymore (probably only exist in S2). These need to be saved again after selecting the correct values.

Empty parameters are now stored as null, while maintaining the correct logic for parameter defaults.

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-13229](https://scireum.myjetbrains.com/youtrack/issue/SE-13229)

### Checklist


jakobvogel (approved):

mkeckmkeck (approved):
Fänds bisschen eleganter da mit List<String> statt Value zu arbeiten.

mko-sci (approved):

idlira (approved):

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

PR #1921, opened 2024-01-15
Filter out duplicated defined custom-entries and only take the custom one

ymo-sci:
### Description
This is needed to "override"  base defined entries in a custom file and not duplicate it in the system (where the difference is not easily visible)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/codelists/CustomLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CustomLookupTable.java
@@ -107,7 +107,7 @@ class CustomLookupTable extends LookupTable {
                              baseTable.performSearch(searchTerm, Limit.UNLIMITED, language))
                      .skip(limit.getItemsToSkip())
                      .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems())
-                     .collect(Collectors.toMap(LookupTableEntry::getCode, Function.identity(), (a, b) -> a))
+                     .collect(filterCustomDuplicateCollector())
                      .values()
                      .stream();
     }
--- a/src/main/java/sirius/biz/codelists/CustomLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/CustomLookupTable.java
@@ -13,11 +13,9 @@ import sirius.kernel.commons.Value;
 import sirius.kernel.settings.Extension;
 
 import javax.annotation.Nonnull;
-import java.util.Map;
+import java.util.HashSet;
 import java.util.Optional;
-import java.util.function.Function;
-import java.util.stream.Collector;
-import java.util.stream.Collectors;
+import java.util.Set;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/codelists/CustomLookupTable.java
```diff
@@ -116,6 +130,13 @@ public int count() {
...
     protected Stream<LookupTableEntry> performQuery(String language, String lookupPath, String lookupValue) {
         return Stream.concat(customTable.performQuery(language, lookupPath, lookupValue),
-                             baseTable.performQuery(language, lookupPath, lookupValue));
+                             baseTable.performQuery(language, lookupPath, lookupValue))
+                     .collect(filterCustomDuplicateCollector())
+                     .values()
+                     .stream();
+    }
+
+    private Collector<LookupTableEntry, ?, Map<String, LookupTableEntry>> filterCustomDuplicateCollector() {
```
idlira, line 139: I would expect a method called filter*** to be used with a Stream.filter.

src/main/java/sirius/biz/codelists/CustomLookupTable.java
```diff
@@ -90,22 +95,31 @@ protected Stream<LookupTableEntry> performSuggest(Limit limit, String searchTerm
         return Stream.concat(customTable.performSuggest(Limit.UNLIMITED, searchTerm, language),
                              baseTable.performSuggest(Limit.UNLIMITED, searchTerm, language))
                      .skip(limit.getItemsToSkip())
-                     .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems());
+                     .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems())
+                     .collect(filterCustomDuplicateCollector())
+                     .values()
```
idlira, line 100: Why not `.filter` with a Predicate? Fells not performant to collect everything and build a new stream.

idlira (changes requested):

src/main/java/sirius/biz/codelists/CustomLookupTable.java
jmuscireum: filter can't access the other elements in the stream. Another solution would be something like:

```java
Map<String, LookupTableEntry> results = new LinkedHashMap<>();

customTable.performSuggest(Limit.UNLIMITED, searchTerm, language)
           .forEach(entry -> results.put(entry.getCode(), entry));

baseTable.performSuggest(Limit.UNLIMITED, searchTerm, language)
         .forEach(entry -> results.putIfAbsent(entry.getCode(), entry));

return results.values().stream().limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems());
```

Implementing `LookupTableEntry#equals` which only compares the code and then using distinct here might be another solution.

In any case the limit needs to be the last part in the chain.

idlira: I think you can try with a `.reduce()` As soon as you get 2 items with the same code, you gotta decide which one to keep.

idlira: or maybe something like that could work as well. So you only need the codes in a temp set.

```java
Set<String> codes = new HashSet<>();
        return Stream.concat(customTable.performSuggest(Limit.UNLIMITED, searchTerm, language),
                             baseTable.performSuggest(Limit.UNLIMITED, searchTerm, language))
                     .skip(limit.getItemsToSkip())
                     .filter(codes::add)
                     .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems());
```

idlira (approved):

sabieber (approved):

jmuscireum (approved):

mkeckmkeck (approved):

PR #1932, opened 2024-02-08
ensure unique labels in import-dictionaries

ymo-sci:
### Description

Sometimes we have enhanced dictionaries where labels with the same name will be added and always an exception due to a doubled alias will be thrown. With this approach we suffix the labels with the name to ensure uniqueness. Example: WEIGHT and NET_WEIGHT both are translated with "Gewicht" and added to the same import-dictionary will result in "Gewicht" and "Gewicht_NET_WEIGHT" now. (see Screenshots)

- fixes: SE-13234
Bildschirmfoto 2024-02-08 um 11 35 41

Bildschirmfoto 2024-02-08 um 11 35 08

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-](https://scireum.myjetbrains.com/youtrack/issue/SIRI-)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/importer/format/ImportDictionary.java
+++ b/src/main/java/sirius/biz/importer/format/ImportDictionary.java
@@ -138,7 +138,7 @@ public class ImportDictionary {
     /**
      * Returns a list of all fields in this record.
      * <p>
-     * Note that this is not neccessarily in the expected order. Use the <tt>mapping function</tt> to specify the order.
+     * Note that this is not necessarily in the expected order. Use the <tt>mapping function</tt> to specify the order.
      *
      * @return the fields in this dictionary
      */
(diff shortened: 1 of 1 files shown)


mkeckmkeck (approved):

jakobvogel (approved):

fhaScireum (commented):
What is the benefit of using the Label/Alias "Gewicht_property_WEIGHT_SUPER" if I could also simply use "WEIGHT_SUPER"? That defeats the purpose of having simple human readable translations that can be used.

mkeckmkeck:
> What is the benefit of using the Label/Alias "Gewicht_property_WEIGHT_SUPER" if I could also simply use "WEIGHT_SUPER"? That defeats the purpose of having simple human readable translations that can be used.

The problem is, that previously "WEIGHT" and "WEIGHT_SUPER" would both have their label "Gewicht" displayed in the job description. But when actually using "Gewicht" in an import, only one of them would have been used. 
The proper solution for such problems would be the elimination of duplicates. But if this is not possible, we should at least be consistent with the handling of the labels.

fhaScireum (approved):

idlira (approved):

PR #1943, opened 2024-02-23
Adds an GH action which controls the PR labels before merging

sabieber:
- The required labels are used for auto-categorizing GA release notes
- The forbidden labels are used to prevent accidentally merging unfinished PRs


idlira (approved):
Just need to have in mind if someone decides to rename labels around.

sabieber:
> Just need to have in mind if someone decides to rename labels around.

Yes but that should already not be done as it would break the auto-creation of GA release notes.

mkeckmkeck (approved):

PR #1959, opened 2024-04-02
Provides some groundwork for script based import customizations.

andyHa:
### Description

Provides some changes on existing code for the feature to be introduced seamlessly.

Relates to:
- https://github.com/scireum/sirius-biz/pull/1960

### Checklist


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -176,6 +176,20 @@ public String createProcess(@Nullable String type,
...
 
+    /**
+     * Fetches the currently active process.
+     *
+     * @return the process for which the current thread is executing or an empty optional if no process is active
+     */
+    public Optional<Supplier<Process>> fetchCurrentProcess() {
+        TaskContextAdapter adapter = TaskContext.get().getAdapter();
+        if (adapter instanceof ProcessEnvironment processEnvironment) {
+            return Optional.of(() -> fetchRequiredProcess(processEnvironment.getProcessId()));
```
sabieber, line 187: **Question:** Why doesn't this use the `Processes::fetchProcess` variant that given an Optional?

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

src/main/java/sirius/biz/process/Processes.java
andyHa: We actually return a Supplier<Process> here instead of a Process - as sometimes we only want to check if a process would exist, without actually fetching it. This method sort of hast two purposes, check if a process is present (and if so, fetch it in a second step, if required)...

PR #1964, opened 2024-04-05
Tenant Scripting Enhancements

idlira:
### Description

- Permits to disable scripts
 
image

- Fixes parameter initialization.
- The feature must be explicitly enabled in jobs by overwriting the `ImportBatchProcessFactory.enableScriptableEvents` method. For this to work, we changed the magic detection logic when only one script exists: with the feature disabled, no parameter is collected, thus the NOOP_DISPATCHER is used. If the parameter is collected but contains only one single entry, its hidden by default. One of the main advantages is, that even for hidden parameters the job will log the _default_ script used during its initialization, eg:

image

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-10898](https://scireum.myjetbrains.com/youtrack/issue/OX-10898)

### Checklist


src/main/resources/default/templates/biz/scripting/mongo-scripts.html.pasta
```diff
@@ -30,6 +31,9 @@
                             <td>
                                 <a href="@apply('/scripting/scripts/%s', script.getIdAsString())">@script.getCode()</a>
                             </td>
+                            <td>
+                                @toUserString(script.isDisabled())
```
sabieber, line 35: **Optional:** Maybe add a `t:dot` so its easier to detect disabled scripts via the color of the dot?

src/main/java/sirius/biz/scripting/mongo/MongoCustomScriptController.java
```diff
@@ -60,6 +55,7 @@ public void listScripts(WebContext webContext) {
      * Modifies / manages the given script.
      *
      * @param webContext the request to handle
+     * @param id         the id of the script to manage
```
jakobvogel, line 58:
```suggestion
     * @param id         the ID of the script to manage
```

src/main/java/sirius/biz/scripting/mongo/MongoCustomScriptController.java
```diff
@@ -75,6 +71,7 @@ public void editScript(WebContext webContext, String id) {
      * Deletes the given script.
      *
      * @param webContext the request to handle
+     * @param id         the id of the script to delete
```
jakobvogel, line 74:
```suggestion
     * @param id         the ID of the script to delete
```

sabieber (approved):

jakobvogel (approved):
The usual question: The use of `ne` constraints does not have adverse effects in terms of DB performance and the use of a search index?

sabieber (approved):

idlira:
> The usual question: The use of `ne` constraints does not have adverse effects in terms of DB performance and the use of a search index?



> The usual question: The use of `ne` constraints does not have adverse effects in terms of DB performance and the use of a search index?

Should not. In fact the new field is not even in the index, I will add it...
We could directly use `.eq(DISABLED, false)` instead since there are no productive data yet.

sabieber (approved):

jakobvogel (approved):
🙏

sabieber (approved):

src/main/java/sirius/biz/scripting/ScriptableEvents.java
```diff
@@ -53,11 +54,10 @@ public void handleEvent(ScriptableEvent event) {
...
      * @param name the name of the dispatcher to fetch
      * @return the dispatcher for the current tenant with the given name or a NOOP dispatcher if no such dispatcher
-     * exists. Note, if an empty <tt>name</tt> is given, the first available dispatcher for the current tenant is used.
-     * This way, if exactly one dispatcher is present, it will be used in all import processes etc.
+     * exists.
      */
-    public ScriptableEventDispatcher fetchDispatcherForCurrentTenant(@Nullable String name) {
-        if (dispatcherRepository == null) {
+    public ScriptableEventDispatcher fetchDispatcherForCurrentTenant(String name) {
+        if (dispatcherRepository == null || Strings.isEmpty(name)) {
```
andyHa, line 60: I think this breaks the "dont show parameter for single script" approach

idlira: I took a different approach hier: the parameter uses the `.hideWhen` method and hides itself when only one entry exists. This way we can differentiate if a single param exists and must be used OR if scripting has been disabled completely.

fhaScireum (approved):

src/main/java/sirius/biz/scripting/ScriptableEvents.java
andyHa: Understood - perfect +1

andyHa (approved):

PR #1965, opened 2024-04-08
Changes query behavior for /fs to from 'prefix' to 'contains'

mkeckmkeck:
### Description
Changes query behavior for /fs to contains
- the assumption is that this will make sense in many use cases e.g. filename = "theSpecialFilename.txt" will be found when searching for "special"
- as a side effect, this will workaround the problem that the prefix query for sql will not find exact matches if the filename contains uppercase characters

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-951](https://scireum.myjetbrains.com/youtrack/issue/SIRI-951)

### Checklist


idlira (approved):

fhaScireum (approved):

idlira (changes requested):
Changing my mind. In OXOMI we could easily have half a million files for one parent, and the search will no longer hit the DB index properly.
I would consider a parameter in the configuration of the storage space, so it's activated when known that the space only contains a reasonable amount of files per parent.

PR #1975, opened 2024-04-11
Permits to only support login via hashed API tokens

andyHa:
API tokens have already been deprecated for some time. Still some legacy systems need to support this approach. However, some systems now at least only rely in hashed tokens, which are way more secure as simply accepting a plaintext token which is also stored as plaintext in the DB.

Therefore, we now permit a more fine grained control over how kind of API tokens and authentication can be used.

Within the system config, next to "accept-api-tokens" one can also specify "accept-hashed-api-tokens". The former controls whether plaintext API keys can be used "just like a password" and the latter controls if a properly hashed version of the token is accepted. These two flags operate independently of each other.

NOTE: If "accept-api-tokens" was / is enabled, please review carefully, if now "accept-hashed-api-tokens" should be enabled, and also, if "accept-api-tokens" could be disabled after this change.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-](https://scireum.myjetbrains.com/youtrack/issue/SIRI-)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/tenants/TenantUserManager.java
+++ b/src/main/java/sirius/biz/tenants/TenantUserManager.java
@@ -169,6 +169,7 @@ public abstract class TenantUserManager<I extends Serializable, T extends BaseEn
 
     protected final String systemTenant;
     protected final boolean acceptApiTokens;
+    protected final boolean acceptHashedApiTokens;
     protected final List<String> availableLanguages;
 
     @Part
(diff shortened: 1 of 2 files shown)


ymo-sci (approved):

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -662,7 +664,7 @@ public UserInfo findUserByCredentials(@Nullable WebContext webContext, String us
             return result;
         }
 
-        if (acceptApiTokens && checkApiToken(loginData, password)) {
+        if (checkApiToken(loginData, password)) {
```
jakobvogel, line 667: The new version looks good to me. I don't understand, though, why there ever was a `acceptApiTokens` here. Does that setting change meaning with this PR?

jakobvogel (approved):

PR #1976, opened 2024-04-12
Lokalise: Translations update

scireumskip:
This pull request was initiated by Lokalise (user Marcus Korinth) at 2024-04-12 09:54:55


sabieber (approved):

mko-sci (approved):

idlira (approved):

src/main/resources/biz_en.properties
```diff
@@ -938,7 +938,7 @@ ScopeDefaultConfigController.showDefaultConfig = default configuration
 ScriptingController.jobMessage = Script with ID '${job}' was executed.
 StandardCategories.misc = General
 StandardCategories.monitoring = Monitoring
-StandardCategories.systemAdministration = system administration
+StandardCategories.systemAdministration = System administration
```
jakobvogel, line 941: Or both with capital letters in the beginning? (How do we do this for other Job categories?)

jakobvogel (approved):

src/main/resources/biz_en.properties
sabieber: Don't think we use title-case very often (or should do so) as this would make keys less versatile. If title casing is desired, this should be done via [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform?retiredLocale=de).

mkeckmkeck (approved):

fhaScireum (approved):

PR #1992, opened 2024-05-15
Introduce new permission to use the open-search and fix some related problems

ymo-sci:
### Description

- add category-links to templates
- fix security-sandbox restrictions
- fix texts in /kb welcome page and adjust them with the new permission
- switch order of user-accounts and tenants in search-results

Bildschirmfoto 2024-05-15 um 09 03 48

### Additional Notes
- This PR fixes or works on following ticket(s): [SE-13658](https://scireum.myjetbrains.com/youtrack/issue/SE-13658)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/resources/default/templates/biz/tycho/kb/welcome_de.html.pasta
+++ b/src/main/resources/default/templates/biz/tycho/kb/welcome_de.html.pasta
@@ -1,5 +1,6 @@
 <div class="pb-4">
-    <img src="/assets/images/login_logo.png" alt="@sirius.kernel.info.Product.getProduct().getName()"
+    <img src="/assets/images/login_logo.png"
+         alt="@sirius.kernel.info.Product.getProduct().getName()"
          style="max-height: 100px"/>
 </div>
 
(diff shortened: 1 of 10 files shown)


sabieber:
In regards to design i think an [external link](https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid) icon next to the plain category title would look cleaner and would not draw away focus from the card as much.

src/main/resources/default/templates/biz/tycho/kb/welcome_de.html.pasta
```diff
@@ -1,19 +1,26 @@
 <div class="pb-4">
-    <img src="/assets/images/login_logo.png" style="max-height: 100px"/>
+    <img src="/assets/images/login_logo.png" alt="@sirius.kernel.info.Product.getProduct().getName()"
+         style="max-height: 100px"/>
```
sabieber, line 3: **Convention:** Please put all attributes on separate lines when the line breaks

sabieber (approved):

idlira (commented):
Could we go with smaller fonts for the main category so they don't look so huge?

We also have a user-wish to show Tenants before Users. If that's OK, could we also tacke it here?

ymo-sci:
> Could we go with smaller fonts for the main category so they don't look so huge?
> 
> We also have a user-wish to show Tenants before Users. If that's OK, could we also tacke it here?

So that "Jobs" is smaller?
 - i´ll check it both.

sabieber (approved):

idlira (approved):

mko-sci (approved):

fhaScireum (approved):

PR #1987, opened 2024-05-07
Truncates incident messages when they exceed configured value

mko-sci:
### Additional Notes

- Reverts changes made in: https://github.com/scireum/sirius-biz/pull/1986
- This PR fixes or works on following ticket(s): [SIRI-960](https://scireum.myjetbrains.com/youtrack/issue/SIRI-960)

PR #1995, opened 2024-05-15
Truncates the error message of the stack trace to avoid huge es-entities

mko-sci:
### Description

PR https://github.com/scireum/sirius-biz/pull/1987 truncated the incident message, but the error message still appeared in its original length wiithin the stacktrace. This PR aims to truncate the message while preserving characters at the end of the message (since often this part is interesting).
The error message is usually found in the beginning and after the `Caused by:` statement...

Our typical stack traces start with the actual error message:
Bildschirmfoto 2024-05-15 um 12 59 53
followed by the actual trace:
Bildschirmfoto 2024-05-15 um 13 00 23
after the trace usually a cause is added, where the original error message might be repeated:
Bildschirmfoto 2024-05-15 um 13 00 47
the cause is then followed by its own trace:
Bildschirmfoto 2024-05-15 um 13 01 23

=> Therefore we need to make sure that both messages (exception and cause) are truncated. Also we need to take into account that an error message could be multiple lines long, for which we need to search until the start of the actual trace...

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-960](https://scireum.myjetbrains.com/youtrack/issue/SIRI-960)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/protocol/Protocols.java
+++ b/src/main/java/sirius/biz/protocol/Protocols.java
@@ -40,6 +40,11 @@ import java.util.logging.Level;
         framework = Protocols.FRAMEWORK_PROTOCOLS)
 public class Protocols implements LogTap, ExceptionHandler, MailLog {
 
+    /**
+     * Defines the maximum number of characters to preserve at the end of an exception message if it is truncated.
+     */
+    private static final int NUMBER_OF_CHARACTERS_TO_PRESERVE_AT_THE_END = 1000;
+
     private static final int DISABLE_ON_ERROR_PERIOD_MILLIS = 1000 * 60;
 
     /**
@@ -128,8 +133,8 @@ public class Protocols implements LogTap, ExceptionHandler, MailLog {
                 storedIncident.getMdc().put(tuple.getFirst(), tuple.getSecond());
             }
             storedIncident.setUser(UserContext.getCurrentUser().getProtocolUsername());
-            storedIncident.setMessage(Strings.limit(incident.getException().getMessage(), maxMessageLength, true));
-            storedIncident.setStack(NLS.toUserString(incident.getException()));
+            storedIncident.setMessage(truncateErrorMessageAndPreserveEnd(incident.getException().getMessage()));
+            storedIncident.setStack(truncateStacktrace(incident));
             storedIncident.setCategory(incident.getCategory());
             storedIncident.setLastOccurrence(LocalDateTime.now());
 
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/protocol/Protocols.java
```diff
@@ -140,6 +145,47 @@ public void handle(Incident incident) throws Exception {
...
 
+    private String truncateStacktrace(Incident incident) {
+        boolean foundStackTrace = false;
+        StringBuilder result = new StringBuilder();
+        StringBuilder stack = new StringBuilder();
+        StringBuilder errorMesssage = new StringBuilder();
+
+        for(String line: NLS.toUserString(incident.getException()).split("\\r?\\n")) {
+            if (!foundStackTrace) {
+                if (line.contains("at") && line.contains(".java")) {
```
idlira, line 156: Don't u wanna add white spaces around "at", since you are probably looking for `something at somewhere`?

idlira (approved):

src/main/java/sirius/biz/protocol/Protocols.java
mko-sci: Bildschirmfoto 2024-05-15 um 12 54 03

Nope. I am trying to find every line of the original error message in case that the error message has several new line chars.
Therefore I am looking for the first line that contains an "at" and an ".java" as these lines mark the start of the actual stacktrace.

mko-sci: Bildschirmfoto 2024-05-15 um 12 55 56

Then I am trying to add up the stacktrace until the occurence of a "Caused by:" as the error message might be repeated here. Then I repeat the algorithm to find the error message and the stacktrace of the cause.

mko-sci:
I know the algorithm looks a bit clunky, I am open to your thoughts or ideas in that regard. =)

src/main/java/sirius/biz/protocol/Protocols.java
```diff
@@ -140,6 +145,47 @@ public void handle(Incident incident) throws Exception {
         }
     }
 
+    private String truncateStacktrace(Incident incident) {
+        boolean foundStackTrace = false;
+        StringBuilder result = new StringBuilder();
+        StringBuilder stack = new StringBuilder();
+        StringBuilder errorMesssage = new StringBuilder();
+
+        for(String line: NLS.toUserString(incident.getException()).split("\\r?\\n")) {
```
jakobvogel, line 154: Code format! (No space between `for` and the bracket.) – The check boxes are there to remind you of that, rather than to be checked automatically. 😜

jakobvogel: I would prefer `NLS.toUserString(incident.getException()).lines().forEach(line -> {…});` as it saves the non-compiled regular expression.

src/main/java/sirius/biz/protocol/Protocols.java
jakobvogel: I agree with @idlira. The check seems to be a bit too sketchy to me. You could check whether you have whitespace, the literal `at`, more whitespace, a Java symbol name, and then, in brackets, a file name and a line number. While a full check may be too much, your check does not really do justice to the regular pattern of such a stack-trace line

src/main/java/sirius/biz/protocol/Protocols.java
jakobvogel: That would also allow to use early returns instead of the nested `ifs` and `elses`.

src/main/java/sirius/biz/protocol/Protocols.java
```diff
@@ -140,6 +145,47 @@ public void handle(Incident incident) throws Exception {
...
+            if (!foundStackTrace) {
+                if (line.contains("at") && line.contains(".java")) {
+                    foundStackTrace = true;
+                    stack.append(line).append("\n");
+                } else {
+                    errorMesssage.append(line).append("\n");
+                }
+            } else {
+                if (line.contains("Caused by:")) {
```
jakobvogel, line 163: Rather starts with?

jakobvogel (commented):
The screenies seem to describe the previous situation. The result of the new code would have been more interesting

mko-sci:
This code is a bit too janky in hindsight. I am working on a better solution.

PR #1999, opened 2024-05-24
SQL: Show num rows and limit as info

sabieber:
### Description

And also move it to above the list, to make it more accessible.

image

image

image

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/jdbc/DatabaseController.java
+++ b/src/main/java/sirius/biz/jdbc/DatabaseController.java
@@ -134,9 +134,10 @@ public class DatabaseController extends BasicController {
             } else if (isModifyStatement(sqlStatement)) {
                 output.property("rowModified", query.executeUpdate());
             } else {
+                int effectiveLimit = webContext.get("limit").asInt(DEFAULT_LIMIT);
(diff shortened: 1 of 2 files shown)


fhaScireum (approved):

idlira (approved):

mkeckmkeck (approved):

ymo-sci (approved):
if it is limited to 1000, but more than 1000 results, shouldn't we print out 
"xxxx (limited to 1000) here" ?

fhaScireum:
> if it is limited to 1000, but more than 1000 results, shouldn't we print out "xxxx (limited to 1000) here" ?

Or we could show e.g. "24 rows" without hinting at the limit as it is superflous and if we reach the 1000 result, show "1000 rows (reached row limit)" or something similar.

sabieber:
We can't know if we exactly hit the limit or are over it and rows are truncated without some logic in the controller/query that queries limit + 1 and then omits the last row, etc.

As this was just a nice to have improvement I tried to keep it simple, but if enough of you want this I can implement it.

jakobvogel (approved):

PR #2000, opened 2024-05-24
only link delete files job on deletable directories.

ymo-sci:
### Description

Some directories are linked as root-directories or linked only readonly where we cannot delete files, so the linking to the job is not helpful.
Bildschirmfoto 2024-05-24 um 17 36 31
before vs now

### Checklist


idlira (changes requested):
This is required in OXOMI so we can bulk delete child files/folders starting from the root (of course the root itself cannot be deleted)

ymo-sci:
Currently not needed, may be possible via config.

PR #2001, opened 2024-05-28
Adds comment for results on SQL-performance regarding countIf- and GROUPBY-queries to framework

scireum-mbo:
### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11019](https://scireum.myjetbrains.com/youtrack/issue/OX-11019)
- This PR is related to PR: https://github.com/scireum/oxomi/pull/11214

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/analytics/explorer/EventTimeSeriesComputer.java
+++ b/src/main/java/sirius/biz/analytics/explorer/EventTimeSeriesComputer.java
@@ -106,6 +106,7 @@ public class EventTimeSeriesComputer<O, E extends Event<E>> implements TimeSerie
      * @see #AGGREGATION_EXPRESSION_COUNT
      */
     public EventTimeSeriesComputer<O, E> addAggregation(String expression, @Nullable String label) {
+        // On first sight using a countIf would be way slower than a GROUPBY, but Clickhouse optimizes this so well that the query is way faster with a countIf
         this.aggregations.add(Tuple.create(expression, label));
         return this;
     }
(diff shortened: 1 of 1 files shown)


sabieber (approved):
**Optional:** Would suggest moving it up into the JavaDoc (with a `<p>` between the first line and the new sentence. This way its also visible when invoking the quick docs (F1) where the method is called.

src/main/java/sirius/biz/analytics/explorer/EventTimeSeriesComputer.java
```diff
@@ -106,6 +106,7 @@ public EventTimeSeriesComputer(Class<E> eventType) {
      * @see #AGGREGATION_EXPRESSION_COUNT
      */
     public EventTimeSeriesComputer<O, E> addAggregation(String expression, @Nullable String label) {
+        // On first sight using a countIf would be way slower than a GROUPBY, but Clickhouse optimizes this so well that the query is way faster with a countIf
```
idlira, line 109: I would make this part of the JavaDoc, not an inline comment.
Secondly I would describe it a bit deeper. Will post a text suggestion here soon...

idlira (changes requested):

src/main/java/sirius/biz/analytics/explorer/EventTimeSeriesComputer.java
idlira: JavaDoc suggestion:
```java
    /**
     * Adds an aggregation expression to collect.
     * <p>
     * In likely most of the cases, the aggregations used with this method are conditional
     * count functions such as <code>countIf(<CLAUSE>)</code>. Performance measurements in Clickhouse
     * showed faster performance using such functions instead of regular <code>GROUP BY</code>
     * operators.
     *
     * @param expression the SQL expression to add to the <tt>SELECT</tt> clause
     * @param label      the label to use or <tt>null</tt> to use the default label
     * @return the computer itself for fluent method calls
     * @see #AGGREGATION_EXPRESSION_COUNT
     */
```

sabieber (approved):

idlira (approved):

PR #2018, opened 2024-07-10
SIRI-975: Uniformly take deprecated Lookup Entries into account or not

fhaScireum:
### Description

The autocomplete route already filtered out deprecated values. The implementations of LookupTable also filtered out deprecated values when using LookupTable#suggest with a filled search term. However LookupTable#scan and LookupTable#suggest without a search term wouldn't filter out deprecated values as it used the #scan method, thus introducing inconsistencies in handling deprecated values.

Whether or not deprecated values should be considered is now handled at low level to prevent:
- Filtering deprecated values multiple times
- Not filtering out deprecated values when no search term is present ("scan" method is used then), but filtering them once the user typed something (e.g. in case a custom autocomplete route is used).
- Limiting the stream of entities before filtering some of them out leaving the user with less suggestions as they should be shown.

The changes are effectively NON breaking, as the existing route is still available (with default to filter out deprecated values) as well as all existing the LookupTable#scan (defaults to filtering out deprecated values) and LookupTable#suggest methods (defaults to not filter out deprecated values).

Note, that only IDB table lookup was tested locally, but it is also the only table with major adjustments.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-975](https://scireum.myjetbrains.com/youtrack/issue/SIRI-975)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/codelists/IDBLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/IDBLookupTable.java
@@ -277,6 +277,7 @@ class IDBLookupTable extends LookupTable {
                                                       String language,
                                                       boolean considerDeprecatedValues) {
         try {
+            // Caveat: IDBTable$QueryBuilder#manyRows can only be used when no filtering is applied afterwards.
             if (considerDeprecatedValues) {
                 return table.query()
                             .searchInAllFields()
@@ -315,6 +316,7 @@ class IDBLookupTable extends LookupTable {
     @Override
     public Stream<LookupTableEntry> scan(String language, Limit limit, boolean considerDeprecatedValues) {
         try {
+            // Caveat: IDBTable$QueryBuilder#manyRows can only be used when no filtering is applied afterwards.
             if (considerDeprecatedValues) {
                 return table.query()
                             .translate(language)
--- a/src/main/java/sirius/biz/codelists/IDBLookupTable.java
+++ b/src/main/java/sirius/biz/codelists/IDBLookupTable.java
@@ -272,14 +272,30 @@ class IDBLookupTable extends LookupTable {
     }
 
     @Override
-    protected Stream<LookupTableEntry> performSuggest(Limit limit, String searchTerm, String language) {
+    protected Stream<LookupTableEntry> performSuggest(Limit limit,
+                                                      String searchTerm,
+                                                      String language,
+                                                      boolean considerDeprecatedValues) {
         try {
+            if (considerDeprecatedValues) {
+                return table.query()
+                            .searchInAllFields()
+                            .searchValue(searchTerm)
+                            .translate(language)
+                            .manyRows(limit, codeField, nameField, descriptionField)
+                            .map(row -> new LookupTableEntry(row.at(0).asString(),
+                                                             row.at(1).asString(),
+                                                             row.at(2).getString()));
+            }
+
             return table.query()
                         .searchInAllFields()
                         .searchValue(searchTerm)
                         .translate(language)
-                        .manyRows(limit, codeField, nameField, descriptionField, COL_DEPRECATED)
+                        .allRows(codeField, nameField, descriptionField, COL_DEPRECATED)
                         .filter(row -> row.at(3).asLong(0) != 1L)
+                        .skip(limit.getItemsToSkip())
+                        .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems())
                         .map(row -> new LookupTableEntry(row.at(0).asString(),
                                                          row.at(1).asString(),
                                                          row.at(2).getString()));
@@ -297,11 +313,21 @@ class IDBLookupTable extends LookupTable {
     }
 
     @Override
-    public Stream<LookupTableEntry> scan(String language, Limit limit) {
+    public Stream<LookupTableEntry> scan(String language, Limit limit, boolean considerDeprecatedValues) {
         try {
+            if (considerDeprecatedValues) {
+                return table.query()
+                            .translate(language)
+                            .manyRows(limit, codeField, nameField, descriptionField, COL_DEPRECATED, COL_SOURCE)
+                            .map(this::processSearchOrScanRow);
+            }
+
             return table.query()
                         .translate(language)
-                        .manyRows(limit, codeField, nameField, descriptionField, COL_DEPRECATED, COL_SOURCE)
+                        .allRows(codeField, nameField, descriptionField, COL_DEPRECATED, COL_SOURCE)
+                        .filter(row -> row.at(3).asLong(0) != 1L)
+                        .skip(limit.getItemsToSkip())
+                        .limit(limit.getMaxItems() == 0 ? Long.MAX_VALUE : limit.getMaxItems())
                         .map(this::processSearchOrScanRow);
         } catch (Exception exception) {
             Exceptions.createHandled()
--- a/src/main/java/sirius/biz/codelists/LookupTable.java
+++ b/src/main/java/sirius/biz/codelists/LookupTable.java
@@ -763,18 +763,6 @@ public abstract class LookupTable {
     }
 
     /**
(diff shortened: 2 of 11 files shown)


mko-sci (approved):

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -771,45 +784,90 @@ public Stream<LookupTableEntry> suggest(@Nullable String searchTerm) {
...
     /**
-     * Enumerates all entries in the table using the current language.
+     * Enumerates all entries in the table aside from deprecated entries using the current language.
      *
      * @param limit the limit to apply to fetch a sane number of entries
      * @return a stream of all entries in this table, or an empty stream is scanning isn't supported
      */
     public Stream<LookupTableEntry> scan(Limit limit) {
-        return scan(NLS.getCurrentLanguage(), limit);
+        return scan(NLS.getCurrentLanguage(), limit, true);
```
idlira, line 831: true? thought default was false...

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -771,45 +784,90 @@ public Stream<LookupTableEntry> suggest(@Nullable String searchTerm) {
...
+    /**
+     * Enumerates all entries in the table aside from deprecated entries using the given language.
      *
      * @param language the language to translate the name and description to
      * @param limit    the limit to apply to fetch a sane number of entries
      * @return a stream of the selected amount of entries in this table
      */
-    public abstract Stream<LookupTableEntry> scan(String language, Limit limit);
+    public Stream<LookupTableEntry> scan(String language, Limit limit) {
+        return scan(language, limit, true);
```
idlira, line 853: true?

src/main/java/sirius/biz/codelists/LookupTableController.java
```diff
@@ -51,6 +51,26 @@ public void suggestFromLookupTable(WebContext webContext,
...
+     *
+     * @param webContext               the web requests calling the autocomplete service
+     * @param tableName                the name of the table for which suggestions should be gathered
+     * @param display                  the requested {@link sirius.biz.codelists.LookupValue.Display display mode} for
+     *                                 the field label
+     * @param extendedDisplay          the requested {@link sirius.biz.codelists.LookupValue.Display display mode} for
+     *                                 the completion label
+     * @param considerDeprecatedValues whether deprecated values should be suggested
+     */
+    @Routed("/system/lookuptable/autocomplete/:1/:2/:3/:4")
```
sabieber, line 68: **Optional:** Should this really be part of the URL or not rather be a parameter that is read from the context? This doesn't feel very RESTy and would be even less breaking that way

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -834,7 +892,7 @@ public Stream<LookupTableEntry> search(@Nullable String searchTerm, Limit limit)
      */
     public Stream<LookupTableEntry> search(@Nullable String searchTerm, Limit limit, String language) {
         if (Strings.isEmpty(searchTerm)) {
-            return scan(language, limit);
+            return scan(language, limit, true);
```
idlira, line 895: true?

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -751,18 +751,31 @@ public static Map<String, ObjectNode> parseMap(ObjectNode root, JsonPointer path
...
+     * Suggests several entries for the given search term using the currently active language.
+     *
+     * @param searchTerm               the term used to filter the suggestions
+     * @param considerDeprecatedValues whether deprecated values should be suggested as well
+     * @return a stream of suggestions for the given term. Note that most probably {@link Stream#limit(long)} should
+     * be used on the result as this might yield quite a bunch of suggestions in order to optimize internal queries.
+     */
+    public Stream<LookupTableEntry> suggest(@Nullable String searchTerm, boolean considerDeprecatedValues) {
+        return suggest(searchTerm, NLS.getCurrentLanguage(), considerDeprecatedValues);
+    }
```
sabieber, line 775: **Optional:** This overload feels a bit pointless in my opinion, as `NLS.getCurrentLanguage()` is available to be called from everywhere (even in templates now as we can call static functions). Would prefer to keep the number of overloads lower in this case as there already are quite a bunch of them, which can get confusing when trying to decide which one is the right one to call.

src/main/java/sirius/biz/codelists/LookupTable.java
```diff
@@ -771,45 +784,90 @@ public Stream<LookupTableEntry> suggest(@Nullable String searchTerm) {
...
-     * Enumerates all entries in the table using the given language.
+     * Enumerates all entries in the table using the current language.
+     *
+     * @param limit                    the limit to apply to fetch a sane number of entries
+     * @param considerDeprecatedValues whether deprecated values should be included in the scan
+     * @return a stream of all entries in this table, or an empty stream is scanning isn't supported
+     */
+    public Stream<LookupTableEntry> scan(Limit limit, boolean considerDeprecatedValues) {
+        return scan(NLS.getCurrentLanguage(), limit, considerDeprecatedValues);
+    }
```
sabieber, line 843: **Optional:** same as above

src/main/java/sirius/biz/codelists/IDBLookupTable.java
```diff
@@ -272,14 +272,30 @@ protected <T> T makeObject(Class<T> type, ObjectNode jsonData) {
...
+            if (considerDeprecatedValues) {
+                return table.query()
+                            .searchInAllFields()
+                            .searchValue(searchTerm)
+                            .translate(language)
+                            .manyRows(limit, codeField, nameField, descriptionField)
+                            .map(row -> new LookupTableEntry(row.at(0).asString(),
+                                                             row.at(1).asString(),
+                                                             row.at(2).getString()));
+            }
```
sabieber, line 289: **Optional:** I think it would be nice to add the caveat/reason for this if-statement (that is mentioned in the commit) as a small comment to prevent accidental breakage via refactoring/cleanup

src/main/java/sirius/biz/codelists/IDBLookupTable.java
```diff
@@ -297,11 +313,21 @@ protected Stream<LookupTableEntry> performSuggest(Limit limit, String searchTerm
...
     @Override
-    public Stream<LookupTableEntry> scan(String language, Limit limit) {
+    public Stream<LookupTableEntry> scan(String language, Limit limit, boolean considerDeprecatedValues) {
         try {
+            if (considerDeprecatedValues) {
+                return table.query()
+                            .translate(language)
+                            .manyRows(limit, codeField, nameField, descriptionField, COL_DEPRECATED, COL_SOURCE)
+                            .map(this::processSearchOrScanRow);
+            }
```
sabieber, line 323: **Optional:** same as above

sabieber (approved):

src/main/java/sirius/biz/codelists/LookupTable.java
fhaScireum: `true` is the default when it comes to all `scan()` methods as the implementations never filtered out deprecated values there.

Look at e.g. `sirius.biz.codelists.IDBLookupTable#scan(String, Limit)`. While it did mark deprecated entries as such, it never filtered them out. Same goes for the other implementations.

src/main/java/sirius/biz/codelists/LookupTable.java
fhaScireum: s. a.

src/main/java/sirius/biz/codelists/LookupTable.java
fhaScireum: s. a.

sabieber (approved):
Thanks a lot 👌🏻

mkeckmkeck (approved):

jmuscireum (approved):

PR #2022, opened 2024-07-18
BREAKING: change the way saving processTypes.

ymo-sci:
until now we always stored getClass().getSimpleName() + ".label" = MyJobFactory.label in the processType field on processes. For overridden implementations of e.g. abstract jobs this leads to not make the processes filterable on /ps. Now we will use the job#getName and store it as the processType. So the filtering url no longer is `/ps?processType=BmecatImportJobFactory.label` but `/ps?processType=import-bmecat`
To migrate existing values we implement a temporary job which resolves all available jobs and resolves the correct new type.

Bildschirmfoto 2024-07-18 um 18 37 58
Bildschirmfoto 2024-07-18 um 18 39 51

Patch-Task: Execute the job in simulation-mode (e.g. on a staging-system) and maybe adjust missing NLS-Keys.
Then execute it with simulation: no to fix all existing values.

BREAKING: If the job isn't executed - old jobs are not visible filterable by type

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-990](https://scireum.myjetbrains.com/youtrack/issue/SIRI-990)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/jobs/FixProcessTypes.java
+++ b/src/main/java/sirius/biz/jobs/FixProcessTypes.java
@@ -18,6 +18,7 @@ import sirius.biz.process.logs.ProcessLog;
 import sirius.biz.tenants.TenantUserManager;
 import sirius.db.es.Elastic;
 import sirius.kernel.commons.Strings;
+import sirius.kernel.commons.Watch;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.PriorityParts;
 import sirius.kernel.di.std.Register;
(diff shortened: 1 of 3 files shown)


mko-sci (commented):
What if the job crashed in the middle?
Is it safe to just re-run it?

src/main/java/sirius/biz/jobs/FixProcessTypes.java
```diff
@@ -0,0 +1,134 @@
...
+
+        process.log(ProcessLog.info().withMessage(Strings.apply("Found %s jobs", jobTypesAndKeys.size())));
+        jobTypesAndKeys.forEach((type, nlsKey) -> {
+            AtomicInteger counter = new AtomicInteger();
+            if (!NLS.exists(nlsKey, null)) {
+                process.log(ProcessLog.warn()
+                                      .withMessage(Strings.apply("No NLS key found for type %s (%s)", type, nlsKey)));
+            }
+
+            elastic.select(Process.class).eq(Process.PROCESS_TYPE, nlsKey).iterateAll(executedProcess -> {
```
idlira, line 80: I think its smarter to streamBlockwise all processes. Per process you lookup the right key for it (and add to a local cache). 
Streamblockwise sorts per ID and is abortable. If the job stops, print the last process ID and add an optional param to resume from the given optional ID (ID >= x). In OX we are talking about 580K procs!
And use the progress tracker to have a hint of execution time.

idlira (changes requested):

src/main/java/sirius/biz/jobs/FixProcessTypes.java
idlira: Also consider already migrated jobs and skip the ones with processType.endsWith(".label"). This will also discard new jobs started after the change.

sabieber (approved):

src/main/java/sirius/biz/jobs/FixProcessTypes.java
ymo-sci: All Jobs which have been migrated already or executed after the change will not be checked a second time, because the processType is no longer MyJobFactory.label
A crash might be possible, maybe a count in the check-mode will be safer than iterating.

mko-sci (approved):

fhaScireum (approved):

mkeckmkeck (approved):

PR #2029, opened 2024-08-19
Skips deleted onboarding videos from deletion marking

idlira:
### Description

Already deleted videos (deleted = true) don't need to be marked again. This might cause the following error if the video has not been physically deleted yet when the recompute loops run:
```
Ein Fehler ist aufgetreten: Tried to update the changed entity MONGOACADEMYVIDEO-2HB7AN57L814PM709QGAFBDDFO (2HB7AN57L814PM709QGAFBDDFO), but actually nothing was changed in the database! There might be an error in one of its properties' transform or equals methods, as the framework indicated a changed property. The following properties are considered changed: academyVideoData_lastUpdated
```

### Checklist


sabieber (approved):

ymo-sci (approved):

jakobvogel (approved):
Are there entities in the database where the `DELETED` flag is missing? (I.e. is "`DELETED == false`" a safe constraint?)

idlira:
> Are there entities in the database where the `DELETED` flag is missing? (I.e. is "`DELETED == false`" a safe constraint?)

Yes, it's safe. The field exists since the entity has been created. There are no entries with DELETED = null.

scireum-mbo (approved):

fhaScireum (approved):

PR #2035, opened 2024-09-25
User Accounts - Extends user account cards and open search results by the option to jump to a tenant if the user has the support role

scireum-mbo:
### Description
After:

https://github.com/user-attachments/assets/66343e26-547d-4c1c-b842-d9ac3ec165f1

https://github.com/user-attachments/assets/5957cfd8-0177-477a-9c33-7f71aec005ce

User accounts:
image

Open Search:
image

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1020](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1020)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/tenants/UserAccountSearchProvider.java
+++ b/src/main/java/sirius/biz/tenants/UserAccountSearchProvider.java
@@ -100,7 +100,10 @@ public abstract class UserAccountSearchProvider<I extends Serializable, T extend
                                                               <i:arg name="user" type="sirius.biz.tenants.UserAccount"/>
                                                               @user (
                                                               <i:if test="UserContext.get().getUser().hasPermission(sirius.biz.tenants.TenantUserManager.PERMISSION_SELECT_TENANT)">
-                                                                <t:smartValue type="tenant" payload="@user.getTenantAsString()" label="@user.getTenant().fetchValue().toString()"/>
+                                                                <t:smartValue type="tenant"
+                                                                    id="@generateId('smarty-useraccount-%s')"
+                                                                    payload="@user.getTenantAsString()"
+                                                                    label="@user.getTenant().fetchValue().toString()"/>
                                                                 <i:else>
                                                                     @user.getTenant().fetchValue().toString()
                                                                 </i:else>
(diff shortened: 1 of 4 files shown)


src/main/resources/default/templates/biz/tenants/user-accounts-list.html.pasta
```diff
@@ -47,6 +47,13 @@
                                                     url="@apply('/user-accounts/select/%s', account.getIdAsString())"/>
                                 </t:dropdownSection>
                             </i:if>
+                            <i:if test="UserContext.get().getUser().hasPermission('flag-system-support')">
```
idlira, line 50: The permission `flag-system-support` is an OXOMI thing. But can't we just use the `permission-select-tenant` instead? Question is valid for all places you evaluate this flag.

idlira (changes requested):

sabieber:
Another idea (I know it was not mentioned in the ticket):
Instead of using a separate card action we could also show the tenant name with `t:smartValue` (like in the process view), which shows the tenant name and opens a tooltip on click which allows to edit or switch to the tenant.

grafik

Something like: `<t:smartValue type="tenant" payload="@process.getTenantId()" label="@process.getTenantName()" />`

src/main/java/sirius/biz/tenants/UserAccountSearchProvider.java
```diff
@@ -101,7 +101,7 @@ public void query(String query, int maxResults, Consumer<OpenSearchResult> resul
                                                               @user (@user.getTenant().fetchValue().toString())
                                                               <br>
                                                               <a href="/user-accounts/select/@user.getIdAsString()" class="card-link">@i18n("TenantController.select")</a>
-                                                              <i:if test="UserContext.get().getUser().hasPermission('flag-system-support')">
+                                                              <i:if test="UserContext.get().getUser().hasPermission('permission-select-tenant')">
```
idlira, line 104: Please use the Java constant `TenantUserManager.PERMISSION_SELECT_TENANT` instead of the hard-coded string in the templates. You will find examples of if using PERMISSION_SELECT_USER_ACCOUNT for instance.

idlira (changes requested):

idlira (approved):

sabieber:
Could you add some screenshots how the new solution looks like?

mko-sci (approved):

mkeckmkeck (approved):

ymo-sci (approved):

sabieber:
On your screenshot of the new Open Search result: Isn't the Text above in Parenthesis the Tenant Name? If so I would prefer that this is turned into the smartValue instead of an additional action (same for the content cards in OX)

idlira (approved):

sabieber (approved):

idlira (approved):

PR #2060, opened 2024-12-09
Processes - Waiting State

sabieber:
### Description
This will indicate a process that is created but not yet actively processed. This may be the case when the maximum number of parallel processes in the queue of the processes is currently execeeded.

grafik

grafik

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-681](https://scireum.myjetbrains.com/youtrack/issue/SIRI-681)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/process/Processes.java
+++ b/src/main/java/sirius/biz/process/Processes.java
@@ -505,6 +505,19 @@ public class Processes {
     }
 
     /**
+     * Marks a process as running.
+     *
+     * @param processId the process to update
+     * @return <tt>true</tt> if the process was successfully modified, <tt>false</tt> otherwise
+     */
+    protected boolean markRunning(String processId) {
+        return modify(processId, process -> process.getState() == ProcessState.WAITING, process -> {
+            process.setStarted(LocalDateTime.now());
+            process.setState(ProcessState.RUNNING);
+        });
+    }
+
+    /**
      * Marks a process as canceled.
      * <p>
      * Note that this also marks the process as {@link Process#ERRORNEOUS}.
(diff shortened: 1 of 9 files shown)


src/main/java/sirius/biz/process/Processes.java
```diff
@@ -503,6 +504,20 @@ protected boolean updateState(String processId, ProcessState newState) {
...
+    /**
+     * Marks a process as running.
+     *
+     * @param processId the process to update
+     * @return <tt>true</tt> if the process was successfully modified, <tt>false</tt> otherwise
+     */
+    protected boolean markRunning(String processId) {
+        return modify(processId, process -> process.getState() == ProcessState.WAITING, process -> {
+            process.setStarted(LocalDateTime.now());
+            process.setWaitingTime((int) Duration.between(process.getCreated(), process.getStarted()).getSeconds());
```
jakobvogel, line 516: As far as I understand it, the waiting time is only set when a process is started, right? Should we also set it when canceling a process before it is started? Can we/should we distinguish between "canceled while waiting" vs "canceled while running"?

jakobvogel (approved):
🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏

mkeckmkeck (approved):

src/main/java/sirius/biz/process/Processes.java
sabieber: That's a good question, on one hand it feels right to set the field also when a job is cancelled. On the other hand, it would decrease the average waiting time when analyzing without representing the real time the process would have to wait until actual execution.

idlira (approved):

scireum-mbo (approved):

mko-sci (approved):

fhaScireum (approved):

PR #2064, opened 2024-12-16
Bumps sirius-biz and applies breaking changes

idlira:
### Description

Marking as BREAKING since this applies a breaking change from sirius-biz.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1039](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1039)
- This PR is related to PR: https://github.com/scireum/sirius-db/pull/668

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/jdbc/SQLVariant.java
+++ b/src/main/java/sirius/biz/storage/layer2/jdbc/SQLVariant.java
@@ -109,7 +109,7 @@ public class SQLVariant extends SQLEntity implements BlobVariant {
 
     @AfterDelete
     protected void onDelete() {
-        SQLBlob sqlBlob = sourceBlob.fetchValue();
+        SQLBlob sqlBlob = sourceBlob.fetchCachedValue();
         if (sqlBlob == null) {
             return;
         }
--- a/src/main/java/sirius/biz/tenants/SAMLController.java
+++ b/src/main/java/sirius/biz/tenants/SAMLController.java
@@ -181,7 +181,7 @@ public class SAMLController<I extends Serializable, T extends BaseEntity<I> & Te
 
     private void verifyUser(SAMLResponse response, UserInfo user) {
         U account = user.getUserObject(getUserClass());
-        T tenant = account.getTenant().fetchValue();
+        T tenant = account.getTenant().fetchCachedValue();
 
         if (!checkIssuer(tenant, response)) {
             throw Exceptions.createHandled().withSystemErrorMessage("SAML Error: Issuer mismatch!").handle();
--- a/src/main/java/sirius/biz/tenants/TenantController.java
+++ b/src/main/java/sirius/biz/tenants/TenantController.java
@@ -407,7 +407,7 @@ public abstract class TenantController<I extends Serializable, T extends BaseEnt
                         .causedByUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
                         .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
                         .forTenant(account.getTenant().getIdAsString(),
-                                   account.getTenant().fetchValue().getTenantData().getName())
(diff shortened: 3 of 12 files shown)


src/main/java/sirius/biz/storage/layer2/jdbc/SQLVariant.java
```diff
@@ -109,7 +109,7 @@ public class SQLVariant extends SQLEntity implements BlobVariant {
 
     @AfterDelete
     protected void onDelete() {
-        SQLBlob sqlBlob = sourceBlob.fetchValue();
+        SQLBlob sqlBlob = sourceBlob.fetchCachedValue();
```
sabieber, line 112: **Question:** Doesn't use of the ref-cache here potentially lead to trying to delete a blob that may be already deleted otherwise?

src/main/java/sirius/biz/tenants/SAMLController.java
```diff
@@ -181,7 +181,7 @@ protected List<T> querySAMLTenants() {
 
     private void verifyUser(SAMLResponse response, UserInfo user) {
         U account = user.getUserObject(getUserClass());
-        T tenant = account.getTenant().fetchValue();
+        T tenant = account.getTenant().fetchCachedValue();
```
sabieber, line 184: **Question:** Wouldn't it be better to use the cache of the `Tenants` part here?

src/main/java/sirius/biz/tenants/TenantController.java
```diff
@@ -407,7 +407,7 @@ public void selectTenant(final WebContext webContext, String tenantId) {
                         .causedByUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
                         .forUser(account.getUniqueName(), account.getUserAccountData().getLogin().getUsername())
                         .forTenant(account.getTenant().getIdAsString(),
-                                   account.getTenant().fetchValue().getTenantData().getName())
+                                   account.getTenant().fetchCachedValue().getTenantData().getName())
```
sabieber, line 410: see above

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -453,7 +453,7 @@ public UserInfo findUserByName(@Nullable WebContext webContext, String user) {
         U account = optionalAccount.get();
 
         userAccountCache.put(account.getUniqueName(), account);
-        tenantsCache.put(account.getTenant().fetchValue().getIdAsString(), account.getTenant().fetchValue());
+        tenantsCache.put(account.getTenant().fetchCachedValue().getIdAsString(), account.getTenant().fetchCachedValue());
```
sabieber, line 456: Don't think its a good idea to fill a cache with a cached value

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -569,7 +569,8 @@ protected U fetchAccount(@Nonnull String accountId) {
                 return null;
             }
             userAccountCache.put(account.getUniqueName(), account);
-            tenantsCache.put(account.getTenant().fetchValue().getIdAsString(), account.getTenant().fetchValue());
+            tenantsCache.put(account.getTenant().fetchCachedValue().getIdAsString(),
```
sabieber, line 572: see above

sabieber (commented):
Don't think the idea was to simply use `fetchCachedValue` everywhere, but rather look at each individual usage and choose the correct/safer variant

src/main/java/sirius/biz/storage/layer2/jdbc/SQLVariant.java
idlira: True, will review the places.

src/main/java/sirius/biz/tenants/SAMLController.java
idlira: It is. Extrapolates the purpose of the change/ticket though which requires additional/special testing.

PR #2079, opened 2025-01-20
HTTPS URL Property Validator

jakobvogel:
### Description

Adds a property validator that enforces a valid HTTPS URL.

Note that the entire property validation framework may be subject to some maintenance soon in order to improve integration with Tycho (and file structure). See [SIRI-1050](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1050).

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11244](https://scireum.myjetbrains.com/youtrack/issue/OX-11244)
- This PR is related to PR: https://github.com/scireum/sirius-kernel/pull/546

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
+++ b/src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
@@ -16,8 +16,6 @@ import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
 
 import javax.annotation.Nonnull;
-import java.net.URI;
-import java.net.URL;
 import java.util.function.Consumer;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
```diff
@@ -0,0 +1,99 @@
...
+     * @see #parseUrl(Property, Object)
+     * @see #createInvalidHttpsUrlException(Property, String)
+     */
+    private URL parseHttpsUrl(Property property, Object value) {
+        URL url = parseUrl(property, value);
+        if (url == null) {
+            return null;
+        }
+
+        if (!"https".equalsIgnoreCase(url.getProtocol())) {
```
idlira, line 58: I think Yannick build recently some central checks about http checks, would be good to reuse those.

jakobvogel: Only if there is a non-trivial benefit. URL parsing is securely done via a Standard Library method, leaving only the protocol comparison as "custom code". IMHO, that is too trivial to invoke yet another helper.

idlira: Hm, I don't see it this way. It was done exactly to avoid 1.000 places checking if a string starts with http. BTW, its in `Strings.isHttpUrl`. *but* it does not enforce https and allows http as well 🫣
My vote is to add a `isHttpsUrl` method among with it, but its just my opinion.

jakobvogel: …and it does not even check whether it is a valid URL after the prefix at all, accepting `https:// ;%@@ lol whatever i don't care` as valid URL, at the same time (ab-)using fully-fledged pattern matching for a simple case-insensitive prefix match. And that is the kind of faulty code we should use everywhere instead of standard methods?

We are approaching the test week. We need to get stuff done. Basically, you attempt to force me into bugfixing another project, open another PR there, wait for it to be processed, update this here, so that I can finally continue the work I am supposed to have done in time for the patch, just to move perfectly fine and effectively trivial code from this PR somewhere else. Honestly, under these circumstances, we really can save the discussions in future retros about planning and extra-hours and do something more productive instead.

But fine, then I'll work on `sirius-kernel` instead of the OX ticket.

src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void validate(Property property, Object value, Consumer<String> validationConsumer) {
+        if (value instanceof String url && Strings.isFilled(url) && !Strings.isHttpsUrl(url)) {
+            validationConsumer.accept(createInvalidHttpsUrlException(property, url).getMessage());
+        }
+    }
+
+    @Override
+    public void beforeSave(Property property, Object value) {
+        if (value instanceof String url && Strings.isFilled(url) && !Strings.isHttpsUrl(url)) {
```
sabieber, line 36: **Very optional:** Could reduce duplication by extracting the actual check/validation (the if-condition) into a private method

sabieber (approved):

idlira (approved):

src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
jakobvogel: The variable `url` created in the first condition is used within the block for creating the error message. Hence, in order to extract a method, it would be something like a `attemptToParseInvalidString(value)` returning an `Optional`. Not sure whether that would be easier to understand and maintain.

However, like the other validators, this validator currently only checks Strings, and entirely ignores other data types. Maybe we could also get rid of the instance-of and generally use `.toString()`.

Thoughts?

src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
idlira: Thanks 🙏

src/main/java/sirius/biz/tenants/HttpsUrlValidator.java
sabieber: Ah right. I think its okay for now then

mkeckmkeck (approved):

PR #2083, opened 2025-01-29
SIRI-1047 List Yes, No or Ja, Nein for allowed boolean values in job-docu, also translated enum values

ymo-sci:
### Description

For BooleanProperty this is already done in sirius.biz.importer.format.BooleanPropertyTransformer#customizeField but if you add one via FieldDefinition.booleanField we also want to add the translated value.

For Enum Values we also already import the translated value like in the export, so we simply also add it to the docu.

### Question:
This only adds the current language version in allowed values, maybe the translated values could also be added for all other possible active languages like in the headlines?

Bildschirmfoto 2025-01-29 um 14 01 52
Bildschirmfoto 2025-01-29 um 14 03 21

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1047](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1047)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/importer/format/EnumPropertyTransformer.java
+++ b/src/main/java/sirius/biz/importer/format/EnumPropertyTransformer.java
@@ -11,7 +11,9 @@ package sirius.biz.importer.format;
 import sirius.db.mixing.properties.EnumProperty;
 import sirius.kernel.di.std.Register;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 /**
  * Generates a {@link FieldDefinition} for an {@link EnumProperty}.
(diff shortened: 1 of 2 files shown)


mko-sci (approved):

mkeckmkeck (approved):
I think using the current language only is ok. We have a valid and quite common non-translated fallback with true/false

sabieber:
> Question: This only adds the current language version in allowed values, maybe the translated values could also be added for all other possible active languages like in the headlines?
Would be quite a lot in OXOMI and I think would decrease clarity.

src/main/java/sirius/biz/importer/format/EnumPropertyTransformer.java
```diff
@@ -32,9 +34,13 @@ protected String determineType(EnumProperty property) {
...
-                                                                                     .getType()).getEnumConstants())
-                                                   .map(Enum::name)
-                                                   .toList()));
+        List<String> allowedValues = new ArrayList<>();
+        allowedValues.addAll(Arrays.stream(((Class<Enum<?>>) property.getField().getType()).getEnumConstants())
+                                   .map(Enum::name)
+                                   .toList());
+        allowedValues.addAll(Arrays.stream(((Class<Enum<?>>) property.getField().getType()).getEnumConstants())
+                                   .map(e -> "$" + e.getClass().getSimpleName() + "." + e.name())
+                                   .toList());
```
sabieber, line 43: Why do we have to stream twice? Just forEach instad of map and toList?
Would also remove the necessity to create two temporary lists.

sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/biz/importer/format/EnumPropertyTransformer.java
ymo-sci: Simplified - thx :)

sabieber (approved):

scireum-mbo (approved):

PR #2089, opened 2025-02-05
Corrects BaseImportHandler#isChanged Checks

fhaScireum:
### Description

Corrects the Check for determining if an entity was changed and thus should be updated. This was wrong for both the SQL and MongoDB implementations as they only checked the `AutoImport` properties. As properties outside of that list may have been changed via custom logic, the check must include them, too.

This issue already already manifested for `ImportTransactionData` as its properties shouldn't be auto imported, but had to be marked as such because of the insufficient `isChanged` logic and because ImportTransactionHelper#deleteUnmarked would otherwise delete entities "seen" by the import but had no other changes.

Note: Alternatively to declaring the 4 TraceData mappings directly, we could implement a new annotation called something like "IgnoreInChangeCheck" to allow additional, project specific fields to be ignored in the check. However, I currently see no use case, because if anything changes besides the "changedAt" and related metadata, we generally want to persist the change.

Tested scenarios:
- Import in sync mode relating to ImportTransactionData and ImportTransactionHelper#deleteUnmarked in an SQL project
- Update of backend user accounts via the import job for both SQL and MongoDB systems.

Untested:
- **Import in sync mode in a MongoDB based project.**
- SQLEntityImportHandler#updatePersistentEntity as it is pretty exotic and isn't used anywhere I know of. I see no reason why it should be used but I didn't want to delete it causing a potentially breaking change either.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1057](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1057)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/importer/BaseImportHandler.java
+++ b/src/main/java/sirius/biz/importer/BaseImportHandler.java
@@ -443,7 +443,6 @@ public abstract class BaseImportHandler<E extends BaseEntity<?>> implements Impo
                          .filter(property -> !property.isAnnotationPresent(IgnoreInImportChangedCheck.class))
                          .map(Property::getName)
                          .map(Mapping::named)
-                         .filter(mapping -> !BaseEntity.ID.equals(mapping))
                          .toList();
     }
 


mko-sci (approved):

scireum-mbo (approved):

idlira (changes requested):
From your comment:
> This issue already already manifested for ImportTransactionData as its properties shouldn't be auto imported, but had to be marked as such because of the insufficient isChanged logic and because ImportTransactionHelper#deleteUnmarked would otherwise delete entities "seen" by the import but had no other changes.

Transactional entities should always be *marked*, which will set a new transaction ID and will force them to be updated as the txnId is _watched_ for changes.

Not marking an entity during import might result in unwanted deletions. 

I kinda understand the goal of your change, but I think its a wrong usage.

fhaScireum:
> From your comment:
> 
> > This issue already already manifested for ImportTransactionData as its properties shouldn't be auto imported, but had to be marked as such because of the insufficient isChanged logic and because ImportTransactionHelper#deleteUnmarked would otherwise delete entities "seen" by the import but had no other changes.
> 
> Transactional entities should always be _marked_, which will set a new transaction ID and will force them to be updated as the txnId is _watched_ for changes.
> 
> Not marking an entity during import might result in unwanted deletions.
> 
> I kinda understand the goal of your change, but I think its a wrong usage.

The fields are marked regardless of if they were an AutoImport annotation or not.

The AutoImport annotation was added to the fields specifically to update an entity even if only these fields are changed.

Now, these fields do no longer need to wear the AutoImport annotation as they are checked regardless and the value is set by the TransactionImportHelper anyway -> Thus the AutoImport annotations on these fields are useless.

src/main/java/sirius/biz/importer/BaseImportHandler.java
```diff
@@ -429,6 +431,25 @@ protected List<Mapping> getAutoImportMappings() {
...
+     * Returns all mappings which should be checked for changes to determine if the persisted entity should be
+     * updated.
+     *
+     * @return a list of all mappings which should be checked for changes
+     */
+    protected List<Mapping> getMappingsToCheckForChanges() {
+        return descriptor.getProperties()
+                         .stream()
+                         .map(Property::getName)
+                         .map(Mapping::named)
```
idlira, line 444: **Question:** Does this works fine with composites and nested objects?

idlira (approved):

mkeckmkeck (approved):

sabieber (approved):

mko-sci (approved):

jakobvogel (approved):

src/main/java/sirius/biz/importer/txn/ImportTransactionData.java
```diff
@@ -43,7 +41,6 @@ public class ImportTransactionData extends Composite {
      */
     public static final Mapping SOURCE = Mapping.named("source");
     @NoJournal
-    @AutoImport(hidden = true)
```
ymo-sci, line 46: Wouldn´t they appear now in the job-documentations?

fhaScireum: No, that has no impact on that. See `sirius.biz.protocol.JournalData#fetchJournaledAndChangedProperties`

ymo-sci (approved):

PR #2092, opened 2025-02-07
Permits to wait longer for a variant conversion

idlira:
### Description

The main purpose of this PR is to:
- permit several hard-coded settings such as retry delay and number of attempts to be customized (same values used as before, so the standard config is non-breaking)
- permits to wait longer for a conversion to finish. Previously we would return a HTTP 503 after around 2s for any successful conversion which could take longer than that. Now we support the `/dasd/...?waitLonger=true` parameter to makes the server retry a few more times (obviously customizable). Defaulting to 3x the normal amount.

One use case for this ALREADY IMPLEMENTED is the conversion for embedding in PDF (BlobPdfReplaceHandler).

To achieve this, the BlobStorageSpace has been refactored (thus BREAKING), in order to remove a *god* method which based on a single flag took various complex code paths. I now hope this is more cleaner to understand. By the way, the [layer2's README.md](https://github.com/scireum/sirius-biz/blob/ili/OX-11845/src/main/java/sirius/biz/storage/layer2/README.md) of the storage framework has been enhanced with hopefully useful diagrams to better help understand how the whole thing fits together.

Note that the URLBuilder has been enhanced with a `waitLonger` builder-method in order to create URLs with the extra setting. Also note that this class should be **refactored to use the kernel's URLBuilder**, but this would require even more refactoring, so maybe another time.

BREAKING changes:
- BlobStorageSpace
  - OLD: `Optional<FileHandle> download(String blobKey, String variant);`
  - NEW: `Optional<FileHandle> download(String blobKey, String variant, boolean waitLonger);`

- BasicBlobStorageSpace
  - A protected method called `tryResolvePhysicalKey` has been removed, in case it has been overwritten (*very unlikely*).

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11848](https://scireum.myjetbrains.com/youtrack/issue/OX-11848)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
+++ b/src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
@@ -1080,13 +1080,14 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
      * Note: This method will not be able to return a file if the blob doesn't yet exist in the requested variant and
      * the conversion is disabled on this node (see {@link #conversionEnabled}).
      *
-     * @param blobKey the blob key of the blob to download
-     * @param variant the variant of the blob to download. Use {@link URLBuilder#VARIANT_RAW} to download the blob itself
+     * @param blobKey    the blob key of the blob to download
+     * @param variant    the variant of the blob to download. Use {@link URLBuilder#VARIANT_RAW} to download the blob itself
+     * @param waitLonger whether to wait longer for the conversion to complete
      * @return a handle to the given object wrapped as optional or an empty one if the object doesn't exist or couldn't
      * be converted
      */
     @Override
-    public Optional<FileHandle> download(String blobKey, String variant) {
+    public Optional<FileHandle> download(String blobKey, String variant, boolean waitLonger) {
         touch(blobKey);
 
         try {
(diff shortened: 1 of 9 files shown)


src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -178,7 +178,14 @@ public abstract class BasicBlobStorageSpace<B extends Blob & OptimisticCreate, D
      */
     private static final String EXECUTOR_STORAGE_CONVERSION_DELIVERY = "storage-conversion-delivery";
 
+    /**
+     * Describes if the variant as obtained from cache, lookup or delegated (where a conversion was requested).
```
sabieber, line 188: Typo: `as` -> `was`

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1523,12 +1524,14 @@ protected String lookupPhysicalKey(String blobKey, @Nonnull String variantName)
...
+     * @param waitLonger  whether to wait longer for the conversion to complete
      * @return the variant for the given blob with the given name or null if no such variant exists
      */
     @Nullable
-    protected V findOrCreateVariant(B blob, String variantName) {
+    protected V findOrCreateVariant(B blob, String variantName, boolean waitLonger) {
         try {
-            return attemptToFindOrCreateVariant(blob, variantName, maxOptimisticLockAttempts - 1);
+            int retries = waitLonger ? maxLongConversionAttempts : maxConversionAttempts;
+            return attemptToFindOrCreateVariant(blob, variantName, retries);
```
sabieber: **Question:** Is this change from `maxOptimisticLockAttempts` to `maxConversionAttempts` intended? These seem to be different limits.

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
```diff
@@ -1977,6 +1981,10 @@ private void delegateConversion(String blobKey, String variant, Response respons
         }
     }
 
+    private boolean isMarkedToWaitLonger(Response response) {
+        return Boolean.parseBoolean(response.getHeader(BlobDispatcher.HEADER_WAIT_LONGER));
```
sabieber, line 1986: **Question:** Can Parse-Errors occur here? Maybe wrap in a try-catch with `return false` in case of an exception.

sabieber (commented):
Nice documentation 👌🏻

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
idlira: From what I can see from the code, no. Anything not evaluating to a String `true` returns false, even null.

src/main/java/sirius/biz/storage/layer2/BasicBlobStorageSpace.java
idlira: Yes, it is intended. I can only believe the previous variable was a copy & paste mistake as we have 2 variables for 2 very distinct purposes: retry vs optimistic locking.

jakobvogel (approved):
👏

sabieber (approved):

mkeckmkeck (approved):

PR #2104, opened 2025-03-11
Missing objects check job: use local counters

idlira:
### Description

We are hitting a problem using process counters as these are propagated via redis and with increased number of parallel tasks, they are updated way too often.

We use local counters instead and print the results as process state more gracefully.

Drive-by: preset URL fill and warning state.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-993](https://scireum.myjetbrains.com/youtrack/issue/SIRI-993)

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
+++ b/src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
@@ -130,6 +130,10 @@ public abstract class MissingBlobObjectCheckJob<B extends BaseEntity<I> & Blob, 
                 process.tryUpdateState(buildStatusMessage());
             }
 
+            if (counters.get(COUNTER_MISSING_BLOBS) > 0 || counters.get(COUNTER_MISSING_VARIANTS) > 0) {
+                process.log(ProcessLog.warn().withMessage("Detected missing objects."));
+            }
+
             process.forceUpdateState(buildStatusMessage());
             if (!TaskContext.get().isActive() && firstIdInBlock != null) {
                 process.log(ProcessLog.warn()
--- a/src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
+++ b/src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
@@ -34,6 +34,7 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
```diff
@@ -280,6 +282,17 @@ protected void collectAcceptedFileExtensions(Consumer<String> fileExtensionConsu
...
+        @Override
+        protected void computePresetFor(@Nonnull QueryString queryString,
+                                        @Nullable Object targetObject,
+                                        Map<String, Object> preset) {
+            fillPresetFromUrl(STORAGE_SPACE_PARAMETER, queryString, preset);
+            fillPresetFromUrl(INCLUDE_REPLICATION_SPACE_PARAMETER, queryString, preset);
+            fillPresetFromUrl(PARALLEL_TASKS_PARAMETER, queryString, preset);
+            fillPresetFromUrl(START_FROM_ID_PARAMETER, queryString, preset);
+            fillPresetFromUrl(CHECK_VARIANTS_PARAMETER, queryString, preset);
+        }
```
fhaScireum, line 294: Is this needed? Job parameters can already be prefilled if the query string in the URL contains the parameter name.

For instance, opening `/job/import-sql-user-accounts?ignoreEmpty=false` (on SQL systems) sets the "ignore empty values" parameter to false.

idlira: Thx... reverted

src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
```diff
@@ -173,9 +183,7 @@ private synchronized void writeLine(Object... columns) throws IOException {
     }
 
     private synchronized void incrementCounter(String counter) {
-        if (TaskContext.get().isActive()) {
-            process.incCounter(counter);
-        }
+        counters.put(counter, counters.get(counter) + 1);
```
jakobvogel, line 192: Would it make sense to extend the jobs framework to allow for manually displaying counter values in the side bar?

jakobvogel (approved):

src/main/java/sirius/biz/storage/util/MissingBlobObjectCheckJob.java
idlira: Not a bad idea. I will write down in my list to see how easily that can be achieved (just not for this job though)

mkeckmkeck (approved):

PR #2135, opened 2025-06-11
Make the importer in the EntityExportJob non final, so that it can be overwritten.

oolscireum:
### Description

Makes the importer in the EntityExportJob non final, so that it can be overwritten.

### Additional Notes

- This PR fixes or works on following ticket(s): [DH-310](https://scireum.myjetbrains.com/youtrack/issue/DH-310)

### Checklist


idlira (approved):
Question: are you not getting any SonarLint warning to make the var back to final since is only set in the constructor? Shouldn't you provide a getter/setter on it?

sabieber (approved):

scireum-mbo (approved):

fhaScireum (approved):

mkeckmkeck (approved):

oolscireum:
> Question: are you not getting any SonarLint warning to make the var back to final since is only set in the constructor? Shouldn't you provide a getter/setter on it?

Currently i get no warning for that.

PR #2142, opened 2025-06-27
Utility Methods for Process Links

jakobvogel:
### Description

Provides methods to clear the external links attached to a process, and to only add a link if it is not yet present. Also fixes a potential NPE.

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/process/ProcessLink.java
+++ b/src/main/java/sirius/biz/process/ProcessLink.java
@@ -9,6 +9,7 @@
 package sirius.biz.process;
 
 import sirius.db.mixing.Nested;
+import sirius.kernel.commons.Strings;
 import sirius.kernel.nls.NLS;
 
 /**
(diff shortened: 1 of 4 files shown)


sabieber (approved):

fhaScireum (approved):

src/main/java/sirius/biz/process/ProcessLink.java
```diff
@@ -49,4 +50,20 @@ public String getLabel() {
     public String getUri() {
         return uri;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (object instanceof ProcessLink otherLink) {
```
idlira, line 56: **Optional**: I would prefer a comparison only over the label, so one can update a link if needed (unless we introduce a `removeLink`).

src/main/java/sirius/biz/process/Processes.java
```diff
@@ -693,6 +693,41 @@ public boolean addLink(String processId, ProcessLink link) {
...
+     * can contribute to the process.
+     *
+     * @param processId the process to update
+     * @param link      the link to add
+     * @return <tt>true</tt> if the process was successfully modified or if the link was present before, <tt>false</tt>
+     * otherwise
+     */
+    public boolean addUniqueLink(String processId, ProcessLink link) {
+        return modify(processId, process -> true, process -> {
+            if (!process.getLinks().contains(link)) {
```
idlira, line 711: **Optional**: should that replace an existing link if based on label (see other comment)

PR #2283, opened 2026-05-27
Enforce rate-limit for backend-login

ymo-sci:
### Description

it seems, that there was rate-limiting years ago, but got removed in https://github.com/scireum/sirius-biz/commit/613257814#diff-11ffed5b48f839e6d599427f3f00a4510ef156fd08267b2fb8b92c1ac138cebeL48-L49 the security-block is now adjusted to only allow 10 wrong logins in 2 minutes, then the rate-limit will stop the user earlier for trying more users/passwords

Bildschirmfoto 2026-05-27 um 16 04 51

- after the 4th bad login a Captcha is included now in the login-form which should be ok for real users with typos

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1217](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1217)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/test/resources/test.conf
+++ b/src/test/resources/test.conf
@@ -102,6 +102,12 @@ isenguard.limit.test {
     limit = 5
 }
 
+isenguard.limit.security {
+    interval = 1m
+    limit = 3
+    type = "ip"
+}
+
 async {
     distributed {
         queues {
--- a/src/main/java/sirius/biz/tenants/TenantUserManager.java
+++ b/src/main/java/sirius/biz/tenants/TenantUserManager.java
@@ -12,6 +12,8 @@ import com.typesafe.config.Config;
 import sirius.biz.analytics.events.EventRecorder;
 import sirius.biz.analytics.events.UserActivityEvent;
 import sirius.biz.analytics.flags.PerformanceFlag;
+import sirius.biz.isenguard.Isenguard;
+import sirius.biz.isenguard.RateLimitingInfo;
 import sirius.biz.model.LoginData;
 import sirius.biz.protocol.AuditLog;
 import sirius.biz.web.SpyUser;
@@ -27,6 +29,7 @@ import sirius.kernel.di.PartCollection;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Parts;
 import sirius.kernel.health.Exceptions;
+import sirius.kernel.health.HandledException;
 import sirius.kernel.health.Log;
 import sirius.kernel.nls.NLS;
 import sirius.kernel.settings.Extension;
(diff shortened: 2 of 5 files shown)


sabieber (approved):
> BREAKING: to use it safe, http.captcha.secret = "secret" must be filled per system - a default is provided.

We got security feedback for other unsafe defaults before so should we really provide an unsafe default here or rather produce an error when the secret is not set when trying to be used?

src/test/resources/test.conf
```diff
@@ -102,6 +102,12 @@ isenguard.limit.test {
     limit = 5
 }
 
+isenguard.limit.security {
+    interval = 1m
+    limit = 3
+    type = "ip"
```
idlira, line 108: Might be a bit edge case but several users in a company will land on our server under the same IP. So, 4 attempts **per user** is not really guaranteed unless we also include the user name in the scope.

ymo-sci: this is the test.conf - for prod a limit of 10 / 2m would be used and only for bad logins - think that should never be the case...

idlira: It's not about being enough or not. It's about the information being displayed to the user being wrong. As soon as 2 people make a mistake, one is stealing the limit from the other. And all you have to do is to setup the limit having the user name with the source IP concatenated.

sabieber: Mhhh I thought about the concatenation idea a bit and I think this could mean an attacker could optimize for it if he finds that out. As in: he could simply iterate over the accounts he wants to breach and make an attempt for each, then rinse and repeat.

idlira: Well, we can leave as is. The goal was to protect for non intentional password change attempts. Users in regular use shall not even feel the difference. But the captcha logic would also trigger a bit sooner in my previous example.

ymo-sci: > As soon as 2 people make a mistake, one is stealing the limit from the other.

Yea ok, but only in the configured 2 minutes... i think bad luck for this edge case

idlira: > he could simply iterate over the accounts he wants to breach and make an attempt for each, then rinse and repeat.

well, than we would need 2 limits: one for each user name (rate limit + CAPTCHA) and additionally one for IP trying to cover the attacker. Again, we can think about it in a later step and go for it with this version.

bkositza (approved):

idlira (approved):

jakobvogel (approved):

fhaScireum (approved):

mkeckmkeck (approved):

ymo-sci:
> > BREAKING: to use it safe, http.captcha.secret = "secret" must be filled per system - a default is provided.
> 
> We got security feedback for other unsafe defaults before so should we really provide an unsafe default here or rather produce an error when the secret is not set when trying to be used?

https://scireum.myjetbrains.com/youtrack/issue/SIRI-1225

PR #2306, opened 2026-06-09
Add reusable sorting support for Tycho tables

mku-sci:
### Description

This PR adds reusable sorting support for Tycho tables.

- Added `TableSortOption` to define request key, translation key, and Mongo mapping.
- Added `TableSorting` to centralize sorting parameters, current UI state, and optional per-page sorting preferences.
- Added `SortOrder` for validated sort direction handling.
- Extended `BasePageHelper` with sortable fields, configurable default sorting, sorting before pagination, and optional per-page sorting persistence.
- Extended `MongoPageHelper` with:
  - `addSortableField(...)`
  - `addSortableOption(...)`
  - `addSortableOptions(...)`
- compact dropdown for selecting sort field and direction, including reset support.

### Example
 
Bildschirmfoto 2026-06-12 um 18 22 19

## Behavior

| Request state | Result |
|---|---|
| No sorting configuration | Existing behavior remains unchanged |
| Missing `sort` parameter | Configured default sorting is applied |
| Missing `order` parameter | Configured default sorting is applied |
| Unknown `sort` value | Configured default sorting is applied |
| Unknown `order` value | Configured default sorting is applied |
| Valid sort field with `asc` | Ascending sorting is applied |
| Valid sort field with `desc` | Descending sorting is applied |

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1009](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1009)

### Checklist


src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -431,9 +469,41 @@ public Q buildUnderlyingQueryWithFacets() {
...
+     * Applies the requested sorting to the current query.
+     * <p>
+     * The method is executed at most once per helper lifecycle. If a valid sort field and
+     * sort order are present in the request, the corresponding registered sort function is
+     * applied. Otherwise, the configured default sorting is used if available.
+     */
+    private void applySorting() {
+        if (sortingApplied || !hasSortingConfiguration()) {
+            return;
+        }
```
ymo-sci, line 512: how is the handling of indices made here?
With many entries in a table and a sorting after a not indexed field could be system-crashing or timeout

sabieber: You enable it only for rows with an index?

src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -121,6 +130,36 @@ public B withTotalCount() {
...
+     * Enables storing and reusing the selected sort order for this page using the given user preference key.
+     * <p>
+     * This is an explicit opt-in and only affects the registered sortable field mechanism based on
+     * {@link TableSorting#PARAM_SORT} and {@link TableSorting#PARAM_ORDER}.
+     *
+     * @param sortingPreferenceKey the page-specific user preference key to use
+     * @return the helper itself for fluent method calls
+     */
+    @SuppressWarnings("unchecked")
+    public B withSortingPreferenceKey(String sortingPreferenceKey) {
```
sabieber, line 159: **Optional:** Maybe some Kind of `withPageSortingHelper` would be nice to encapsule the configuration of the sorting and also the logic a bit more?

But feel free to wait for other opinions on this

src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -347,6 +386,20 @@ public final B addSortFacet(Tuple<String, Consumer<Q>>... sortOptions) {
...
+     * Registers a sortable field by associating a key with a sort function.
+     *
+     * @param key the unique identifier for the sortable field
+     * @param sortFunction a BiConsumer that defines how sorting should be applied
+     *                     using the given key and sort order
+     * @return the current instance for method chaining
+     */
+    @SuppressWarnings("unchecked")
+    protected final B registerSortableField(String key, BiConsumer<Q, SortOrder> sortFunction) {
+        sortableFields.put(Objects.requireNonNull(key, "key"), Objects.requireNonNull(sortFunction, "sortFunction"));
```
sabieber: **Optional:** Usually we check stuff like this via `if (Strings.isEmpty) { ... return }` as `""` may also break the logic.

src/main/java/sirius/biz/web/BasePageHelper.java
```diff
@@ -431,9 +485,66 @@ public Q buildUnderlyingQueryWithFacets() {
...
     }
 
+    /**
+     * Applies the requested sorting to the current query.
+     * <p>
+     * The method is executed at most once per helper lifecycle. If a valid sort field and
+     * sort order are present in the request, the corresponding registered sort function is
+     * applied. Otherwise, the configured default sorting is used if available.
+     */
+    private void applySorting() {
```
sabieber, line 509: **Optional:** This could then live in the suggested `PageSortHelper`

src/main/java/sirius/biz/web/MongoPageHelper.java
```diff
@@ -181,6 +181,47 @@ public MongoPageHelper<E> addBooleanAggregation(Mapping field) {
...
+    }
+
+    /**
+     * Adds multiple sortable options based on reusable technical option definitions.
+     *
+     * @param options the option definitions to register
+     * @return the helper itself for fluent method calls
+     */
+    public MongoPageHelper<E> addSortableOptions(Iterable<TableSortOption> options) {
+        java.util.Objects.requireNonNull(options, "options").forEach(this::addSortableOption);
```
sabieber: **Question:** How do other methods of the `PageHelper` handle misconfiguration? In a similar way or do they create `HandledExceptions` via the `Exceptions` helper?

src/main/java/sirius/biz/web/SortOrder.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Arrays;
+
+/**
+ * Represents the supported directions for table sorting.
+ */
+public enum SortOrder {
+    ASC(TableSorting.ORDER_ASC),
+    DESC(TableSorting.ORDER_DESC);
+
+    private final String parameterValue;
```
sabieber, line 29: Please document this enum field, so callers know what they can use it for

src/main/java/sirius/biz/web/TableSorting.java
```diff
@@ -0,0 +1,238 @@
...
+ */
+@Register(classes = TableSorting.class)
+public class TableSorting {
+
+    public static final String PARAM_SORT = "sort";
+    public static final String PARAM_ORDER = "order";
+    public static final String PARAM_CLEAR_SORT = "clear-sort";
+
+    public static final String ORDER_ASC = "asc";
+    public static final String ORDER_DESC = "desc";
```
sabieber: **Optional:** I would inverse this, as in define the raw String in the enum values and reference them here like `SortOrder.ASC.getParameterValue()`. If thats possible when binding to a static final field, not sure about it.

src/main/java/sirius/biz/web/TableSortOption.java
```diff
@@ -0,0 +1,49 @@
...
+package sirius.biz.web;
+
+import sirius.db.mixing.Mapping;
+
+import java.util.Objects;
+
+/**
+ * Describes a whitelisted field which can be used for table sorting and displayed as an option in a sorting control.
+ */
+public class TableSortOption {
```
sabieber: **Optional:** Could be a `record`

src/main/resources/default/taglib/t/tableSortingButton.html.pasta
```diff
@@ -0,0 +1,233 @@
+<i:arg name="baseUrl" type="String"/>
```
sabieber, line 1: **Optional:** Would include the represented UI element in the tag name. As in: `tableSortingButton`

src/main/resources/default/taglib/t/tableSorting.html.pasta
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        function handleSelectionChange() {
+            const _selection = syncUi();
+            navigateIfComplete(_selection.sortKey, _selection.orderValue);
+        }
+
+        _toggleButton.addEventListener("click", function () {
+            setMenuOpen(_menu.hidden);
+        });
```
sabieber: **Optional:** We can use arrow notation in the backend, like:

`_toggleButton.addEventListener("click", () => setMenuOpen(_menu.hidden));`

src/main/java/sirius/biz/web/TableSortOption.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final String key;
+    private final String labelKey;
```
sabieber: **Optional:** A `label` that gets resolved via `NLS.smartGet` like we use for job parameters would be more flexible. Then we could provide a static label "Custom label" or translated label "$Model.name".

src/main/java/sirius/biz/web/TableSortOption.java
```diff
@@ -0,0 +1,49 @@
...
+import sirius.db.mixing.Mapping;
+
+import java.util.Objects;
+
+/**
+ * Describes a whitelisted field which can be used for table sorting and displayed as an option in a sorting control.
+ */
+public class TableSortOption {
+
+    private final String key;
```
sabieber: **Question:** Do we really need a key, or could we transpose it from the provided mapping? (Maybe as a second constructor?)

src/main/resources/default/taglib/t/tableSortingButton.html.pasta
```diff
@@ -0,0 +1,233 @@
...
+<i:arg name="sortOptions" type="java.util.List"/>
```
sabieber: **Optional:** I would really like it if this would take in a `Page` (and the `Page` includes the necessary info) ... but don't know if thats possible without much rework.

Benefit:
When integrating we would not need to pass in another parameter into the template that is using the tag but simply including `t:tableSortingButton` and setting up the page helper in Java would be enough.

sabieber (approved):

src/main/java/sirius/biz/web/TableSortOption.java
mku-sci: Good point. Since the mapping is the source of truth, we don't need a separate key anymore. I updated TableSortOption to use mapping.getName() directly as the sort key and removed the additional key from the record.

src/main/java/sirius/biz/web/MongoPageHelper.java
mku-sci: No, they generally use regular exceptions for misconfiguration, not HandledExceptions.

BasePageHelper#getParameterValue as an Example throws an IllegalStateException if no parameter is provided or web context is attached.

sabieber (approved):
Nice!

idlira (approved):

src/main/java/sirius/biz/web/TableSorting.java
```diff
@@ -70,5 +119,120 @@ public String fetchCurrentOrderValue(WebContext webContext) {
...
+    public static String readStoredOrderValue(String userPreferencesKey) {
+        if (!Strings.isFilled(userPreferencesKey)) {
+            return "";
+        }
+
+        return normalizeOrderValue(readStoredPreference(orderPreferenceKey(userPreferencesKey)));
+    }
+
+    public static void storeSortingPreference(String userPreferencesKey, String sortKey, SortOrder sortOrder) {
+        if (!Strings.isFilled(userPreferencesKey) || !Strings.isFilled(sortKey) || sortOrder == null) {
```
fhaScireum: Mimimi: Ideally use `Strings.isEmpty(...)` instead of `!Strings.isFilled(...)`. Same goes for all other places.

fhaScireum (approved):

sabieber (approved):

src/main/java/sirius/biz/web/MongoPageHelper.java
```diff
@@ -181,6 +181,47 @@ public MongoPageHelper<E> addBooleanAggregation(Mapping field) {
...
+     */
+    public MongoPageHelper<E> addSortableField(String key, Mapping mapping) {
+        return registerSortableField(key, (query, sortOrder) -> {
+            if (sortOrder == SortOrder.ASC) {
+                query.orderAsc(mapping);
+            } else {
+                query.orderDesc(mapping);
+            }
+        });
+    }
```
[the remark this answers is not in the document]
sabieber: Some basic tests would indeed be nice

src/test/java/sirius/biz/web/sorting/MongoPageHelperSortingTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class MongoPageHelperSortingTest {
```
sabieber: I need to finally add an AGENTS.md to all repos to prefer Kotlin unit tests :)

sabieber (approved):

sabieber (approved):

PR #2355, opened 2026-07-14
SIRI-1247 Remove Captcha again as it breaks API Routes where captcha submission is not possible

ymo-sci:
### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1247](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1247)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/tenants/TenantUserManager.java
+++ b/src/main/java/sirius/biz/tenants/TenantUserManager.java
@@ -19,6 +19,7 @@ import sirius.biz.protocol.AuditLog;
 import sirius.biz.web.SpyUser;
 import sirius.db.mixing.BaseEntity;
 import sirius.db.mixing.Mixing;
+import sirius.kernel.Sirius;
 import sirius.kernel.cache.Cache;
 import sirius.kernel.cache.CacheManager;
 import sirius.kernel.commons.Explain;
@@ -165,7 +166,7 @@ public abstract class TenantUserManager<I extends Serializable, T extends BaseEn
     /**
      * Contains the number of failed password login attempts after which the next attempt requires a CAPTCHA.
      */
-    private static final int FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA = 4;
+    private static final int DEFAULT_FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA = 4;
 
     /**
      * Contains the prefix added to the account number which is added as artificial role.
@@ -710,7 +711,9 @@ public abstract class TenantUserManager<I extends Serializable, T extends BaseEn
         String ip = webContext.getRemoteIP().getHostAddress();
         return isenguard.checkRateLimitReached(ip,
                                                SECURITY_RATE_LIMIT_REALM,
-                                               FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA);
+                                               Sirius.getSettings()
+                                                     .get("security.passwordLoginCaptcha.failedAttemptsBeforeCaptcha")
+                                                     .asInt(DEFAULT_FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA));
     }
 
     private void verifyPasswordLoginCaptchaIfRequired(@Nullable WebContext webContext) {
--- a/src/main/java/sirius/biz/tenants/TenantUserManager.java
+++ b/src/main/java/sirius/biz/tenants/TenantUserManager.java
@@ -19,7 +19,6 @@ import sirius.biz.protocol.AuditLog;
 import sirius.biz.web.SpyUser;
 import sirius.db.mixing.BaseEntity;
 import sirius.db.mixing.Mixing;
-import sirius.kernel.Sirius;
 import sirius.kernel.cache.Cache;
 import sirius.kernel.cache.CacheManager;
 import sirius.kernel.commons.Explain;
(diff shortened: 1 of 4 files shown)


fhaScireum (approved):

src/test/kotlin/sirius/biz/tenants/TenantUserManagerTest.kt
```diff
@@ -12,29 +12,52 @@ import org.junit.jupiter.api.BeforeAll
...
+        }
+    }
+
     @Test
     fun `Only failed password logins count towards rate limit`() {
         TenantsHelper.getTestUser()
 
         val userManager = UserContext.get().userManager as SQLTenantUserManager
-        val webContext = WebContext.getCurrent()
+        val webContext = TestRequest.GET("/").withRemoteIp(InetAddress.getByName("127.0.0.41"))
```
hso-scireum: Minor input but I think a comment explaining why the explicit IP address is used here might be useful

src/test/kotlin/sirius/biz/tenants/TenantUserManagerTest.kt
```diff
@@ -45,15 +68,31 @@ class TenantUserManagerTest {
...
 
-    companion object {
-        @Part
-        @JvmStatic
-        private lateinit var oma: OMA
+    @Test
+    fun `Password login captcha is only required for login form requests`() {
+        val failedAttemptsBeforeCaptcha =
+            Sirius.getSettings().get("security.passwordLoginCaptcha.failedAttemptsBeforeCaptcha").asInt(4)
+        val remoteIp = InetAddress.getByName("127.0.0.42")
```
hso-scireum: Same for this instance (explanatory comment)

src/test/kotlin/sirius/biz/tenants/TenantUserManagerTest.kt
```diff
@@ -45,15 +68,31 @@ class TenantUserManagerTest {
...
-        fun setup() {
-            oma.readyFuture.await(Duration.ofSeconds(60))
+        repeat(failedAttemptsBeforeCaptcha) {
+            isenguard.registerCallAndCheckRateLimitReached(
+                loginViaForm.getRemoteIP().hostAddress,
+                TenantUserManager.SECURITY_RATE_LIMIT_REALM,
+                failedAttemptsBeforeCaptcha + 1
+            ) { RateLimitingInfo.fromWebContext(loginViaForm, null) }
+        }
+        assertFalse(TenantUserManager.isPasswordLoginCaptchaRequired(loginViaParameter))
```
hso-scireum: Minor addition: If multiple assertions are made in a test case, using `assertAll` and passing multiple test lambdas has one crucial benefit: If one assertion fails, the other assertions are still evaluated. This is helpful in case multiple assertions fail and the test would need to be executed multiple times in order to detect all failures

hso-scireum (commented):
I have some minor comments but the PR looks very good overall :)

mkeckmkeck (commented):
This makes the CAPTCHA implementiation quite useless as attackers can bypass it easily, doesn't it? 
But real user get annoying CAPTCHAs. Can we work this out somehow smarter?

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -710,7 +711,9 @@ public static boolean isPasswordLoginCaptchaRequired(@Nullable WebContext webCon
         String ip = webContext.getRemoteIP().getHostAddress();
         return isenguard.checkRateLimitReached(ip,
                                                SECURITY_RATE_LIMIT_REALM,
-                                               FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA);
+                                               Sirius.getSettings()
+                                                     .get("security.passwordLoginCaptcha.failedAttemptsBeforeCaptcha")
+                                                     .asInt(DEFAULT_FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA));
```
sabieber: **Optional:** Wouldn't it be easier to simply define a separate configurable realm for the captcha? Or is the connection to the "security" realm necessary?

Also reading the config on every call could be expensive

sabieber (approved):

jakobvogel (approved):

src/main/java/sirius/biz/tenants/TenantUserManager.java
```diff
@@ -165,7 +166,7 @@ public abstract class TenantUserManager<I extends Serializable, T extends BaseEn
     /**
      * Contains the number of failed password login attempts after which the next attempt requires a CAPTCHA.
      */
-    private static final int FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA = 4;
+    private static final int DEFAULT_FAILED_PASSWORD_LOGIN_ATTEMPTS_BEFORE_CAPTCHA = 4;
```
idlira: **Optional:** I would drop this default constant and load it via `@ConfigValue`, as you already sets a default in the configuration.

idlira (approved):

src/main/java/sirius/biz/tenants/TenantUserManager.java
sabieber: And I would not define a new separate Config but rather a separate Isenguard-Realm

fhaScireum (approved):

jakobvogel (approved):

sabieber (approved):

PR #2406, opened 2026-08-17
Add MultiSelectEntityParameter for entity multi-selects with autocomplete

ymo-sci:
### Description

Job parameters so far only supported multi-selects over a fixed list of string options (MultiSelectStringParameter) or single-entity selection with autocomplete (EntityParameter). This adds a base class for selecting multiple entities via autocomplete routes, needed e.g. to restrict export jobs to a set of linked entities.

New:
* MultiSelectEntityParameter: multi-select parameter for BaseEntity values. Uses an Autocompleter or a custom autocomplete URI for suggestions, serializes the selection as "|"-delimited entity ids and renders only the currently selected entities as preselected options (suggestions are provided at runtime via the autocomplete). Falls back to the plural label of the entity descriptor if no label is given.
* EntityParameterUtil: extracted the mechanics shared between EntityParameter and MultiSelectEntityParameter (autocompleter resolution, descriptor/mapper access, entity lookup incl. the "mustExist" error and the tenant access check). Both parameter types cannot share a base class due to incompatible value types (single entity vs. list), therefore the shared logic lives in this util which both delegate to. Parameter callbacks are passed as method references so that overrides in subclasses are picked up. The public/protected API of EntityParameter remains unchanged.

Changed:
* MultiSelectParameter now owns the complete "|"-encoding: it implements checkAndTransformValue (accepting both form value lists and already encoded strings) and resolveFromString, subclasses only implement checkAndTransformSingleValue / resolveSingleValueFromString for a single value. The contract is documented: return null to silently skip meaningless values, throw a HandledException whenever dropping a value would silently change the meaning of the job.
* MultiSelectParameter#computeValueUpdate now emits a [{value, text}] array for all multi-selects as expected by TokenAutocomplete#val - previously a plain string array would have silently cleared the field. The option name/label are derived via the new createValueName / createValueLabel hooks.
* The templates selectMultiString.html.pasta and the would-be entity variant are unified into selectMulti.html.pasta which passes the new MultiSelectParameter#getSuggestionUri (empty string for fixed option lists, autocomplete URL for entity parameters).

Fixed:
* assertTenantAccess no longer NPEs when the tenants framework is not part of the product.
* A configured but unregistered Autocompleter is now reported via an incident instead of silently degrading into a select without suggestions.
* MultiSelectStringParameter#getValues no longer fetches the entries map and splits the context value once per entry.

### Additional Notes
* Resolving a stored value now returns an empty optional instead of an empty list when no value can be resolved anymore, so require() and presence checks detect selections which became invalid after submission. Values which can no longer be resolved (e.g. deleted entities) are still skipped silently on resolve as this path is also used when rendering the job form; strict validation happens on every job submit/start via checkAndTransform.
* A single already-encoded string ("a|b", e.g. from a preset) is now split and validated per token by checkAndTransformValue instead of being rejected as a whole, consistent with resolveFromString.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

- This PR fixes or works on following ticket(s): [SE-15248](https://scireum.myjetbrains.com/youtrack/issue/SE-15248)
- This PR is related to PR:  https://github.com/scireum/sellsite/pull/14096

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/jobs/params/EntityParameterProxy.java
+++ b/src/main/java/sirius/biz/jobs/params/EntityParameterProxy.java
@@ -37,7 +37,7 @@ import java.util.function.Supplier;
  *
  * @param <V> the type of entities handled by the parameter
  */
-public class EntityParameterUtil<V extends BaseEntity<?>> {
+public class EntityParameterProxy<V extends BaseEntity<?>> {
 
     @Part
     private static Mixing mixing;
@@ -83,10 +83,10 @@ public class EntityParameterUtil<V extends BaseEntity<?>> {
      * @param customAutocompleteUriSupplier provides the custom autocomplete URI to use (may supply <tt>null</tt>)
      * @param labelSupplier                 provides the label of the parameter (used in error messages)
      */
-    public EntityParameterUtil(Supplier<Class<V>> typeSupplier,
-                               Supplier<Class<? extends Autocompleter<V>>> autocompleterSupplier,
-                               Supplier<String> customAutocompleteUriSupplier,
-                               Supplier<String> labelSupplier) {
+    public EntityParameterProxy(Supplier<Class<V>> typeSupplier,
+                                Supplier<Class<? extends Autocompleter<V>>> autocompleterSupplier,
+                                Supplier<String> customAutocompleteUriSupplier,
+                                Supplier<String> labelSupplier) {
         this.typeSupplier = typeSupplier;
         this.autocompleterSupplier = autocompleterSupplier;
         this.customAutocompleteUriSupplier = customAutocompleteUriSupplier;
@@ -115,7 +115,7 @@ public class EntityParameterUtil<V extends BaseEntity<?>> {
      * {@link Autocompleter} nor a custom autocomplete URI is available
      */
     @Nullable
-    public String getAutocompleteUrl() {
+    public String determineAutocompleteUrl() {
         if (autocompleterSupplier.get() == null) {
             return customAutocompleteUriSupplier.get();
         }
--- a/src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
+++ b/src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
@@ -49,11 +49,30 @@ public class EntityParameterUtil<V extends BaseEntity<?>> {
     @Part
     private static GlobalContext globalContext;
 
+    /**
+     * Provides the type of entities handled by the parameter.
+     */
     private final Supplier<Class<V>> typeSupplier;
+
+    /**
+     * Provides the {@link Autocompleter} type used to compute suggestions (may supply <tt>null</tt>).
+     */
     private final Supplier<Class<? extends Autocompleter<V>>> autocompleterSupplier;
+
+    /**
+     * Provides the custom autocomplete URI used if no {@link Autocompleter} is configured (may supply
+     * <tt>null</tt>).
+     */
     private final Supplier<String> customAutocompleteUriSupplier;
+
+    /**
+     * Provides the label of the parameter, used in error messages.
+     */
     private final Supplier<String> labelSupplier;
 
+    /**
+     * Caches the {@link EntityDescriptor} of the entity type as determined by {@link #getDescriptor()}.
+     */
     private EntityDescriptor descriptor;
 
     /**
--- a/src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
+++ b/src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
@@ -116,21 +116,31 @@ public class EntityParameterUtil<V extends BaseEntity<?>> {
      */
     @Nullable
     public String getAutocompleteUrl() {
-        Class<? extends Autocompleter<V>> autocompleterType = autocompleterSupplier.get();
-        if (autocompleterType != null) {
-            Optional<? extends Autocompleter<V>> autocompleter = findAutocompleter();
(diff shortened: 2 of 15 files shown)


src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
```diff
@@ -0,0 +1,196 @@
...
+ * entity vs. a list of entities) and therefore cannot share a common base class. Hence, the shared logic
+ * (autocompleter resolution, descriptor and mapper access, entity lookups and tenant checks) is provided via
+ * this support class which both parameter types delegate to.
+ * <p>
+ * The parameter callbacks are passed in as {@link Supplier suppliers} (i.e. method references onto the
+ * parameter itself) so that overrides in parameter subclasses are properly picked up.
+ *
+ * @param <V> the type of entities handled by the parameter
+ */
+public class EntityParameterUtil<V extends BaseEntity<?>> {
```
sabieber: **Optional:** Don't really like that class name. Maybe something like `EntityParameterBehaviour` or something equivalent would be better.
For a Util I would expect either a registered Part or a constant class, not something that gets instantiated.

src/main/java/sirius/biz/jobs/params/EntityParameterProxy.java
```diff
@@ -0,0 +1,196 @@
...
+    @Nullable
+    private static Tenants<?, ?, ?> tenants;
+
+    @Part
+    private static GlobalContext globalContext;
+
+    private final Supplier<Class<V>> typeSupplier;
+    private final Supplier<Class<? extends Autocompleter<V>>> autocompleterSupplier;
+    private final Supplier<String> customAutocompleteUriSupplier;
+    private final Supplier<String> labelSupplier;
```
sabieber, line 71: **Optional:** Could we document that field so we don't need to jump to the constructor documentation when trying to understand what these fields mean in the logic below?

src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Returns the autocompletion URL used to determine suggestions for inputs provided by the user.
+     *
+     * @return the autocomplete URL used to provide suggestions for user input or <tt>null</tt> if neither an
+     * {@link Autocompleter} nor a custom autocomplete URI is available
+     */
+    @Nullable
+    public String getAutocompleteUrl() {
```
sabieber: **Optional:** A lot of logic for a "get..." function

src/main/java/sirius/biz/jobs/params/MultiSelectStringParameter.java
```diff
@@ -11,23 +9,15 @@
...
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
- * Provides a multi select parameter from a list of key-value pairs.
+ * Provides a multi select parameter from a fixed list of key-value pairs.
  */
 public class MultiSelectStringParameter extends MultiSelectParameter<String, MultiSelectStringParameter> {
```
sabieber, line 19: **Question:** Parameters are untestable? Would really prefer some tests that ensure unchanged (except for the bugfixes) behaviour of the pre-existing parameter types with those deep changes.

src/main/java/sirius/biz/jobs/params/MultiSelectParameter.java
```diff
@@ -33,11 +57,105 @@ protected MultiSelectParameter(String name, String label) {
...
+        return "/templates/biz/jobs/params/selectMulti.html.pasta";
     }
 
+    /**
+     * Returns the serialized name of the given value, used as the <tt>value</tt> of its option in the UI.
+     *
+     * @param value the value to derive the name from
+     * @return the serialized name of the given value
+     */
+    protected abstract String createValueName(V value);
```
fhaScireum, line 89: Breaking Change: DH has a MultiDataSetsParameter class, that must adapt to the changes.

src/main/java/sirius/biz/jobs/params/EntityParameterUtil.java
```diff
@@ -0,0 +1,225 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
fhaScireum:
```suggestion
 * https://www.scireum.de - info@scireum.de
```

src/main/java/sirius/biz/jobs/params/MultiSelectEntityParameter.java
```diff
@@ -0,0 +1,152 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
fhaScireum:
```suggestion
 * https://www.scireum.de - info@scireum.de
```

src/main/java/sirius/biz/jobs/params/MultiSelectParameter.java
ymo-sci: --> DH PR will follow, thx for finding!

sabieber (approved):

fhaScireum (approved):

idlira (approved):

PR #2416, opened 2026-08-21
Record all decimal places of amounts in the change journal

ymo-sci:
### Description

Previously, changed values were journaled via NLS.toUserString, which formats decimals using the two-digit pattern NLS.patternDecimal. As the journal string is persisted, a change in the third or a later decimal place was recorded as two identical values (e.g. "1,23 -> 1,23").

Filled amounts are now formatted by the new JournalData.formatValue using the localized decimal format, with maximumFractionDigits raised to the actual scale of the value (stripTrailingZeros), so all significant decimal places are recorded. DelegateJournalData uses the same helper for its change and delete journals. Amount is the only decimal property type in sirius-db, so no other types need special handling; the BigDecimal is formatted directly, avoiding the doubleValue() precision loss of NLS.toUserString.

Edge cases / intentional behavior:
- Localized formatting is kept (decimal comma, grouping separators of the default language) and at least two decimal places are always shown, so typical entries ("1.500,00 -> 1.600,00") look as before.
- Trailing zeros stemming from the @Numeric scale of the column are stripped; only significant decimal places are shown.
- Values may now legitimately show up to five decimal places even for "money-like" fields: user input parsed via Amount.of is normalized to Amount.SCALE (5) before the @Numeric rounding applies. The journal simply reveals what is actually persisted.
- Fields written via Amount.ofRounded (e.g. import or manual controller paths) keep their full scale and may show more decimal places (up to the @Numeric scale, e.g. 15).

Empty amounts are still recorded as an empty string. Existing journal entries remain rounded, as the strings are persisted at write time.

- fixes: SIRI-1265

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1265](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1265)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/protocol/DelegateJournalData.java
+++ b/src/main/java/sirius/biz/protocol/DelegateJournalData.java
@@ -22,7 +22,6 @@ import sirius.kernel.commons.Monoflop;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.health.Exceptions;
-import sirius.kernel.nls.NLS;
 
 import javax.annotation.Nonnull;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


sabieber (approved):
Nice!

idlira (approved):

jakobvogel (approved):

fhaScireum (approved):

mku-sci (approved):

jmuscireum (approved):

PR #2424, opened 2026-08-28
Adds change tracing and journaling to scheduler entries

ymo-sci:
### Description

BREAKING: SchedulerEntry now extends Traced and Journaled. Custom implementations must provide getJournal() (getTrace() is inherited when extending BizEntity/MongoBizEntity).

- Adds a JournalData to SQLSchedulerEntry and MongoSchedulerEntry. JournalData only contains transient fields, so no schema change.
- Silences the journal in markExecuted so that the execution counters (lastExecution, numberOfExecutions, runs) updated by the scheduler loop do not create a journal entry per run. Only user-driven changes and deletions are journaled.
- Shows the tracing/journal button in the scheduler entry editor.
Bildschirmfoto 2026-08-28 um 14 01 25

### Additional Notes

- This PR fixes or works on following ticket(s): NO Ticket
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
+++ b/src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
@@ -55,6 +55,9 @@ public class SQLSchedulerEntryProvider implements SchedulerEntryProvider<SQLSche
     @Override
     public void markExecuted(SQLSchedulerEntry job, LocalDateTime timestamp) {
         job.getSchedulerData().rememberExecution(timestamp);
+        // Only user-driven changes should be journaled - the execution counters updated here must not
+        // create a journal entry per run.
+        job.getJournal().setSilent(true);
         oma.update(job);
     }
 }
--- a/src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
+++ b/src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
@@ -55,8 +55,8 @@ public class SQLSchedulerEntryProvider implements SchedulerEntryProvider<SQLSche
     @Override
     public void markExecuted(SQLSchedulerEntry job, LocalDateTime timestamp) {
         job.getSchedulerData().rememberExecution(timestamp);
-        // Only user-driven changes should be journaled - the execution counters updated here must not
-        // create a journal entry per run.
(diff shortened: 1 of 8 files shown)


src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
```diff
@@ -55,6 +55,9 @@ public SQLSchedulerEntry fetchFullInformation(SQLSchedulerEntry job) {
     @Override
     public void markExecuted(SQLSchedulerEntry job, LocalDateTime timestamp) {
         job.getSchedulerData().rememberExecution(timestamp);
+        // Only user-driven changes should be journaled - the execution counters updated here must not
+        // create a journal entry per run.
+        job.getJournal().setSilent(true);
```
sabieber, line 60: **Optional:** Why not simply mark the execution fields as `@NoJournal`?
Same below

sabieber (approved):

src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
fhaScireum: Sollten wir nicht lieber `@NoJournal` bei den entsprechenden Feldern setzen?

ymo-sci: setSilent is needed for rememberExecution - otherwise each execution "edits" the schedulerEntry and so will be journal-logged.
For runs in SchedulerData ("Verbleibende Ausführungen") we want to log editing in the backend.
adding NoJournal is redundant for the both fields cause setSilent already suppresses all journals

fhaScireum (approved):

src/main/java/sirius/biz/jobs/scheduler/jdbc/SQLSchedulerEntryProvider.java
sabieber: This is only true for fields edited by `rememberExecution` that are not `@NoJournal` annotated.

If we forget `setSilent` in another method those technical fields will be logged again. Don't think that can be called redundant :)

jakobvogel (approved):

fhaScireum (approved):

jmuscireum (approved):

PR #2450, opened 2026-09-19
Drops the trailing gap below the last element in a KB card

jakobvogel:
### Description

A knowledge base card pads itself, and Bootstrap gives a trailing paragraph or list a bottom margin of its own, so the two stack up and every section and hint box ends in a visibly empty strip.

Articles written in Tagliatelle work around this with a manual `class="mb-0"` on the closing element — that class appears in 10 files in MEMOIO and 3 in OXOMI. **Markdown articles have no such escape**: Markdown carries no class syntax and `KnowledgeBaseMarkdownRenderer` runs with `escapeHtml(true)`, so the author cannot reach the element at all. That makes it a framework problem rather than an authoring omission.

The card markup is emitted from two places, so both now carry a `kb-section` class and the stylesheet can be scoped to the construct instead of to `.card`, which is every card in every Tycho application:

| | |
|---|---|
| `k:section` | also backs `k:infoSection` and `k:warnSection`, and the generated chapter/article listings |
| `AlertNodeRenderer` | writes the same structure by hand for a Markdown alert |

Two cases need more than a reset on the trailing element:

- **Margin collapsing.** A container with no bottom padding collapses its last child's bottom margin outwards, so zeroing a `blockquote` or a list of paragraph-wrapped items changes nothing — the innermost block has to be reached as well. Bootstrap gives `ol, ul, dl` only a `padding-left` and `blockquote` no padding at all.
- **`!important` utilities.** A hint box closing a section is a card carrying `mb-4`, and the Bootstrap spacing utilities are `!important`, so that margin outlives a plain reset and has to be overridden in kind.

The nesting in the stylesheet is not cosmetic: the SASS parser rejects `> :last-child` and `> *:last-child` outright (`Unexpected Token`), but accepts the same selector written as `&:` inside a nested block. Verified by compiling `kb.scss` with `Parser` + `Generator` directly and reading the generated selectors.

**Known gap, deliberately not fixed here:** `k:previewImage` emits a `<script>` and a `<style>` after its own `col mb-4` wrapper, so no `:last-child` rule reaches the wrapper. A section ending in a preview image keeps its margin and wants a separate fix.

### Additional Notes

- No ticket — found while writing the first Markdown KB articles for MEMOIO (MIO-6580), which is where the missing `mb-0` escape hatch surfaced.
- Draft: I would like a second opinion on the `!important` override before this is merged. It has specificity (0,3,1) and therefore beats an author's `<k:section class="mb-5">` as the last nested block. No article does this today (zero hits across sirius-biz, MEMOIO and OXOMI), so it is a latent trap rather than a live bug, but it is a deliberate trade worth a review.

### Checklist

- [x] Code change has been tested and works locally — `KnowledgeBaseMarkdownRendererTest` 20/20; `kb.scss` compiled with sirius-web's own SASS engine and the generated selectors checked; rendered pages inspected in a running MEMOIO instance
- [x] Code was formatted via IntelliJ and follows best practices — both Java files report "Formatted well"; **SonarLint has not been run** (it cannot run headlessly)
- [x] Patch Tasks: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakobvogel:
Both review points addressed.

**Nested lists** — confirmed, the rule did nothing there. A list with paragraph-wrapped items, and a list nested inside one, measured 33px before *and* after; the tight nested case only looked fine because Bootstrap already zeroes `ul ul`. The reset now walks the whole last-child chain instead of a hand-picked list of elements, which also covers a blockquote inside a list item. All shapes now land at the 17px floor.

**Regression coverage** — added `KnowledgeBaseStylesheetTest`, which compiles `kb.scss` through the SASS engine and asserts the chain and the `!important` override. Collapsing the chain back to one level parses cleanly but fails the test, which was the gap.

mkeckmkeck (approved):

sabieber (approved):

